Issue #36, opened 2014-12-30
Allow use_class_as_property decorator to be used inside DOMElement classes


whimboo:
What @chmanchester didn't like lately is the usage of lambda for instantiating BaseLib classes in UI classes, e.g. `return MenuBar(lambda: self.marionette)`:
https://github.com/mozilla/firefox-greenlight-tests/blob/master/greenlight/lib/ui/base_window.py#L47

This is currently necessary because any class based on DOMElement does not hold a client getter. It would be good to make use of the decorator even in UI modules.

At the same time we should really change all occurrences of self.client to self.marionette to be consistent across the whole repository. For me it is always confusing when to use self.marionette and when self.client.

whimboo:
PR #37 has been merged.

PR #37, opened 2014-12-30
Allow use_class_as_property decorator to be used inside DOMElement classes (#36)

whimboo:
This PR fixes issue #36. @chmanchester can you please have a look at?


chmanchester:
This looks like a reasonable approach, and agreed we should only expose one of "client" and "marionette". @ahal introduced the pattern, he may have more ideas here.

whimboo:
Andrew replied to me via IRC that he agrees on this proposal. I will re-run the tests and if all is still  passing, I will get it landed. Thanks!

Issue #39, opened 2015-01-06
Enhance and fix the prefs module for all the features we need


whimboo:
I found some issues and missing pieces in the prefs.js module. We should get this fixed so that we have a complete module, which has the same feature set as our module for Mozmill tests.

whimboo:
PR #40 has been merged.

PR #40, opened 2015-01-07
Enhance and fix the prefs module for all the features we need (#39)

whimboo:
This fixes issue #39. @chmanchester can you please have a look? With all the changes we have everything what we need and are nearly identical with the mozmill prefs module, except for setting complex preferences. Checking our Mozmill tests I actually haven't found any instance so far where this would be necessary. Means I'm happy to not include it right now.

### DIFF
--- a/greenlight/lib/api/prefs.py
+++ b/greenlight/lib/api/prefs.py
@@ -2,109 +2,206 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
+from marionette.errors import MarionetteException
+
 from ..base import BaseLib
 
 
-class DefaultPrefBranch(BaseLib):
+class Preferences(BaseLib):
     archive = {}
 
-    @classmethod
-    def _cast(cls, value):
-        """
-        Interpolate a preference from a string.
+    def get_pref(self, pref_name, default_branch=False, interface=None):
+        """Retrieves the value of a preference.
 
-        - integers will get cast to integers
-        - true/false will get cast to True/False
-        - anything enclosed in single quotes will be treated as a string with
-          the ''s removed from both sides
-        """
+        To retrieve the value of a preference its name has to be specified. By
+        default it will be read from the `user` branch, and returns the current
+        value. If the default value is wanted instead, ensure to flag that by
+        passing `True` via default_branch.
 
-        if not isinstance(value, basestring):
-            return value  # no op
-        quote = "'"
-        if value == 'true':
-            return True
-        if value == 'false':
-            return False
-        try:
-            return int(value)
-        except ValueError:
-            pass
-        if value.startswith(quote) and value.endswith(quote):
-            value = value[1:-1]
-        return value
+        It is also possible to retrieve the value of complex preferences, which
+        do not represent an atomic type like `basestring`, `int`, or `boolean`.
+        Specify the interface of the represented XPCOM object via `interface`.
 
-    def get_pref(self, pref):
-        """
-        Retrive a preference.
+        :param pref_name: The preference name
+        :param default_branch: Optional, flag to use the default branch,
+         default to `False`
+        :param interface: Optional, interface of the complex preference,
+         default to `None`. Possible values are: `nsILocalFile`,
+         `nsISupportsString`, and `nsIPrefLocalizedString`
 
-        :param pref: The preference to inspect,e.g
-                     browser.tabs.remote.autostart
-        :returns: The value of the specified pref.
+        :returns: The preference value.
         """
+        assert pref_name is not None
+
+        # Bug 1118825 - None is causing an exception to be raised
+        interface = interface or ''
+
         with self.marionette.using_context('chrome'):
-            value = self.marionette.execute_script("""
-              let pref = arguments[0];
-              let prefBranch = Cc["@mozilla.org/preferences-service;1"]
-                              .getService(Ci.nsIPrefBranch);
-              let type = prefBranch.getPrefType(pref);
+            return self.marionette.execute_script("""
+              Cu.import("resource://gre/modules/Services.jsm");
+
+              let pref_name = arguments[0];
+              let default_branch = arguments[1];
+              let interface = arguments[2];
+
+              let prefBranch;
+              if (default_branch) {
+                prefBranch = Services.prefs.getDefaultBranch("");
+              }
+              else {
+                prefBranch = Services.prefs;
+              }
+
+              // If an interface has been set, handle it differently
+              if (interface != '') {
+                return prefBranch.getComplexValue(pref_name,
+                                                  Ci[interface]).data;
+              }
+
+              let type = prefBranch.getPrefType(pref_name);
 
               switch (type) {
                 case prefBranch.PREF_STRING:
-                  return prefBranch.getCharPref(pref);
+                  return prefBranch.getCharPref(pref_name);
                 case prefBranch.PREF_BOOL:
-                  return prefBranch.getBoolPref(pref);
+                  return prefBranch.getBoolPref(pref_name);
                 case prefBranch.PREF_INT:
-                  return prefBranch.getIntPref(pref);
+                  return prefBranch.getIntPref(pref_name);
                 case prefBranch.PREF_INVALID:
                   return null;
               }
-            """, script_args=[pref])
+            """, script_args=[pref_name, default_branch, interface])
 
-        return self._cast(value)
+    def reset_pref(self, pref_name):
+        """Resets a user set preference.
 
-    def set_pref(self, pref, value):
+        Every modification of a preference will turn its status into a user-set
+        preference. To restore the default value of the preference, call this
+        function once. Further calls have no effect as long as the value hasn't
+        changed again.
+
+        In case the preference is not present on the default branch, it will be
+        completely removed.
+
+        :param pref_name: The preference to reset
+
+        :returns: `True` if a user preference has been removed
         """
-        Sets the specified pref to value. Also archives the old value so that
-        the pref can be restored with `restore_pref`.
+        assert pref_name is not None
 
-        :param pref: The preference to set.
-        :param value: The value to set the preference to.
-        """
         with self.marionette.using_context('chrome'):
-            self.archive[pref] = self.get_pref(pref)
+            return self.marionette.execute_script("""
+              Cu.import("resource://gre/modules/Services.jsm");
+              let prefBranch = Services.prefs;
 
-            ret = self.marionette.execute_script("""
-              let pref = arguments[0];
+              let pref_name = arguments[0];
+
+              if (prefBranch.prefHasUserValue(pref_name)) {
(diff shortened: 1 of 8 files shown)


greenlight/lib/api/prefs.py
```diff
@@ -2,109 +2,174 @@
...
-        - integers will get cast to integers
-        - true/false will get cast to True/False
-        - anything enclosed in single quotes will be treated as a string with
-          the ''s removed from both sides
+        :param pref_name: The preference to clear
         """
+        with self.marionette.using_context('chrome'):
+            return self.marionette.execute_script("""
+              Cu.import("resource://gre/modules/Services.jsm");
+              let prefBranch = Services.prefs.QueryInterface(Ci.nsIPrefBranch);
```
chmanchester: We know what Services.prefs supports, so is the QI necessary here?

greenlight/lib/api/prefs.py
```diff
@@ -2,109 +2,174 @@
...
-            value = value[1:-1]
-        return value
+              let pref_name = arguments[0];
+
+              if (prefBranch.prefHasUserValue(pref_name)) {
+                prefBranch.clearUserPref(pref_name);
+                return true;
+              }
+              else {
+                return false;
```
chmanchester: If the return is meaningful to us here, maybe we should just wrap prefHasUserValue separately.

greenlight/lib/api/prefs.py
```diff
@@ -2,109 +2,174 @@
...
+              let pref_name = arguments[0];
+              let default_branch = arguments[1];
+              let interface = arguments[2];
+
+              let prefBranch = undefined;
+              if (default_branch) {
+                prefBranch = Services.prefs.getDefaultBranch("");
+              }
+              else {
+                prefBranch = Services.prefs.QueryInterface(Ci.nsIPrefBranch);
```
chmanchester: As above, this seems to work without the QI.

greenlight/lib/api/prefs.py
```diff
@@ -2,109 +2,174 @@
...
+
+        :returns: The value of the specified preference.
         """
+        assert pref_name
+
+        # Bug 1118825 - None is causing an exception to be raised
+        interface = interface or ''
+
         with self.marionette.using_context('chrome'):
             value = self.marionette.execute_script("""
```
chmanchester: This would be consistent with clear_user_pref if it was just "return self.marionette.execute_script(""""

greenlight/lib/api/prefs.py
```diff
@@ -2,109 +2,174 @@
...
-              let prefBranch = Cc["@mozilla.org/preferences-service;1"]
-                              .getService(Ci.nsIPrefBranch);
-              let type = prefBranch.getPrefType(pref);
+              Cu.import("resource://gre/modules/Services.jsm");
+
+              let pref_name = arguments[0];
+              let default_branch = arguments[1];
+              let interface = arguments[2];
+
+              let prefBranch = undefined;
```
chmanchester: "let prefBranch;" is equivalent.

greenlight/lib/api/prefs.py
```diff
@@ -2,109 +2,174 @@
...
+        :param interface: Interface to use for the complex value (default None)
+                          Possible values are: nsILocalFile, nsISupportsString,
+                          and nsIPrefLocalizedString
+
+        :returns: The value of the specified preference.
         """
+        assert pref_name
+
+        # Bug 1118825 - None is causing an exception to be raised
+        interface = interface or ''
```
chmanchester: Alternatively you could build a list of args to pass to the script and not include interface if it's None, then just check for "if (default_branch) {" so this doesn't have to be changed again when that bug is fixed.

greenlight/lib/api/prefs.py
whimboo: Good catch! You are right, it also QI's it itself. So we should be able to remove that.

greenlight/lib/api/prefs.py
```diff
@@ -2,109 +2,174 @@
...
-        the pref can be restored with `restore_pref`.
+        Also archives the old value so that the preference can be restored with
+        `restore_pref`.
 
-        :param pref: The preference to set.
+        :param pref_name: Name of the preference to set.
         :param value: The value to set the preference to.
         """
+        assert pref_name is not None
+        assert value is not None
```
chmanchester: You could change this to "assert value and type(value) in [str, int, bool]" and save checking on this in javascript and asserting again below.

greenlight/lib/api/prefs.py
whimboo: It's just a nice to have. I'm even happy to remove it. Shall I do?

greenlight/lib/api/prefs.py
whimboo: Oh sure. It's a left-over from the `_cast()` call we had before. It can indeed return immediately.

greenlight/lib/api/prefs.py
```diff
@@ -2,109 +2,174 @@
...
 
-        :param pref: The preference to restore. It must be a preference that
-                     was previously set using `set_pref`.
+        If the preference is not a default preference and has been newly
+        created, it will be removed.
+
+        :param name: The name of the preference to restore.
         """
-        return self.set_pref(pref, self.archive[pref])
+        assert pref_name
```
chmanchester: Asserting arguments aren't falsey strikes me as a slightly odd pattern. I don't expect we'll ever find "", 0, [], {}, None, etc. in our pref archive. It's not really hurting anything, but it might be more straightforward to start this function with "if pref_name not in self.archive: raise ...".

greenlight/lib/api/prefs.py
whimboo: Erm, sure :)

greenlight/lib/tests/test_prefs.py
```diff
@@ -0,0 +1,150 @@
...
+        self.prefs.set_pref(self.string_pref, 'unittest')
+        self.assertEqual(self.prefs.get_pref(self.string_pref), 'unittest')
+        self.prefs.restore_pref(self.string_pref)
+        self.assertEqual(self.prefs.get_pref(self.string_pref), orig_value)
+
+        # Test with multiple set_pref call
+        self.prefs.set_pref(self.string_pref, 'unittest')
+        self.prefs.set_pref(self.string_pref, 'unittest2')
+        self.assertEqual(self.prefs.get_pref(self.string_pref), 'unittest2')
+        self.prefs.restore_pref(self.string_pref)
```
chmanchester: Maybe add a check that restoring the pref again throws.

greenlight/lib/tests/test_prefs.py
```diff
@@ -0,0 +1,150 @@
...
+        # boolean as int
+        self.prefs.set_pref(self.int_pref, False)
+        self.assertEqual(self.prefs.get_pref(self.int_pref), 0)
+
+        self.prefs.set_pref(self.int_pref, True)
+        self.assertEqual(self.prefs.get_pref(self.int_pref), 1)
+
+        # int as string
+        self.prefs.set_pref(self.string_pref, 54)
+        self.assertEqual(self.prefs.get_pref(self.string_pref), '54')
```
chmanchester: If the casting behavior is in the underlying pref service, I don't think testing here adds much value.

greenlight/lib/api/prefs.py
whimboo: We do not check for the types of those values for consistency but to figure out which type a new pref has to be. We fail at the moment because so far we are trying to get the type via the stored pref, which in this case does not exist. So the assert as you propose here will only make sure that the right types are allowed for that method. As long as we do not use complex prefs this should be fine. Do you still want to have this?

greenlight/lib/api/prefs.py
chmanchester: I just noticed this is emulating the mozmill prefs library, so if leaving it will simplify test conversion that's fine.

greenlight/lib/api/prefs.py
whimboo: Then we can simply remove this line, because we automatically will reach the except block which raises the Marionette exception.

greenlight/lib/tests/test_prefs.py
whimboo: Strange. I thought I had that, but you are right. It's missing.

greenlight/lib/tests/test_prefs.py
whimboo: I added some of those checks when we still had the `_cast()` method which actually returned the wrong type of value. So for basestring our `get_pref()` method returned an int.

Also with those tests I just wanted to make sure that we could cover changes in Firefox when our unit tests will be run on buildbot at one time. If we don't have them, we will only notice those things in our tests. Well, we should not do those strange things anyway, but I like safety. Tests are run quickly, and I don't see it would hurt. What do you think?

greenlight/lib/api/prefs.py
whimboo: The mozmill prefs lib is partly outdated given that since Firefox 6.0 this method doesn't throw anymore. So it could have been removed. One good thing here simply is that you will know if a preference you specified did exist and was most likely removed.

greenlight/lib/api/prefs.py
chmanchester: I think it would be clearer to someone trying to debug this and would preserve the behavior of the current function almost exactly.

greenlight/lib/tests/test_prefs.py
chmanchester: No, I don't think they hurt. If this would have caught the bug you're referring leave them in.

chmanchester:
Overall this looks good if you want to fixup and merge. Happy to discuss check it out again if you prefer.

greenlight/lib/api/prefs.py
chmanchester: Ah, right. I was just thinking because clearUserPref doesn't hurt when one isn't set, this could just be "prefBranch.clearUserPref(pref_name); return prefHasUserValue(pref_name);", so I started to think this is really a separate matter. Your call.

greenlight/lib/api/prefs.py
whimboo: Ok, I will add this then.

greenlight/lib/api/prefs.py
whimboo: Well, you cannot call `prefHasUserValue()` after `clearUserPref()` because this will always be `False`. At this time the preference has already been removed. So I have chosen this logic instead of assigning the value to a new temporary variable.

whimboo:
Thanks Chris! I pushed another commit with all the review comments included. Before I merge this in I want to enhance the documentation for this module.

greenlight/lib/api/prefs.py
chmanchester: Oh, of course. That sounds fine.

whimboo:
I pushed a new commit with the API docs. I haven't added examples yet, which we can do later once it's more clear how to structure the docs.

Issue #41, opened 2015-01-08
Add Travis CI


whimboo:
We should add Travis CI to let our latest code automatically tested in PRs and when merged.

whimboo:
PR #42 has been merged.

PR #42, opened 2015-01-08
Add Travis CI (#41)

whimboo:
This PR will fix issue #41. I might have to play a bit before it's working correctly.


whimboo:
@chmanchester has a great idea on IRC to pin the Nightly version to a specific version to avoid problems with incompatibilities. 

I would propose that we pin to the exact same version of Nightly the marionette client has been released off. That way we should be kinda safe.

whimboo:
@chmanchester mind having a look at this PR? If you spot anything please let me know.

.travis.yml
```diff
@@ -0,0 +1,29 @@
...
+env: DISPLAY=:99
+
+before_install:
+    - sudo apt-get update
+    - sh -e /etc/init.d/xvfb start
+
+install:
+    - sudo apt-get install flashplugin-nonfree
+    - pip install mozdownload pep8
+    - python setup.py develop
```
whimboo: I wonder if we should move the call to setup.py down and run pep8 before. That would save us some seconds and an unnecessary installation of the marionette package and all deps.

greenlight/lib/api/keys.py
```diff
@@ -14,6 +14,7 @@ def __init__(self, marionette_getter):
 
     @property
     def ACCEL(self):
-        if self.marionette_getter().session_capabilities['platformName'] == 'DARWIN':
+        if self.marionette_getter().session_capabilities['platformName'] == \
+                'DARWIN':
```
chmanchester: Here and elsewhere please use implicit line continuations via parentheses for long expressions. Pep8 suggests as much, but I guess it isn't included in the linter.

greenlight/lib/tests/test_tabs.py
```diff
@@ -36,7 +36,8 @@ def setUp(self):
 
         def tabs_loaded(m):
             label = m.execute_script("""
-                return gBrowser.tabs[gBrowser.tabs.length-1].getAttribute('label');
+              return gBrowser.tabs[gBrowser.tabs.length-1].
+                              getAttribute('label');
```
chmanchester: Prefer the trailing dot on the next line.

setup.py
```diff
@@ -7,8 +7,8 @@
 PACKAGE_VERSION = '0.1'
 
 deps = [
-    'manifestparser',
-    'marionette-client',
+    'manifestparser == 0.9',
+    'marionette-client == 0.8.6',
```
chmanchester: marionette-client depends on manifestparser, you should be able to remove the manifestparser dependency.

greenlight/tests/functional/keyboard_shortcuts/test_browser_window.py
```diff
@@ -39,9 +39,11 @@ def test_search_field(self):
 
         keys = [self.keys.ACCEL]
         if self.marionette.session_capabilities['platformName'] == 'LINUX':
-            keys.append(self.browser.get_localized_entity('searchFocusUnix.commandkey'))
+            keys.append(self.browser.get_localized_entity(
+                        'searchFocusUnix.commandkey'))
```
chmanchester: Maybe I've gotten used to working on things that don't have a strict 80 character limit, but I think slightly longer lines are more readable for some of these cases.

.travis.yml
chmanchester: Sounds like a good idea. Today is the beginning of my experience with travis, but this all seems pretty foolproof.

greenlight/lib/api/keys.py
whimboo: Which linter are you using? I have `Python Flake8 Lint` installed in Sublime Edit and it doesn't blame anything.

I will rewrite this code above given that it doesn't make sense to do this check all the time. It's enough to do in __init__.

greenlight/lib/tests/test_tabs.py
whimboo: Sure. That's even how it should be. :)

setup.py
whimboo: I was told to add those dependencies in case your own package makes use of it. Given that this is not the case so far, we can indeed remove this dependency.

greenlight/tests/functional/keyboard_shortcuts/test_browser_window.py
whimboo: This will be refactored anyway soonish, so I thought this would be fine.

whimboo:
I pushed two more commits which fix the review comments, and improve the travis config file.

All tests are passing, so I'm going to get this merged. Thanks  @chmanchester for your review!

whimboo:
Rebased the commits and merged to master as https://github.com/mozilla/firefox-greenlight-tests/commit/1c9bf47db35875291c86299ab34f2d23bc25680b.

greenlight/lib/api/keys.py
chmanchester: I'm not using a linter, pep8 has a note about it in the "Maximum Line Length" section.

greenlight/tests/functional/keyboard_shortcuts/test_browser_window.py
chmanchester: I don't think it's acceptable to attach a linter to our CI that requires us to re-write our code in less readable ways. @whimboo, can we set the limit to 100 or disable this check?

whimboo: Most of the necessary changes on this PR are temporarily and will go away soon. We haven't had any problems with 80 char limits in the past years, so not sure what your experience was. Lets ask @davehunt how it was working for gaia ui tests, and which coding styles especially line length they prefer.

chmanchester: We don't impose an 80 character limit on python in mozilla central. It's a subjective matter, but I am strongly of the opinion that in certain cases in python, lines longer than 80 characters are more readable than what is necessary to break them up. I would prefer to leave this particular matter to individual contributors and reviewers.

whimboo: Please also see the A-Team bootcamp: http://ateam-bootcamp.readthedocs.org/en/latest/reference/python-style.html. There is no exclusion for that at the moment. Just follow PEP8, and that would include 80 chars. Maybe that needs to be updated then too.

chmanchester: PEP 8 acknowledges an exception to this guideline among teams that agree to it.

davehunt: There's no strict line length limit for Gaia UI. I personally try to stick to 80 characters unless going over makes the code considerably more readable.

whimboo: Sounds like a good compromise. So what about lifting it up to 100 chars? We can tweak that easily by specifying the `--max-line-length` option.

PR #43, opened 2015-01-09
Add an assertion in FirefoxTestCase to ensure tests don't leak window ha...

chmanchester:
...ndles and fix tests not to do so.

### DIFF
--- a/greenlight/harness/testcase.py
+++ b/greenlight/harness/testcase.py
@@ -19,10 +19,20 @@ class FirefoxTestCase(MarionetteTestCase, Puppeteer):
         MarionetteTestCase.setUp(self, *args, **kwargs)
         Puppeteer.set_marionette(self, self.marionette)
 
+        self._start_handle_count = len(self.marionette.window_handles)
         self.marionette.set_context('chrome')
         self.browser = self.windows.current
 
     def tearDown(self, *args, **kwargs):
+        self.prefs.restore_all_prefs()
         MarionetteTestCase.tearDown(self, *args, **kwargs)
 
-        self.prefs.restore_all_prefs()
+        # This assertion should be run after all other tearDown code
+        # so that in case of a failure, further tests will not run
+        # in a state that is more inconsistent than necessary.
+        win_count = len(self.marionette.window_handles)
+        self.assertEqual(win_count, self._start_handle_count,
+                         "A test must not leak window handles. "
+                         "This test started the browser with %s open "
+                         "top level browsing contexts, but ended with %s." %
+                         (self._start_handle_count, win_count))
--- a/greenlight/lib/tests/test_tabs.py
+++ b/greenlight/lib/tests/test_tabs.py
@@ -48,9 +48,7 @@ class TestTabs(FirefoxTestCase):
             gBrowser.removeAllTabsBut(gBrowser.tabs[0]);
         """)
 
-        # Needs to be set by the harness itself
-        self.prefs.restore_pref('browser.tabs.warnOnClose')
-        self.prefs.restore_pref('browser.tabs.warnOnCloseOtherTabs')
+        self.wait_for_condition(lambda _: len(self.browser.tabbar.tabs) == 1)
 
         FirefoxTestCase.tearDown(self)
 
--- a/greenlight/tests/functional/toolbar/test_back_forward.py
+++ b/greenlight/tests/functional/toolbar/test_back_forward.py
@@ -12,6 +12,14 @@ class TestBackForward(FirefoxTestCase):
     def setUp(self):
         FirefoxTestCase.setUp(self)
 
+        with self.marionette.using_context("content"):
+            self.marionette.navigate("about:blank")
+
+        self.marionette.execute_script("""
+            let count = gBrowser.sessionHistory.count;
+            gBrowser.sessionHistory.PurgeHistory(count);
+        """)
+
         self.test_urls = [
             'layout/mozilla.html',
             'layout/mozilla_mission.html',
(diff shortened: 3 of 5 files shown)


chmanchester:
@whimboo, can you take a look at this?

chmanchester:
That failure looks unrelated.

greenlight/harness/testcase.py
```diff
@@ -19,10 +19,17 @@ def setUp(self, *args, **kwargs):
...
-        MarionetteTestCase.tearDown(self, *args, **kwargs)
+        win_count = len(self.marionette.window_handles)
+        self.assertEqual(win_count, self._start_handle_count,
+                         "A test must not leak window handles. "
+                         "This test started the browser with %s open "
+                         "top level browsing contexts, but ended with %s." %
+                         (self._start_handle_count, win_count))
 
         self.prefs.restore_all_prefs()
+        MarionetteTestCase.tearDown(self, *args, **kwargs)
```
whimboo: I wonder if we shouldn't close all the remaining open windows in tearDown ourselves? Just as what I did with restore_all() for preferences. If a test really keeps windows open, we might log some warnings. Keep in mind that we cannot do the above because assertEqual would cause an early return from tearDown and any following code will not be executed. This even applies to `MarionetteTestCase.tearDown()`.

greenlight/lib/tests/test_tabs.py
```diff
@@ -47,9 +47,7 @@ def tearDown(self):
             gBrowser.removeAllTabsBut(gBrowser.tabs[0]);
         """)
 
-        # Needs to be set by the harness itself
-        self.prefs.restore_pref('browser.tabs.warnOnClose')
-        self.prefs.restore_pref('browser.tabs.warnOnCloseOtherTabs')
+        self.wait_for_condition(lambda _: len(self.browser.tabbar.tabs) == 1)
```
whimboo: This is not really related to the topic of this PR. Have you seen any failures? The whole code needs a rewrite anyhow.

greenlight/tests/functional/toolbar/test_back_forward.py
```diff
@@ -12,6 +12,14 @@ class TestBackForward(FirefoxTestCase):
...
         FirefoxTestCase.setUp(self)
 
+        with self.marionette.using_context("content"):
+            self.marionette.navigate("about:blank")
+
+        self.marionette.execute_script("""
+            let count = gBrowser.sessionHistory.count;
+            gBrowser.sessionHistory.PurgeHistory(count);
+        """)
+
```
whimboo: This is not related at all to this PR. Can you explain why this has been added? What this test really would need is to wait for an event, and that module is still to create once I'm done with the observers.

greenlight/harness/testcase.py
chmanchester: I think things could get worse if we allow tests to leak windows. They could easily do things like leave marionette switched into the wrong handle and cause future tests to fail. You mean we're going to have a problem because subsequent tear down code will not run in case of a failure of this assertion? I think that might be ok -- later tests might fail, but the problem will be fairly apparent given the output, and usually straightforward to fix.

whimboo:
Can you also please merge in latest master so the Travis failure will go away? Thanks.

greenlight/lib/tests/test_tabs.py
chmanchester: This fails the assertion I added, because removeAllTabsBut is (apparently) asynchronous, but we don't wait. Removing the restore_pref calls is unrelated, but we don't need them now with your restore_all method.

greenlight/tests/functional/toolbar/test_back_forward.py
chmanchester: This test fails the assertion I added. It relies on having a certain amount of history in the tab so the back button ends up disabled later on. It's only passing now because it runs in a leaked tab from an earlier test.

greenlight/harness/testcase.py
whimboo: I'm still in favor of a `closeAllChromeWindows` method in windows.js, which can be executed in tearDown. It should close all windows beside the last open browser window (optionally added as exception). We do this in our Mozmill tests, and this is pretty failure proof.

Raising an exception as above will not help, giving that we do not close any left open windows. But we play even bad with MarionetteTestCase.tearDown, in case of not executing it.

chmanchester: How is there an issue when the assertion doesn't fail? I don't think I see what you mean. I would perfer disallowing tests that leave things in an inconsistent state, because in general I think we'll find it's hard to come up with a one size fits all cleanup routine, but we can discuss further.

greenlight/lib/tests/test_tabs.py
whimboo: When I look at http://mxr.mozilla.org/mozilla-central/source/browser/base/content/tabbrowser.xml#1938 I don't see that it is asynchronous, but this check doesn't hurt. So all fine here! Thanks for the explanation.

greenlight/tests/functional/toolbar/test_back_forward.py
whimboo: Ok, so lets this go in as is until we have better methods for handling tabs, and setting up and tearing down tests.

greenlight/harness/testcase.py
whimboo: There can always be cases a test accidentally doesn't close the window. It might not always be even the fault of the test, but just some underlying code has been thrown an assertion, and the line for closing all the remaining windows is not reached.

If there was really no assertion before and windows are still open, an extra assertion might be good but at the same time we should really close by default all the other open windows except the last browser window. With that action we can make the suite more foolproof by ensuring that the next test has the correct conditions and will not fail. In your current case all the windows will still kept open until a test closes them all or all tests are done, right?

chmanchester: We talked about this on irc, and agreed that while there's more to be done in terms of a robust tearDown of each test case, this assertion is worth taking so long as it runs after all other cleanup to in case of failure it doesn't make matters worse for further tests.

Issue #44, opened 2015-01-09
Rename repository to firefox-ui-tests


whimboo:
To stay in sync with the naming of test repositories my suggestion would be that we call our repository `firefox-ui-tests`. The benefit is that it is clear what it covers, and is general enough to even survive possible future test harness changes.

I talked with @AutomatedTester on IRC and he is happy with that name. @ahal and @chmanchester what do you both think of that name?

whimboo:
Also cc'ing @ctalbert, @jonallengriffin, and @jmaher for feedback.

whimboo:
The rename has been done. I will update the repository content early tomorrow morning.

whimboo:
PR #46 fixed the remaining todos. So this is complete now.

PR #46, opened 2015-01-13
Rename repository to firefox-ui-tests (#44)

whimboo:
@chmanchester mind having a look?


chmanchester:
I'm not a fan of how this looks: I know it's more or less consistent with gaiatest, but firefoxtests seems too general. Is firefox_ui_tests too ugly?

whimboo:
Chris the CLI is only temporary. At the end we will have a couple of different entry scripts.

chmanchester:
README.md will want updates with this as well.

whimboo:
@chmanchester I fixed the readme file, and updated the console command to `firefox-ui-tests`. If you don't mind I would keep the `firefoxtests` subfolder regarding all the underscores we would have to use. As alternative we could move out the harness to `/harness` which would make it easier later to uncouple the dependencies also for the puppeteer modules. I could do that in a follow-up issue.

chmanchester:
You've pointed out that the repository structures here will likely change as we progress, so this is probably good for now.

whimboo:
That is true, and we might have to find a solution soon, before we have to update too many of the tests. Maybe I can do that soon for the harness pieces. Thanks for all your comments!

whimboo:
I squashed the commits and merged the PR as https://github.com/mozilla/firefox-ui-tests/commit/9109e23acf38ae2ac0ac0dca6e4f9b74a8aa96a7.

PR #47, opened 2015-01-15
Add e10s to default prefs in the runner, skip tests navigating to about:...

chmanchester:
... pages.

### DIFF
--- a/firefox_puppeteer/__init__.py
+++ b/firefox_puppeteer/__init__.py
@@ -57,6 +57,14 @@ class Puppeteer(object):
         See the :class:`~api.prefs.Preferences` reference.
         """
 
+    @use_class_as_property('api.appinfo.AppInfo')
+    def appinfo(self):
+        """
+        Provides access to members of the appinfo  api.
+
+        See the :class:`~api.appinfo.AppInfo` reference.
+        """
+
 
 class DOMElement(HTMLElement):
     """
--- a/firefox_puppeteer/api/appinfo.py
+++ b/firefox_puppeteer/api/appinfo.py
@@ -1,0 +1,22 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from ..base import BaseLib
+
+
+class AppInfo(BaseLib):
+
+    @property
+    def browserTabsRemoteAutostart(self):
+        return self._get_property("browserTabsRemoteAutostart")
+
+    def _get_property(self, prop_name):
+        with self.marionette.using_context('chrome'):
+            return self.marionette.execute_script("""
+              try {
+                return Services.appinfo[arguments[0]];
+              } catch (e) {
+                return null;
+              }
+            """, script_args=[prop_name])
--- a/firefox_ui_harness/decorators.py
+++ b/firefox_ui_harness/decorators.py
@@ -1,0 +1,15 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from marionette import SkipTest
+
+from firefox_puppeteer.api import appinfo
+
+
+def skip_if_e10s(target):
+    def wrapper(self, *args, **kwargs):
+        if self.appinfo.browserTabsRemoteAutostart:
+            raise SkipTest("Skipping due to e10s")
+        return target(self, *args, **kwargs)
+    return wrapper
--- a/firefox_ui_harness/runtests.py
+++ b/firefox_ui_harness/runtests.py
@@ -2,6 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+import copy
 import sys
 
 from marionette import BaseMarionetteTestRunner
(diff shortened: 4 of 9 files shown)


chmanchester:
@whimboo I'm happy to rebase this after whatever you're working on, but wanted to get something going so we don't have any surprises around this.

firefox_ui_harness/decorators.py
```diff
@@ -0,0 +1,21 @@
...
+
+def skip_if_e10s(target):
+    def wrapper(self, *args, **kwargs):
+        with self.marionette.using_context('chrome'):
+            multi_process_browser = self.marionette.execute_script("""
+            try {
+              return Services.appinfo.browserTabsRemoteAutostart;
+            } catch (e) {
+              return false;
+            }""")
```
whimboo: It will be good to have that as a method in an API module, so we have a single definition of it.

firefox_ui_harness/runtests.py
```diff
@@ -14,10 +15,18 @@
...
+        "browser.tabs.remote.autostart": True,
+    }
 
     def __init__(self, *args, **kwargs):
         if not kwargs.get('server_root'):
             kwargs['server_root'] = firefox_ui_tests.resources
+        prefs = kwargs.get('prefs', {})
+        extra_prefs = copy.deepcopy(ReleaseTestRunner.extra_prefs)
+        extra_prefs.update(prefs)
+        kwargs['prefs'] = extra_prefs
```
whimboo: Why can't we do this directly via `kwargs['prefs'] = prefs.update(extra_prefs)`? 

nit: please separate the block from the server_root code.

firefoxtests/harness/decorators.py
```diff
@@ -0,0 +1,21 @@
...
+            try {
+              return Services.appinfo.browserTabsRemoteAutostart;
+            } catch (e) {
+              return false;
+            }""")
+
+        if multi_process_browser:
+            raise SkipTest("Skipping due to e10s")
+        return target(self, *args, **kwargs)
+    return wrapper
```
whimboo: This is still a left-over for the old location. Please remove it.

firefox_ui_harness/runtests.py
chmanchester: This is just so we don't modify the class variable unnecessarily. Either way is probably ok in this case.

firefox_puppeteer/api/appinfo.py
```diff
@@ -0,0 +1,21 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from ..base import BaseLib
+
+
+class AppInfo(BaseLib):
+
+    @property
+    def browserTabsRemoteAutostart(self):
```
whimboo: I wouldn't call it like the preference. If there is a name change, we would have to update a lot of tests maybe. Lets call it `multi_process`, `e10s_enabled`, or simply `e10s`.

firefox_puppeteer/api/appinfo.py
```diff
@@ -0,0 +1,21 @@
...
+        return self._get_property("browserTabsRemoteAutoStart")
+
+    def _get_property(self, prop_name):
+        with self.marionette.using_context('chrome'):
+            return self.marionette.execute_script("""
+            try {
+              return Services.appinfo[arguments[0]];
+            } catch (e) {
+              return null;
+            }""", script_args=[prop_name])
```
whimboo: Please indent the inner code by 2 blanks and put the last closing bracket of the JS code in its own line.

firefox_ui_harness/decorators.py
```diff
@@ -0,0 +1,15 @@
...
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from marionette import SkipTest
+
+from firefox_puppeteer.api import appinfo
+
+
+def skip_if_e10s(target):
+    def wrapper(self, *args, **kwargs):
+        if appinfo.AppInfo.browserTabsRemoteAutostart:
```
whimboo: How does that work without instantiating the class? The property is not static.

firefox_puppeteer/api/appinfo.py
chmanchester: I'm mirroring the appinfo member, "browserTabsRemoteAutostart". I don't want to add another name that means the same thing.

firefox_ui_harness/decorators.py
chmanchester: Whoops, that probably only works by accident, I'll update it to be static.

firefox_puppeteer/api/appinfo.py
whimboo: Ok, but then please no camel-case but underscores. I wonder if we can return the whole appInfo object inside the execute_script() block so that we do not need dozen of helper properties like that.

firefox_ui_harness/decorators.py
whimboo: But how is the marionette getter passed in? Is an instance created silently here?

firefox_puppeteer/api/appinfo.py
chmanchester: I don't want to add another name that means the same thing. We could return appinfo but that might disagree with our marshalling algorithm now or in the future.

firefox_ui_harness/decorators.py
chmanchester: There's no instance, this is broken right now. I don't know that we're going to need that many properties of appinfo that aren't supplied by session_capabilities, I sort of wonder this is worth it right now vs. the original version.

firefox_puppeteer/api/appinfo.py
chmanchester: I pushed a fix so this actually checks the this correctly. I'd like to leave the idea of mirroring more properties to a follow up as needed (if we get to this, I think this would be a reasonable opportunity to use __getattr__).

whimboo: Would you mind to file a bug for getattr, which describes exactly what to do? It would be a good bug for a contributor starting on our project. Thanks.

firefox_ui_harness/decorators.py
whimboo: What's currently missing is the locale and the XPCOMABI. Do you think we can add those? At least the locale?

firefox_puppeteer/__init__.py
```diff
@@ -57,6 +57,12 @@ def prefs(self):
         See the :class:`~api.prefs.Preferences` reference.
         """
 
+    @use_class_as_property('api.appinfo.AppInfo')
+    def appinfo(self):
+        """
+        Provides access to members of the appinfo  api.
```
whimboo: Please also add a .rst file and a reference to the class similar to the other puppeteer properties. So we can extend that later.

whimboo:
Beside the documentation comment all looks fine to me. Please do not forget to squash the commits and to update the commit message so it includes the bug and reviewer.

PR #48, opened 2015-01-15
Bug 1122015 - Refactor repository into sub packages. r=chmanchester

whimboo:
This PR is the fix for https://bugzilla.mozilla.org/show_bug.cgi?id=1122015.


.travis.yml
```diff
@@ -28,6 +29,8 @@ before_script:
     - tar -xvf *firefox*.tar.bz2
 
 script:
+    # Run tests for each sub package
+    - firefox-ui-tests --binary firefox/firefox firefox_puppeteer/tests/manifest.ini
```
chmanchester: Can we not default to a top manifest anymore? That would be preferable, I think.

whimboo: Please reload the page. It's fixed now inside the harness itself.

PR #49, opened 2015-01-19
Bug 1123393 - Remove workarounds for None value in execute_script calls. r=chmanchester

whimboo:


firefox_puppeteer/api/prefs.py
```diff
@@ -53,7 +50,7 @@ def get_pref(self, pref_name, default_branch=False, interface=None):
               }
 
               // If an interface has been set, handle it differently
-              if (interface != '') {
+              if (interface) {
```
chmanchester: I'd consider "interface !== null" to reflect the cases the default is provided to the python method. Just checking for any truthy value means if someone passes an empty string to this function accidentally, the parameter will be ignored rather than getting passed to the underlying api and hopefully throwing.

whimboo: Sure, I will get this fixed. Thanks

PR #50, opened 2015-01-23
Bug 1123401 - Implement handling of chrome windows

whimboo:
Just for a sneak peak. This PR is not ready yet and needs some more methods and cleanups. But if you have time lets give me feedback.


firefox_puppeteer/__init__.py
```diff
@@ -48,6 +40,14 @@ def keys(self):
         See the :class:`~api.keys.Keys` reference.
         """
 
+    @property
+    def platform(self):
+        """Returns the lowercased platform name.
+
+        :returns: Platform name
+        """
+        return self.marionette.session_capabilities['platformName'].lower()
```
chmanchester: I think this is the sort of thing that should go in FirefoxTestCase.

firefox_puppeteer/api/keys.py
```diff
@@ -1,20 +0,0 @@
-# This Source Code Form is subject to the terms of the Mozilla Public
-# License, v. 2.0. If a copy of the MPL was not distributed with this file,
-# You can obtain one at http://mozilla.org/MPL/2.0/.
-
-import marionette
-
-
-class Keys(marionette.keys.Keys):
```
chmanchester: Removing this broke some tests. I've made use of it outside of the windows class, what's the plan for that?

firefox_puppeteer/errors.py
```diff
@@ -0,0 +1,17 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from marionette.errors import *
```
chmanchester: Please don't use wildcard imports.

firefox_puppeteer/tests/test_windows.py
```diff
@@ -3,98 +3,181 @@
...
 
         windows = self.windows.all
-        self.assertEquals(len(windows), 3)
+        self.assertEquals(len(windows), len(self.marionette.chrome_window_handles))
 
+        # Switch to the 2nd window
+        self.windows.switch_to(windows[1].handle)
+        self.assertEquals(windows[1].handle, self.marionette.current_chrome_window_handle)
+
+        # Make the window unique by its loaded URL
```
chmanchester: I don't understand this comment. Window handles from the server are unique.

firefox_puppeteer/tests/test_windows.py
```diff
@@ -3,98 +3,181 @@
...
+        # Open two more windows
+        for index in range(0, 2):
+            self.marionette.execute_script(""" window.open(); """)
 
         windows = self.windows.all
-        self.assertEquals(len(windows), 3)
+        self.assertEquals(len(windows), len(self.marionette.chrome_window_handles))
 
+        # Switch to the 2nd window
+        self.windows.switch_to(windows[1].handle)
```
chmanchester: Assert the length is what you want before indexing directly.

firefox_puppeteer/tests/test_windows.py
```diff
@@ -3,98 +3,181 @@
...
-            with self.marionette.using_context('content'):
-                return self.marionette.get_url() == url
+        def find_by_url(win):
+            with win.marionette.using_context('content'):
+                return win.marionette.get_url() == url
 
-        self.windows.switch_to(is_my_window)
-        self.assertEquals(windows[0].handle,
-                          self.marionette.current_chrome_window_handle)
+        self.windows.switch_to(find_by_url)
```
chmanchester: Finding by url doesn't totally make sense, because a window can have multiple tabs, right? Maybe this will be clear in a follow up about tabs.

firefox_puppeteer/ui/windows.py
```diff
@@ -0,0 +1,479 @@
...
+        :returns: The `window handle` of the focused chrome window.
+        """
+        with self.marionette.using_context('chrome'):
+            return self.marionette.execute_script("""
+              Cu.import("resource://gre/modules/Services.jsm");
+              var win = Services.wm.getMostRecentWindow("");
+              return win.QueryInterface(Ci.nsIInterfaceRequestor)
+                        .getInterface(Ci.nsIDOMWindowUtils)
+                        .outerWindowID.toString();
+            """)
```
chmanchester: Oh no! Please let's try to find a way around this. This is exploiting an implementation detail of the server that we really might need to change in the future.

firefox_puppeteer/ui/windows.py
```diff
@@ -0,0 +1,479 @@
...
+
+        :param exceptions: Optional, list of handles or :class:`BaseWindow` instances
+        not to close
+        """
+        handles_to_keep = exceptions or []
+        handles_to_keep = [entry.handle if isinstance(entry, BaseWindow) else entry
+                           for entry in exceptions]
+
+        for handle in self.marionette.chrome_window_handles:
+            if handle not in handles_to_keep:
```
chmanchester: Would use a set() here.

firefox_puppeteer/ui/windows.py
```diff
@@ -0,0 +1,479 @@
...
+        """Focuses the chrome window with the given handle.
+
+        :param handle: The handle of the chrome window
+        """
+        self.switch_to(handle)
+
+        with self.marionette.using_context('chrome'):
+            self.marionette.execute_script(""" window.focus(); """)
+
+        wait = Wait(self.marionette, timeout=5, interval=.1)
```
chmanchester: I think these parameters are the same as the defaults.

firefox_puppeteer/ui/windows.py
```diff
@@ -0,0 +1,479 @@
...
+        wait.until(lambda m: handle == self.focused_chrome_window_handle)
```
chmanchester: This will be true as soon as switch_to returns. Are you anticipating we'll need this in a later version using events or observers?

firefox_puppeteer/ui/windows.py
```diff
@@ -0,0 +1,479 @@
...
+
+    def switch_to(self, target):
+        """Switches context to the specified chrome window.
+
+        :param target: The window to switch to. `target` can be a `handle` or a
+                       callback that returns True in the context of the desired
+                       window.
+
+        :returns: The old `window_handle`. This makes it easy to switch back
+                  to the original window later.
```
chmanchester: This comment or the implementation is out of sync.

firefox_puppeteer/ui/windows.py
```diff
@@ -0,0 +1,479 @@
...
+            window_type = self.marionette.get_window_type()
+        finally:
+            # Ensure to switch back to the original window
+            if handle != current_handle:
+                self.marionette.switch_to_window(current_handle)
+
+        window = BaseWindow(lambda: self.marionette, handle)
+
+        if window_type == 'navigator:browser':
+            window.__class__ = BrowserWindow
```
chmanchester: Can't you just construct a BrowserWindow _or_ a BaseWindow conditionally?

firefox_puppeteer/ui/windows.py
```diff
@@ -0,0 +1,479 @@
...
+        target_handle = None
+
+        if target in self.marionette.chrome_window_handles:
+            target_handle = target
+        elif callable(target):
+            current_handle = self.marionette.current_chrome_window_handle
+
+            # switches context if callback for a chrome window returns `True`.
+            for handle in self.marionette.chrome_window_handles:
+                self.marionette.switch_to_window(handle)
```
chmanchester: It doesn't seem like switching to each is necessary as a part of the search.

firefox_puppeteer/ui/windows.py
```diff
@@ -0,0 +1,479 @@
...
+                window = self.create_window_instance(handle)
+                if target(window):
+                    target_handle = handle
+                    break
+
+            # if no handle has been found switch back to oridinal window
```
chmanchester: Typo: "original"

firefox_puppeteer/ui/windows.py
```diff
@@ -0,0 +1,479 @@
...
+            if not target_handle:
+                self.marionette.switch_to_window(current_handle)
+
+        if target_handle is None:
+            raise errors.NoSuchWindowException("No window found for '{}'"
+                                               .format(target))
+
+        # only switch if necessary
+        if target_handle != self.marionette.current_chrome_window_handle:
+            self.marionette.switch_to_window(target)
```
chmanchester: I think this needs to be self.marionette.switch_to_window(target_handle). We should have a test that fails this.

chmanchester: Oh, I see. This is never true because we switch in to each one in the loop. Let's rework that.

firefox_puppeteer/ui/windows.py
```diff
@@ -0,0 +1,479 @@
...
+        self._inner_window = self.marionette.find_element(By.CSS_SELECTOR, ':root')
+        self._l10n = L10n(self.get_marionette)
+        self._windows = Windows(self.get_marionette)
+
+    def __eq__(self, other):
+        return self.handle is not None and (self.handle == other.handle)
+
+    @property
+    def closed(self):
+        """Returns closed state of the crhome window
```
chmanchester: Typo: "chrome".

firefox_ui_harness/runtests.py
```diff
@@ -16,7 +16,7 @@
 
 class ReleaseTestRunner(BaseMarionetteTestRunner):
     extra_prefs = {
-        "browser.tabs.remote.autostart": True,
+        "browser.tabs.remote.autostart": False,
```
chmanchester: I really hope we don't end up regretting this.

firefox_ui_harness/testcase.py
```diff
@@ -24,15 +24,19 @@ def setUp(self, *args, **kwargs):
...
-                         "A test must not leak window handles. "
-                         "This test started the browser with %s open "
-                         "top level browsing contexts, but ended with %s." %
-                         (self._start_handle_count, win_count))
+        # win_count = len(self.marionette.window_handles)
+        # self.assertEqual(win_count, self._start_handle_count,
+        #                 "A test must not leak window handles. "
+        #                 "This test started the browser with %s open "
+        #                 "top level browsing contexts, but ended with %s." %
+        #                 (self._start_handle_count, win_count))
```
chmanchester: What's this about?

firefox_ui_tests/functional/private_browsing/test_about_private_browsing.py
```diff
@@ -12,8 +14,7 @@ class TestAboutPrivateBrowsing(FirefoxTestCase):
 
     def setUp(self):
         FirefoxTestCase.setUp(self)
-        self.pb_url = self.marionette.absolute_url('private_browsing/'
-                                                   'about.html?')
+        self.pb_url = self.marionette.absolute_url('private_browsing/about.html?')
```
chmanchester: Please don't keep a lot of unrelated changes around. It makes reviewing a lot harder.

firefox_puppeteer/tests/test_windows.py
```diff
@@ -3,98 +3,181 @@
...
-        self.assertEquals(first_window.window,
-                          self.marionette.find_element(By.CSS_SELECTOR,
-                                                       ':root'))
+        self.assertFalse(self.browser.closed)
 
-        # This action will fail if the current window is not a browser window
-        first_window.send_shortcut('ctrl-n')
+    def test_base_window_open_close(self):
+        win1 = self.windows.current
+        win1.__class__ = BaseWindow  # force BaseWindow instance
```
chmanchester: I don't know how this is meant to work, but I'd like to find another way to achieve this if we can.

firefox_puppeteer/__init__.py
whimboo: I thought about this first too but abandoned this idea because it would not be available for sub modules in that package. Something I would like to have is that modules would be able to access API classes and properties like those. Therefor the puppeteer modules need to be refactored so it only holds API level stuff. I would like to do that later.

firefox_puppeteer/api/keys.py
whimboo: I know and I have to fix the tests to make use of `self.browser.send_shortcut()`. This work is part of the todo items I added to the feedback request on Bugzilla.

firefox_puppeteer/errors.py
whimboo: I know but I do not see another option here to easily extend the Marionette errors. I would like that our tests only have to import this module and have access to all available exception classes. Do you have any other option in mind?

firefox_puppeteer/tests/test_windows.py
whimboo: The comment doesn't match any code here. So its a left-over from the very first experiments I think.

firefox_puppeteer/tests/test_windows.py
whimboo: I reverted line 23 to have a hard-check against 3.

firefox_puppeteer/tests/test_windows.py
whimboo: That's a left-over from Andrew's work. I will flag it with a TODO item for tabs.

firefox_puppeteer/ui/windows.py
whimboo: Can you please be more specific here? I don't get what you mean here. Marionette doesn't offer us a way to get the focused element including window handles, so I do not see another way to get this implemented.

firefox_puppeteer/ui/windows.py
whimboo: What would be the benefits of a set() here?

firefox_puppeteer/ui/windows.py
whimboo: Not completely. The timeout seems to be needed here given that I can see that in the marionette client docs: `marionette.Wait(marionette, timeout=None, interval=0.1, ...)`

firefox_puppeteer/ui/windows.py
whimboo: switch_to() does not set the focus to the newly selected chrome window or tab. The focus will still be on the old handle. Otherwise not sure what you are asking here.

firefox_puppeteer/ui/windows.py
whimboo: Yes, one other TODO item as listed on Bugzilla.

firefox_puppeteer/ui/windows.py
whimboo: The test should basically fail for an unknown window. But I think we can do that. There is really no need to have an instance of a BaseWindow which is more like an abstract class. I will move the instantiation into the if block. Thanks.

firefox_puppeteer/ui/windows.py
whimboo: You have to when you want to run the callback against the actual chrome window. Otherwise it would not work on the target window but the currently selected one, and any kind of element retrieval would fail.

firefox_puppeteer/ui/windows.py
whimboo: Ups, good catch! Changing variable names can treat left-overs.

whimboo: The current code only enters this block for a handle as parameter. But given the naming this line would fall back to the parameter name which will be identical. So no way to get this tested.

firefox_ui_harness/runtests.py
whimboo: Not until end of next week. Maybe one of the upcoming e10s fixes might solve it until then.

firefox_ui_harness/testcase.py
whimboo: Just for testing while I was messing around with open windows etc. It will be reverted or even improved with the final patch.

firefox_puppeteer/tests/test_windows.py
whimboo: Ok, I figured out a way better way. As you say this was most likely not the best way given that properties of the browser window might still be around.

whimboo:
A commit for the review comments and fixes for functional tests have been pushed. The remaining todos will get their own commits.

firefox_puppeteer/ui/windows.py
chmanchester: "handles" from the server are intended to be opaque, but you're using this to compare against a handle from the server, and it works because this is how it's implemented. My conclusion from a recent issue with the server was that for window handles corresponding to tabs we need to re-implement then not to use outerWindowID, but a different ID we generate ourselves. If the same ends up being true for chrome windows, this will break.

firefox_puppeteer/ui/windows.py
chmanchester: My assumption is that the runtime of the "in" operator inside the loop is a linear function of the size of the list, and we go over every handle, so this loop is worst case polynomial runtime. The upper bound on the count of handles is probably low enough in practice that this will not be a problem, but it's probably good to avoid polynomial runtime algorithms when it's easy to do so!

firefox_puppeteer/ui/windows.py
chmanchester: I probably don't understand exactly how this will end up being used, but the callback could switch, right?

firefox_puppeteer/api/keys.py
chmanchester: But a neighboring library won't have self.browser, and I made use of this functionality in a toolbars class.

firefox_puppeteer/errors.py
chmanchester: I would leave them separate. If we need our own exceptions, that's fine, but marionette's always correspond to something directly from the protocol its using to talk to the browser, accessing them from the same place might just cause confusion.

firefox_puppeteer/ui/windows.py
whimboo: As what I know from Mozmill the outer window id is unique for chrome windows. But I do not see another way right now how to get this implemented. Marionette doesn't give us a way. So not sure what else we should do here.

firefox_puppeteer/ui/windows.py
whimboo: So you mean I should create a set via set(chrome_window_handles) - set(handles_to_keep) and close all those remaining chrome windows as in the resulting set?

chmanchester: Or just convert handles_to_keep outside of the loop.

firefox_puppeteer/ui/windows.py
whimboo: It could but then you would have to add this all the time. You really want to find a specific property of the chrome window, so a switch has to happen anyway.

For a use case see the following test. It's not ideal yet and will have to be updated for sure once the tabs module lands.

https://github.com/mozilla/firefox-ui-tests/pull/50/files#diff-ca6f0d344ce27b0ea7ac42051609281aR42

firefox_puppeteer/api/keys.py
whimboo: I see, so we have to do the refactoring for keyboard shortcuts later. I will bring this module and the docs back

firefox_puppeteer/ui/windows.py
chmanchester: This might be ok for the short term, we might need to add something to marionette that lets us get a reference to an actual window element from a handle if this goes wrong in the future.

firefox_puppeteer/tests/test_windows.py
```diff
@@ -3,98 +3,189 @@
...
-                          self.marionette.current_chrome_window_handle)
 
-        def is_my_window(win):
-            with self.marionette.using_context('content'):
-                return self.marionette.get_url() == url
+        # TODO: A window can have multiple tabs, so this may need an update
+        # when the tabs module gets implemented
+        def find_by_url(win):
+            with win.marionette.using_context('content'):
+                return win.marionette.get_url() == url
```
chmanchester: Who is win.marionette? This is confusing, there's is only one marionette.

firefox_puppeteer/ui/windows.py
chmanchester: I still think it's preferable to switch in the callback, because switching windows can end up doing actual work (like "activating" marionette in a newly seen window), and a caller might not expect to go through that for every open window when switching. I'll leave this to your discretion.

firefox_puppeteer/errors.py
whimboo: As discussed on IRC we will keep them separate then. The fix will be part of one of the next commits.

firefox_puppeteer/tests/test_windows.py
whimboo: The callbacks have a parameter, which is the actual BaseWindow instance to operate on. Keep in mind that switch_to() iterates over all the existent chrome windows. Once we have the tabs module it will most likely look like `win.tabbar.location == url`. But we might wanna find a better test here.

firefox_puppeteer/ui/windows.py
whimboo: That's what we discussed on Monday and we agreed on that necessary switching should happen inside the ui modules. I still want to remind that "activating" does not bring the window into the foreground aka focus(). The window state keeps as it is. So UI wise nothing is changing.

firefox_puppeteer/tests/test_windows.py
chmanchester: I just mean that there's only one marionette, and it only relates to one browser state  at a point in time. So we're saying win.marionette, but it's calling the same method as if we'd said self.marionette, but it sort of implies that we meaningfully have a marionette per window.

whimboo: Please see my example above how it will look like later. You won't see that we call something via marionette in the future here. This is just to ensure we keep the right context and do not switch between global and local scope. Exactly that would be confusing.

firefox_puppeteer/ui/windows.py
```diff
@@ -0,0 +1,486 @@
...
+
+        There is an optional `exceptions` list, which can be used to exclude
+        specific chrome windows from being closed.
+
+        :param exceptions: Optional, list or a single entry of handles or
+         :class:`BaseWindow` instances not to close
+        """
+        handles_to_keep = exceptions or []
+        if not isinstance(handles_to_keep, list):
+            handles_to_keep = [handles_to_keep]
```
chmanchester: Sigh. I really don't like APIs that let you have multiple input types, but I know from prior debates that I'm probably in the minority. I'll leave it to your discretion.

firefox_puppeteer/ui/windows.py
```diff
@@ -0,0 +1,486 @@
...
+            if handle != current_handle:
+                self.marionette.switch_to_window(current_handle)
+
+        if window_type == 'navigator:browser':
+            window = BrowserWindow(lambda: self.marionette, handle)
+        else:
+            raise errors.UnknownWindowError('Unknown window type "%s" for handle: "%s"' %
+                                            (window_type, handle))
+
+        if expected_class is not None and type(window) is not expected_class:
```
chmanchester: Something about this doesn't work out to me. We only ever allow creation of BrowserWindow s in this method, so if we ever have expected_class set to something other than BrowserWindow it would fail. Can't individual tests just assert what kind of window type they expect?

firefox_puppeteer/ui/windows.py
```diff
@@ -0,0 +1,486 @@
...
+
+        :param handle: The handle of the chrome window
+        """
+        self.switch_to(handle)
+
+        with self.marionette.using_context('chrome'):
+            self.marionette.execute_script(""" window.focus(); """)
+
+        wait = Wait(self.marionette, timeout=5)
+        wait.until(lambda m: handle == self.focused_chrome_window_handle)
```
chmanchester: The default timeout is 5.

firefox_puppeteer/ui/windows.py
chmanchester: We didn't discuss switching to every window in a loop as a part of a search. Looking at the example in a test above, I think this leads to tests that are harder to understand, because I need to understand this loop to understand how the callback ever evaluates to true. I don't agree with this API, but I will leave it to your discretion.

firefox_puppeteer/ui/windows.py
```diff
@@ -0,0 +1,486 @@
...
+                    target_handle = handle
+                    break
+
+            # if no handle has been found switch back to original window
+            if not target_handle:
+                self.marionette.switch_to_window(current_handle)
+
+        if target_handle is None:
+            raise NoSuchWindowException("No window found for '{}'"
+                                        .format(target))
```
chmanchester: You can combine this if with the one directly before it.

firefox_puppeteer/ui/windows.py
```diff
@@ -0,0 +1,486 @@
...
+
+    def __init__(self, marionette_getter, window_handle):
+        BaseLib.__init__(self, marionette_getter)
+
+        self._handle = window_handle
+        self._l10n = L10n(self.get_marionette)
+        self._windows = Windows(self.get_marionette)
+
+        # If we don't access self.marionette here it will be None all the time??!
+        self.marionette
```
chmanchester: I ran into a bug like this. There's an issue in BaseLib. The fix is in my toolbars patch.

firefox_puppeteer/ui/windows.py
```diff
@@ -0,0 +1,486 @@
...
+
+    def __eq__(self, other):
+        return self.handle is not None and (self.handle == other.handle)
```
chmanchester: self.handle should never be None, right? Can you assert as much in the constructor?

firefox_puppeteer/ui/windows.py
```diff
@@ -0,0 +1,486 @@
...
+            'alt': Keys.ALT,
+            'cmd': Keys.COMMAND,
+            'ctrl': Keys.CONTROL,
+            'meta': Keys.META,
+            'shift': Keys.SHIFT,
+        }
+
+        # Append all to press modifier keys
+        keys = []
+        for modifier in kwargs.keys():
```
chmanchester: This can just be `modifier in kwargs`, although I would iterate over the keymap, and assert that each kwarg is in the map (I would hate for this to fail silently because I misspelled one of the kwargs).

firefox_ui_harness/testcase.py
```diff
@@ -24,15 +24,23 @@ def setUp(self, *args, **kwargs):
...
-                         "A test must not leak window handles. "
-                         "This test started the browser with %s open "
-                         "top level browsing contexts, but ended with %s." %
-                         (self._start_handle_count, win_count))
+        try:
+            # Marionette needs an existent window to be selected. Take the first
+            # browser window which usually should not have been closed
+            # TODO: We might have to make this more error prone in case the
+            # original window has been closed.
+            self.browser.switch_to()
```
chmanchester: This is fine, but keep in mind it doesn't save us if we're switched into a tab that gets closed.

firefox_ui_harness/testcase.py
```diff
@@ -24,15 +24,23 @@ def setUp(self, *args, **kwargs):
...
+            # so that in case of a failure, further tests will not run
+            # in a state that is more inconsistent than necessary.
+            win_count = len(self.marionette.window_handles)
+            self.assertEqual(win_count, self._start_handle_count,
+                             "A test must not leak window handles. "
+                             "This test started the browser with %s open "
+                             "top level browsing contexts, but ended with %s." %
+                             (self._start_handle_count, win_count))
+        finally:
+            MarionetteTestCase.tearDown(self, *args, **kwargs)
```
chmanchester: I support this finally block.

firefox_ui_tests/functional/private_browsing/test_about_private_browsing.py
```diff
@@ -34,17 +34,18 @@ def testCheckAboutPrivateBrowsing(self):
...
-            access_key = self.browser.get_localized_entity(
-                'privatebrowsingpage.openPrivateWindow.accesskey')
+            # TODO: Augment send_shortcut() to DOMElement and let Marionette return
+            # such an object instead of HTMLElement, so it can be send to the element
 
             # Send keys to the top html element.
-            top_html = self.marionette.find_element(By.TAG_NAME, 'html')
-            top_html.send_keys(self.keys.SHIFT, self.keys.ACCEL, access_key)
+            # top_html = self.marionette.find_element(By.TAG_NAME, 'html')
+            # top_html.send_keys(self.keys.SHIFT, self.keys.ACCEL, access_key)
```
chmanchester: Don't leave commented out code hanging around.

firefox_ui_tests/functional/private_browsing/test_about_private_browsing.py
```diff
@@ -34,17 +34,18 @@ def testCheckAboutPrivateBrowsing(self):
...
+            key = 'privatebrowsingpage.openPrivateWindow.accesskey'
+            self.browser.send_shortcut(self.browser.get_localized_entity(key),
+                                       accel=True, shift=True)
 
         self.wait_for_condition(lambda mn: len(self.windows.all) == 2)
-        self.windows.switch_to(lambda win: win.is_private)
-        self.browser_pb = self.windows.current
-
+        self.browser_pb = self.windows.switch_to(lambda win: win.is_private)
         self.assertTrue(self.browser_pb.is_private)
```
chmanchester: This assertion is redundant. Probably was before, but as long as we're here...

firefox_puppeteer/ui/windows.py
chmanchester: Actually, can we just select the :root element and then compare with Services.wm.getMostRecentWindow("") when we need to do this?

firefox_puppeteer/ui/windows.py
whimboo: I will file a bug so we can get this discussed and updated after the landing here.

firefox_puppeteer/ui/windows.py
whimboo: This method will get all the different kinds of windows. BrowserWindow is just the beginning, and more will follow soon. So I clearly see an advantage of having this check here at a single place and not scattered around ui libraries and tests.

firefox_puppeteer/ui/windows.py
whimboo: This is not true. It defaults to None:
https://dxr.mozilla.org/mozilla-central/source/testing/marionette/client/marionette/wait.py#30

But I see where the problem is, it needs an update of the documentation so it says 5s but not None. I assume this would be good for a first good bug.

firefox_puppeteer/ui/windows.py
whimboo: Chris, you cannot determine a specific value of e.g. an element if you do not switch to this window! As you said somewhere else the marionette object exists only once and is tied to a specific window. If you want to search for another window how would you retrieve e.g. the private browsing state? This simply won't work.

firefox_puppeteer/ui/windows.py
whimboo: No we cannot because we also have to check if the given handle is correct. If I would move this code into the callback block, we would try to operate on an invalid handle. Also I do not want to move out the inner if block because thats only related to the callback block. For a normal handle we do not have to switch before.

firefox_puppeteer/ui/windows.py
whimboo: Hurray!! Good to hear that. Thanks

firefox_puppeteer/ui/windows.py
whimboo: Makes sense. I will also check for the marionette_getter to be callable.

firefox_puppeteer/ui/windows.py
whimboo: I would still like to keep iterating over the kwargs because `len(kwargs) <= len(keymap)`. But we should indeed check if it is a valid modifier even in case of being False.

firefox_ui_harness/testcase.py
whimboo: This case would work because we switch back to the chrome window. What will not work is if a test opens a new window, but closes the original browser window. It's clearly something to improve, but I would this push out until we have more knowledge in handling windows, tabs, and about puppeteer at all.

firefox_ui_tests/functional/private_browsing/test_about_private_browsing.py
whimboo: I wanted to leave it in here so I know what to revert later. I'm not sure right now if sending the access key to the window is the final right approach. Given that I brought back the keys module, I can revert that one for now.

firefox_ui_tests/functional/private_browsing/test_about_private_browsing.py
whimboo: Why is this assertion redundant? We expect to have a private browsing window, but if something fails in the switch_to() code, we will fail here. So I really do not see why this check is redundant because it tests that switch_to() works as expected.

firefox_puppeteer/ui/windows.py
whimboo: :root is a DOMWindow but not ChromeWindow instance! Both windows are completely different so you cannot compare nor use getMostRecentWindow with it. I would leave it as is, and maybe

whimboo:
@chmanchester I pushed a new commit with all the remaining fixes from your last review. Can you please have a look if that is fine? If yes, lets get it squashed and landed.

firefox_puppeteer/ui/windows.py
chmanchester: Just a thought. Please file a follow up.

firefox_ui_harness/testcase.py
chmanchester: Chrome windows and tabs are handled separately in the server. This doesn't work if we're switched into a tab that gets closed.

chmanchester:
Go ahead and merge it. We can leave anything else to a follow up.

PR #52, opened 2015-01-26
Locationbar

chmanchester:


whimboo:
Chris, lots of tests are failing. Please fix those failures first, then I will check that PR.

chmanchester:
This works locally, so I'm grasping at straws here, but my guess right now is that when I'm waiting for the state to be "open", querying that each time is stealing focus and causing the autcomplete results to go away.

chmanchester:
@whimboo, anything come to mind based on your experience with the mozmill equivalent of these libraries?

chmanchester:
It sort of seems like something else is stealing focus. I noticed in the screenshot I dumped we have a notification in view about e10s, I'll try turning that off.

firefox_ui_harness/decorators.py
```diff
@@ -13,3 +14,11 @@ def wrapper(self, *args, **kwargs):
             raise SkipTest("Skipping due to e10s")
         return target(self, *args, **kwargs)
     return wrapper
+
+
+def skip_under_xvfb(target):
+    def wrapper(self, *args, **kwargs):
+        if os.environ.get('MOZ_XVFB'):
+            raise SkipTest("Skipping due to travis CI")
```
whimboo: I wouldn't necessarily add the relation to travis here. Just say it is because of xvfb.

firefox_puppeteer/ui/browser_window.py
```diff
@@ -56,6 +56,14 @@ def tabbar(self):
...
         """
 
+    @use_class_as_property('ui.autocompleteresults.AutocompleteResults')
+    def autocompleteresults(self):
+        """Provides utility members for accessing and manipulationg the
+        locationbar.
+
+        See the :class:`~ui.locationbar.LocationBar` reference.
+        """
+
```
whimboo: The autocomplete results are not an element related to the browser, but for the location bar which is also the parent. So please move this to the LocationBar class.

firefox_puppeteer/ui/browser_window.py
```diff
@@ -38,7 +38,7 @@ def is_private(self):
             return PrivateBrowsingUtils.isWindowPrivate(chromeWindow);
         """, script_args=[self.window])
 
-    @use_class_as_property('ui.navbar.NavBar')
+    @use_class_as_property('ui.toolbars.NavBar')
```
whimboo: The changes here will have to be re-made once my window patch lands.

firefox_puppeteer/tests/test_toolbars.py
```diff
@@ -0,0 +1,132 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from marionette import Wait
+
+from firefox_ui_harness.testcase import FirefoxTestCase
+from firefox_ui_harness.decorators import skip_under_xvfb
```
whimboo: nit: exchange the lines to keep the order.

firefox_puppeteer/ui/toolbars.py
```diff
@@ -0,0 +1,337 @@
...
+        return self.marionette.find_element('id', 'PanelUI-menu-button')
+
+
+class LocationBar(BaseLib):
+    """Various utilities for interacting with the location bar (the text area
+    of the ui that typically displays the current url).
+    """
+
+    dtds = ["chrome://branding/locale/brand.dtd",
+            "chrome://browser/locale/browser.dtd"]
```
whimboo: We should not redefine all the DTDs in each class. Instead we should pass through a BaseWindow reference to each UI class. I'm happy to move this out to later but really file a new bug for it.

chmanchester: Do you mean have every UI library inherit from BaseWindow? That sounds like a bad design!

firefox_puppeteer/ui/toolbars.py
```diff
@@ -0,0 +1,337 @@
...
+from ..api.keys import Keys
+from ..api.l10n import L10n
+from ..base import BaseLib
+from ..decorators import use_class_as_property
+
+
+class NavBar(BaseLib):
+
+    @use_class_as_property('ui.toolbars.LocationBar')
+    def locationbar(self):
```
whimboo: nit: lets sort class members alphabetically so they are also easier to find.

firefox_puppeteer/ui/toolbars.py
```diff
@@ -0,0 +1,337 @@
...
+    of the ui that typically displays the current url).
+    """
+
+    dtds = ["chrome://branding/locale/brand.dtd",
+            "chrome://browser/locale/browser.dtd"]
+
+    def __init__(self, *args, **kwargs):
+        BaseLib.__init__(self, *args, **kwargs)
+        # TODO: A "utility" module that sets up the client directly would be
+        # useful here.
```
whimboo: Or way better get rid of the delayed marionette getter/setter. That's one of the things I also want to see and which brings us closer to the Gaia UI tests.

firefox_puppeteer/ui/toolbars.py
```diff
@@ -0,0 +1,337 @@
...
+    @use_class_as_property('ui.toolbars.IdentityPopup')
+    def identity_popup(self):
+        """Provides utility members for accessing and manipulationg the
+        locationbar.
+
+        See the :class:`~ui.toolbars.IdentityPopup` reference.
+        """
+
+    @property
+    def urlbar(self):
```
whimboo: nit: please sort properties alphabetically. I think it will also apply to other lines of code.

firefox_puppeteer/ui/toolbars.py
```diff
@@ -0,0 +1,337 @@
...
+
+    @property
+    def stop_button(self):
+        return self.marionette.find_element(By.ID, 'urlbar-stop-button')
+
+    @property
+    def urlbar_input(self):
+        """:returns: The urlbar_input element
+        """
+        return self.urlbar.find_element('anon attribute', {'anonid': 'input'})
```
whimboo: Lovely how we can do that with Marionette now! I kinda like that.

firefox_puppeteer/ui/toolbars.py
```diff
@@ -0,0 +1,337 @@
...
+    @property
+    def history_drop_marker(self):
+        return self.urlbar.find_element('anon attribute', {'anonid': 'historydropmarker'})
+
+    @property
+    def contextmenu(self):
+        """:returns: The urlbar contextmenu element.
+        """
+        parent = self.urlbar.find_element('anon attribute', {'anonid': 'textbox-input-box'})
+        return parent.find_element('anon attribute', {'anonid': 'input-box-contextmenu'})
```
whimboo: I think you don't have to go through `textbox-input-box` here. It's at the same level as the `urlbar_input` property. So we should be able to find it directly via the urlbar.

firefox_puppeteer/ui/toolbars.py
```diff
@@ -0,0 +1,337 @@
...
+
+    @property
+    def notification_popup(self):
+        return self.marionette.find_element(By.ID, "notification-popup")
+
+    @property
+    def value(self):
+        """:returns: The currently displayed value of the urlbar.
+        """
```
whimboo: The first line contains a description of the property/element, while `:returns` should be in its own line. Same for all the others.

firefox_puppeteer/ui/toolbars.py
```diff
@@ -0,0 +1,337 @@
...
+        return self.urlbar.get_attribute('value')
+
+    def get_contextmenu_entry(self, prop, subtype):
+        """
+        :param subtype: The type of entry to retrieve.
+        :returns: The urlbar contextmenu entry.
+        """
+        entries = self.contextmenu.find_elements('css selector', 'menuitem')
```
whimboo: Actually this should use a Menu API. Please make a TODO comment here.

firefox_puppeteer/ui/toolbars.py
```diff
@@ -0,0 +1,337 @@
...
+    def notification_popup(self):
+        return self.marionette.find_element(By.ID, "notification-popup")
+
+    @property
+    def value(self):
+        """:returns: The currently displayed value of the urlbar.
+        """
+        return self.urlbar.get_attribute('value')
+
+    def get_contextmenu_entry(self, prop, subtype):
```
whimboo: I don't think we need the prop parameter here. It will always be `cmd`. `subtype` I would better rename to action. Maybe we can add a @see link to the appropriate xul file? Otherwise a list of possible actions would be good to know.

firefox_puppeteer/ui/toolbars.py
```diff
@@ -0,0 +1,337 @@
...
+        :param subtype: The type of entry to retrieve.
+        :returns: The urlbar contextmenu entry.
+        """
+        entries = self.contextmenu.find_elements('css selector', 'menuitem')
+        filter_on = 'cmd_%s' % subtype
+        found = [e for e in entries if e.get_attribute(prop) == filter_on]
+        return found[0] if len(found) else None
+
+    def close_context_menu(self):
+        self.contextmenu.send_keys(Keys.ESCAPE)
```
whimboo: Same TODO comment as above.

firefox_puppeteer/ui/toolbars.py
```diff
@@ -0,0 +1,337 @@
...
+
+    def clear(self):
+        """Clears the contents of the url bar (via the DELETE shortcut).
+
+        Note that the operation is synchronous: it does not return until
+        the value of the url bar is cleared.
```
whimboo: All of our methods are usually synchronous. I would say we highlight only asynchronous methods.

nit 'DELETE' button

firefox_puppeteer/ui/toolbars.py
```diff
@@ -0,0 +1,337 @@
...
+        """
+        self.focus('shortcut')
+        self.urlbar.send_keys(Keys.DELETE)
+        Wait(self.marionette).until(
+            lambda _: self.urlbar.get_attribute('value') == '')
+
+    def load_url(self, url):
```
whimboo: I would say we call this `navigate()` to keep at least a bit in sync with Marionette itself.

firefox_puppeteer/ui/toolbars.py
```diff
@@ -0,0 +1,337 @@
...
+        """Load the specified url in the location bar by synthesized
+        keystrokes.
+        """
+        self.focus('shortcut')
+        self.urlbar.send_keys(url + Keys.ENTER)
+
+    def reload_url(self, evt='shortcut', force=False):
```
whimboo: Instead of `evt` please use `trigger` so we are in sync with the window classes. Also I would keep the name of the method `reload`. I think it is clear enough to understand what it does.

firefox_puppeteer/ui/toolbars.py
```diff
@@ -0,0 +1,337 @@
...
+                    "shortcut", "shortcut2", or "button").
+        :param force: Whether to cause a forced reload.
+        """
+        # TODO: The force parameter is ignored for the moment. Use
+        # mouse event modifiers or actions when they're ready.
+        # Bug 1097705 tracks this feature in marionette.
+        if evt == 'button':
+            self.reload_button.click()
+        elif evt == 'shortcut':
+            cmd_key = self.l10n.get_localized_entity(LocationBar.dtds,
```
whimboo: nit: self.dtds.

firefox_puppeteer/ui/toolbars.py
whimboo: Given that we do not have a menu entry, we should better use `button` as default here. That's something we haven't updated in the mozmill module yet.

firefox_puppeteer/ui/toolbars.py
```diff
@@ -0,0 +1,337 @@
...
+        """
+        return self.marionette.find_element(By.ID,
+                                            'PopupAutoCompleteRichResult')
+
+    @property
+    def results(self):
+        """Returns the container node for autocomplete results.
+        """
+        return self.popup.find_element('anon attribute',
+                                       {'anonid': 'richlistbox'})
```
whimboo: fyi we do not have a limit of 80 chars anymore. You can go up to 100 chars now.

firefox_puppeteer/ui/toolbars.py
```diff
@@ -0,0 +1,337 @@
...
+
+    def is_open(self):
+        return self.popup.get_attribute('state') == 'open'
```
whimboo: This can become a property.

firefox_puppeteer/ui/toolbars.py
```diff
@@ -0,0 +1,337 @@
...
+
+    def get_visible_results(self):
```
whimboo: Make this a property too with the name `visible_results`

firefox_puppeteer/ui/toolbars.py
chmanchester: Ok, but now that this is written sorting them by hand seems like a horrendous use of time. Ctr-f works fine!

firefox_puppeteer/ui/toolbars.py
```diff
@@ -0,0 +1,337 @@
...
+        """Returns a list of visible result nodes."""
+        return self.marionette.execute_script("""
+          let rv = [];
+          let node = arguments[0];
+          let count = node.itemCount;
+          for (let i = 0; i < count; ++i) {
```
whimboo: nit: not really a need to define count here, just put node.itemCount into the brackets directly.

firefox_puppeteer/ui/toolbars.py
```diff
@@ -0,0 +1,337 @@
...
+            let item = node.getItemAtIndex(i);
+            if (!item.hasAttribute("collapsed")) {
+              rv.push(item);
+            }
+          }
+          return rv;
+        """, script_args=[self.results])
+
+    def close(self, force=False):
+        if self.is_open():
```
whimboo: Lets negate and return early

firefox_puppeteer/ui/toolbars.py
```diff
@@ -0,0 +1,337 @@
...
+            if force:
+                self.marionette.execute_script("""
+                  arguments[0].hidePopup();
+                """, script_args=[self.popup])
+            else:
+                (self.marionette.find_element('id', 'urlbar')
+                                .send_keys(Keys.ESCAPE))
```
whimboo: Once we have the reference to a browser window, this could be come `browser.navbar.locationbar.send_keys`

firefox_puppeteer/ui/toolbars.py
chmanchester: If you can figure out a way to lazily load library modules and have them available during setUp, by all means do it!

firefox_puppeteer/ui/toolbars.py
```diff
@@ -0,0 +1,337 @@
...
+        result in the urlbar.
+
+        :param result: The result to inspect for matches.
+        :param result_type: The type of match to search for (one of "title", "url").
+        """
+
+        assert match_type in ('title', 'url'), ('match_type provided must be one of'
+                                                '"title" or "url", not %s' % match_type)
+
+        if match_type == "title":
```
whimboo: nit: please stay with single quotes around all strings in Python.

firefox_puppeteer/ui/toolbars.py
```diff
@@ -0,0 +1,337 @@
...
+            descnode = self.marionette.execute_script("""
+              return arguments[0].boxObject.firstChild.childNodes[1].childNodes[0];
+            """, script_args=[result])
+        if match_type == "url":
+            descnode = self.marionette.execute_script("""
+              return arguments[0].boxObject.lastChild.childNodes[2].childNodes[0];
+            """, script_args=[result])
```
whimboo: There is only a single difference in the index. It would be better to also push the index into the execute_script() call, and have only one such thing.

firefox_puppeteer/ui/toolbars.py
```diff
@@ -0,0 +1,337 @@
...
+    def get_matching_text(self, result, match_type):
+        """Retuns an array of strings of the matching text within a autocomplete
+        result in the urlbar.
+
+        :param result: The result to inspect for matches.
+        :param result_type: The type of match to search for (one of "title", "url").
+        """
+
+        assert match_type in ('title', 'url'), ('match_type provided must be one of'
+                                                '"title" or "url", not %s' % match_type)
```
whimboo: I would prefer a else case in the following if block.

firefox_puppeteer/ui/toolbars.py
chmanchester: I can't get this to work without accessing it from its direct parent.

firefox_puppeteer/ui/toolbars.py
```diff
@@ -0,0 +1,337 @@
...
+            if (node.nodeName == 'span') {
+              rv.push(node.innerHTML);
+            }
+          }
+          return rv;
+        """, script_args=[descnode])
+
+
+class IdentityPopup(BaseLib):
+    """Library wrappping selectors for interacting with the identity popup.
```
whimboo: spelling mistake -p

firefox_puppeteer/ui/toolbars.py
```diff
@@ -0,0 +1,337 @@
...
+    """
+
+    def is_open(self):
```
whimboo: a property please.

firefox_puppeteer/ui/toolbars.py
chmanchester: This module is a very thing wrapper around a set of selectors. I'm having trouble coming up with useful things to say for the sake of this convention.

firefox_puppeteer/tests/test_toolbars.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+class TestLocationBar(FirefoxTestCase):
+
+    def test_reload(self):
+        event_types = ["shortcut", "shortcut2", "button"]
+        for event in event_types:
+            # TODO: Until we have waitForPageLoad, this only tests API
```
whimboo: I assume there is no such thing in Marionette what we could use? I was hoping that we get it for free. :/

firefox_puppeteer/tests/test_toolbars.py
```diff
@@ -0,0 +1,132 @@
...
+            # compatibility.
+            self.browser.navbar.locationbar.reload_url(event, True)
+            self.browser.navbar.locationbar.reload_url(event, False)
```
whimboo: Please use force=foobar here.

firefox_puppeteer/tests/test_toolbars.py
```diff
@@ -0,0 +1,132 @@
...
+
+    def test_focus_and_clear(self):
+        locationbar = self.browser.navbar.locationbar
+        locationbar.urlbar.send_keys("zyx")
```
whimboo: I think that this is obsolete.

firefox_puppeteer/ui/toolbars.py
chmanchester: Alright, but this reflects the current mozmill schema. For those of us converting libraries, how are we supposed to know when the mozmill libraries are incorrect?

firefox_puppeteer/tests/test_toolbars.py
```diff
@@ -0,0 +1,132 @@
...
+        self.assertEqual(locationbar.value, 'zyx')
+        locationbar.clear()
+        self.assertEqual(locationbar.value, '')
+
+    def test_load_url(self):
+        data_uri = 'data:text/html,<title>Title</title>'
+        locationbar = self.browser.navbar.locationbar
+        locationbar.clear()
+        locationbar.load_url(data_uri)
```
whimboo: I think that load_url() should clear the location bar value itself. That is something we never fixed in our Mozmill module.

firefox_puppeteer/ui/toolbars.py
chmanchester: I disagree. The behavior is really different, I would be confused to find out this navigate differed from marionette's.

firefox_puppeteer/tests/test_toolbars.py
```diff
@@ -0,0 +1,132 @@
...
+
+        with self.marionette.using_context('content'):
+            Wait(self.marionette).until(
+                lambda mn: mn.get_url() == data_uri)
+
+    def test_urlbar_input(self):
+        urlbar_input = self.browser.navbar.locationbar.urlbar_input
+        self.assertEqual('input', urlbar_input.get_attribute('localName'))
+        self.assertIn('urlbar-input', urlbar_input.get_attribute('className'))
```
whimboo: I don't think this assert helps us. We get this element via this class, so if it fails the former assert will fail too.

firefox_puppeteer/ui/toolbars.py
chmanchester: "reload" is a Python builtin, I don't want to shadow it.

firefox_puppeteer/tests/test_toolbars.py
```diff
@@ -0,0 +1,132 @@
...
+        contextmenu = self.browser.navbar.locationbar.contextmenu
+        self.assertEqual('menupopup', contextmenu.get_attribute('localName'))
+
+    def test_contextment_entry(self):
+        contextmenu_entry = self.browser.navbar.locationbar.get_contextmenu_entry('cmd', 'paste')
+        self.assertEqual('cmd_paste', contextmenu_entry.get_attribute('cmd'))
+
+    def test_reloadbutton(self):
+        reload_button = self.browser.navbar.locationbar.reload_button
+        self.assertEqual('Browser:ReloadOrDuplicate', reload_button.get_attribute('command'))
```
whimboo: Please stay with the type of element as for the former property checks. We don't want to fail if internal method names are changing.

firefox_puppeteer/ui/toolbars.py
chmanchester: It's a class variable, so I think this is clearer, and an individual instance isn't likely to shadow it and change the behavior.

firefox_puppeteer/tests/test_toolbars.py
```diff
@@ -0,0 +1,132 @@
...
+        favicon = self.browser.navbar.locationbar.favicon
+        self.assertEqual('image', favicon.get_attribute('localName'))
+
+    def test_history_drop_marker(self):
+        drop_marker = self.browser.navbar.locationbar.history_drop_marker
+        self.assertEqual('dropmarker', drop_marker.get_attribute('localName'))
+
+    def test_stopbutton(self):
+        stop_button = self.browser.navbar.locationbar.stop_button
+        self.assertEqual('toolbarbutton', stop_button.get_attribute('localName'))
```
whimboo: I wonder if we can combine all those individual test methods into a single method.

firefox_puppeteer/tests/test_toolbars.py
```diff
@@ -0,0 +1,132 @@
...
+    @skip_under_xvfb
+    def test_popup_elements(self):
+        # TODO: This test is not very robust because it relies on the history
+        # in the default profile.
+        autocompleteresults = self.browser.navbar.locationbar.autocomplete_results
+        self.assertFalse(autocompleteresults.is_open())
+        self.browser.navbar.locationbar.urlbar.send_keys('a')
+        results = autocompleteresults.results
+        Wait(self.marionette).until(
+            lambda _: autocompleteresults.is_open())
```
whimboo: Can't we use self.wait_for_condition() here and all the other places?

firefox_puppeteer/tests/test_toolbars.py
```diff
@@ -0,0 +1,132 @@
...
+        visible_result_count = len(autocompleteresults.get_visible_results())
+        self.assertTrue(visible_result_count > 0)
+        self.assertEqual(visible_result_count,
+                         int(results.get_attribute('itemCount')))
```
whimboo: Lets make sure you also close the popup in teardownTest.

firefox_puppeteer/tests/test_toolbars.py
```diff
@@ -0,0 +1,132 @@
...
+            lambda _: autocompleteresults.is_open())
+        visible_results = autocompleteresults.get_visible_results()
+        self.assertTrue(len(visible_results) > 0)
+        for result in visible_results:
+            title_matches = autocompleteresults.get_matching_text(result, "title")
+            url_matches = autocompleteresults.get_matching_text(result, "url")
+            all_matches = title_matches + url_matches
+            self.assertTrue(len(all_matches) > 0)
+            for match_fragment in all_matches:
+                self.assertIn(match_fragment, (input_text, input_text.upper()))
```
whimboo: This test is mostly the same as `test_popup_elements`. I would suggest we get rid of the other.

firefox_puppeteer/ui/toolbars.py
chmanchester: There's also the firstChild/lastChild difference. I can't think of a cleaner way to this off the top of my head.

whimboo:
Please also make sure to update the docs and add appropriate files.

firefox_puppeteer/tests/test_toolbars.py
chmanchester: Obsolete how?

firefox_puppeteer/tests/test_toolbars.py
chmanchester: This is directly from the mozmill version.

firefox_puppeteer/tests/test_toolbars.py
chmanchester: Again from the mozmill version. I can delete the test as you see fit.

firefox_puppeteer/tests/test_toolbars.py
chmanchester: Yeah, this is tedious. On the bright side the failures will be easier to debug.

firefox_puppeteer/tests/test_toolbars.py
chmanchester: Indeed it is equivalent!

firefox_puppeteer/tests/test_toolbars.py
chmanchester: This is a test for get_matching_text. It's a little redundant, but we'll know which part is broken if this test fails and the other does not.

chmanchester:
Whoops, teardown dies when we don't run the test... I'll push a fixup.

firefox_puppeteer/ui/toolbars.py
whimboo: Great. I filed bug 1127786 for the Puppeteer refactoring.

firefox_puppeteer/ui/toolbars.py
whimboo: Strange. Ok, so lets leave it in for now.

firefox_puppeteer/ui/toolbars.py
whimboo: There is no problem if some text is duplicated. For having a good documentation we should make sure we follow the guidelines. Here an example:

""" Returns the currently displayed text of the location bar.

:returns: Displayed text

firefox_puppeteer/ui/toolbars.py
whimboo: What about passing in the match_type into execute_script as an additional argument? That way we have the js code located right next to each other.

firefox_puppeteer/tests/test_toolbars.py
whimboo: You just want to reset the value here, a check if send_keys and clear works correct comes some lines below.

firefox_puppeteer/tests/test_toolbars.py
whimboo: So leave it then if you think it's worth.

firefox_puppeteer/tests/test_toolbars.py
whimboo: Yes, please remove it, but test for the localName instead.

firefox_puppeteer/tests/test_toolbars.py
whimboo: I'm not so convinced but you can leave it in. I just don't want that we have a couple of minutes for unit test runs before the tests you want to run are getting executed. It's just waging in necessity vs. time safer.

firefox_puppeteer/tests/test_toolbars.py
```diff
@@ -0,0 +1,135 @@
...
+class TestAutoCompleteResults(FirefoxTestCase):
+    def setUp(self):
+        FirefoxTestCase.setUp(self)
+        self.browser.navbar.locationbar.clear()
+
+    def tearDown(self):
+        try:
+            autocompleteresults = self.browser.navbar.locationbar.autocomplete_results
+            autocompleteresults.close(force=True)
+        except NoSuchElementException:
```
whimboo: Where does this fail exactly? in close()? I think we should make this failure proof so it only operates on the elemnet if it really exists, especially in the case of an exception.

chmanchester: It fails to find the autocomplete results element. I'm not sure what's wrong, but it only happens when the test doesn't run.

firefox_puppeteer/ui/toolbars.py
chmanchester: I'll push a new commit with docs and docstrings in order.

chmanchester: I do think that redundant documentation is problematic. I think I'd be a lot less likely to trust it or bother to keep it up to date.

chmanchester:
@whimboo This is updated and I think we've resolved the main issues. I'll squash and rebase once you mark r+ in bugzilla.

firefox_puppeteer/tests/test_toolbars.py
whimboo: Do you have a complete traceback for it?

chmanchester: I'm sure you could find one in travis.

firefox_puppeteer/docs/index.rst
```diff
@@ -32,15 +32,10 @@ Libraries
 The following libraries are currently implemented. More will be added in the
 future. Each library is available from an instance of the FirefoxTestCase class.
 
-.. autoclass:: Puppeteer
-   :members:
```
whimboo: Why do you remove the Puppeteer API reference documentation?

firefox_puppeteer/tests/test_toolbars.py
whimboo: Any feedback here Chris?

firefox_puppeteer/docs/index.rst
chmanchester: It makes it very confusing because some libraries are exposed from the base Puppeteer and other aren't. I don't think this needs to be a part of the user facing documentation.

firefox_puppeteer/tests/test_toolbars.py
chmanchester: Nope.

firefox_puppeteer/tests/test_toolbars.py
whimboo: Travis removes old builds automatically. So I will not be able to find it there. So please add a TODO comment here, so we can check that later.

firefox_puppeteer/docs/index.rst
whimboo: We only expose most-used API modules here. We don't need any ui module on that class.

PR #56, opened 2015-01-31
Bug 1127831 - Implement handling of tabs

whimboo:
Currently this is a WIP for tab handling. I will give more details on the listed bug.


firefox_puppeteer/base.py
```diff
@@ -2,10 +2,11 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
+import errors
```
chmanchester: Should this be a relative import?

firefox_puppeteer/base.py
```diff
@@ -2,10 +2,11 @@
...
+
 
 class BaseLib(object):
-    """A trivial base class that handles lazily setting the "client" class
-    attribute."""
+    """A base class that handles lazily setting the "client" class attribute."""
```
chmanchester: This change is unrelated to the pull request and increases line length.

firefox_puppeteer/base.py
```diff
@@ -22,3 +23,21 @@ def marionette(self):
...
+
+
+class UIBaseLib(BaseLib):
+    """A base class for all UI element wrapper classes inside a chrome window."""
+
+    def __init__(self, marionette_getter, window):
+        BaseLib.__init__(self, marionette_getter)
+
+        # importing globally doesn't work
+        from .ui.windows import BaseWindow
```
chmanchester: Local import statements have global side effects, so I'd rather omit this check (or convert to something like a comparison against type(window).__name__) and file a follow up to get this figured out.

firefox_puppeteer/tests/manifest.ini
```diff
@@ -1,6 +1,8 @@
 [test_l10n.py]
+disabled = Currently broken because tabbar cannot be found in tearDown
```
chmanchester: I have no problem with this when running locally on mac or linux.

firefox_puppeteer/tests/test_tabs.py
```diff
@@ -2,78 +2,152 @@
...
 from firefox_ui_harness.testcase import FirefoxTestCase
 
 
-class TestTabs(FirefoxTestCase):
+class TestTabBar(FirefoxTestCase):
+
+    def test_basics(self):
+        tabbar = self.browser.tabbar
+
+        self.assertEqual(tabbar.window, self.browser)
```
chmanchester: I think this reveals a problem with the overall strategy of attaching generic functionality to a window object and pushing the reference through to subclasses. This relationship is circular, and makes the idea of a "window" in our code as reflecting a ui element less meaningful.

firefox_puppeteer/tests/test_tabs.py
```diff
@@ -2,78 +2,152 @@
...
+        # Open a new tab by each trigger method
+        open_strategies = ('button',
+                           'menu',
+                           'shortcut',
+                           lambda tab: tabbar.newtab_button.click()
+                           )
+        for trigger in open_strategies:
+            new_tab = tabbar.open_tab(trigger=trigger)
+            self.assertEqual(len(tabbar.tabs), 2)
+            self.assertEqual(new_tab.handle, self.marionette.current_window_handle)
```
chmanchester: This could use an assertion on the lines of "self.assertEqual(new_tab.handle, tabbar.tabs[1].handle)"

firefox_puppeteer/tests/test_tabs.py
```diff
@@ -2,78 +2,152 @@
...
+
+            tabbar.close_tab()
+            self.assertEqual(len(tabbar.tabs), 1)
+            self.assertNotEqual(new_tab.handle, self.marionette.current_window_handle)
```
chmanchester: I think this would make more sense as if it made the assertion that current_window_handle is the first tab on the tab bar.

firefox_puppeteer/tests/test_tabs.py
```diff
@@ -2,78 +2,152 @@
...
+                            'shortcut',
+                            lambda tab: tab.close_button.click())
+        for trigger in close_strategies:
+            new_tab = tabbar.open_tab()
+            self.assertEqual(len(tabbar.tabs), 2)
+            self.assertEqual(new_tab.handle, self.marionette.current_window_handle)
+
+            tabbar.close_tab(trigger=trigger)
+            self.assertEqual(len(tabbar.tabs), 1)
+            self.assertNotEqual(new_tab.handle, self.marionette.current_window_handle)
```
chmanchester: Same comment above about the assertions in this loop.

firefox_puppeteer/ui/tabbar.py
```diff
@@ -3,126 +3,315 @@
...
 from .. import DOMElement
-from ..base import BaseLib
+from ..base import UIBaseLib
 
 
-class Tabs(BaseLib):
+class TabBar(UIBaseLib):
 
-    # TODO arrow scrollers, drop down list
+    # Properties for visual elements of the tabs toolbar #
```
chmanchester: This should be a docstring.

firefox_puppeteer/ui/tabbar.py
```diff
@@ -3,126 +3,315 @@
...
+        :returns: :class:`~ui.menu.MenuPanel` instance
         """
-        return (self.marionette.find_element('id', 'tabbrowser-tabs')
-                               .find_element('anon attribute',
-                                             {'anonid': 'tabs-newtab-button'}))
+        return MenuPanel(lambda: self.marionette, self.window)
 
     @property
-    def active_tab(self):
+    def newtab_button(self):
```
chmanchester: These class members aren't in alphabetic order. I don't think it's that valuable of a requirement, but at least be consistent!

firefox_puppeteer/ui/tabbar.py
```diff
@@ -3,126 +3,315 @@
...
         """
-        tabs = (self.marionette.find_element('id', 'tabbrowser-tabs')
-                               .find_elements('tag name', 'tab'))
-        return [self.TabElement(tab) for tab in tabs]
+        # Get handles from tab exceptions, and find those which can be closed
+        for tab in self.tabs:
+            if tab not in exceptions:
+                tab.close(force=True)
+
+    def close_tab(self, tab=None, *args, **kwargs):
```
chmanchester: You should know the arguments expected ahead of time here, I'd keep away from "*args, **kwargs" when you can.

firefox_puppeteer/ui/tabbar.py
```diff
@@ -3,126 +3,315 @@
...
+        elif trigger == 'button':
+            self.window.tabbar.newtab_button.click()
+        elif trigger == 'menu':
+            # TODO: Make use of menubar class once it supports ids
+            menu = self.window.marionette.find_element('id', 'menu_newNavigatorTab')
+            menu.click()
+        elif trigger == 'shortcut':
+            self.window.send_shortcut(self.window.get_localized_entity('tabCmd.commandkey'),
+                                      accel=True)
+        # elif - need to add other cases
```
chmanchester: If this still a todo? What are they?

firefox_puppeteer/ui/tabbar.py
```diff
@@ -3,126 +3,315 @@
...
+                tab.switch_to()
+                if target(tab):
                     return tab
 
-            raise NoSuchElementException('Tab with a label containing "{}"" not'
-                                         ' found'.format(target))
+            self.marionette.switch_to_window(start_handle)
+            raise errors.UnknownTabError("No tab found for '{}'".format(target))
+
+        raise ValueError("The 'target' parameter must either be an index or a callable")
```
chmanchester: I guess this should be InvalidValueError, like the others.

firefox_puppeteer/ui/tabbar.py
```diff
@@ -3,126 +3,315 @@
...
-        Wraps a tab element.
+        return self.tab.find_element('anon attribute', {'anonid': 'close-button'})
+
+    @property
+    def tab(self):
+        """The inner tab DOM element.
+
+        :returns: Tab DOM element
         """
+        return self._tab
```
chmanchester: tab.tab is a weird api. How about tab.tab_element?

firefox_puppeteer/ui/windows.py
```diff
@@ -424,14 +429,21 @@ def navbar(self):
         See the :class:`~ui.navbar.NavBar` reference.
         """
 
-    @use_class_as_property('ui.tabbar.Tabs')
+    @property
```
chmanchester: Why this change? This implementation is doing about the same thing as use_class_as_property.

firefox_ui_tests/functional/private_browsing/test_about_private_browsing.py
```diff
@@ -3,7 +3,7 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 from marionette import By
-from marionette.errors import NoSuchElementException
+# from marionette.errors import NoSuchElementException
```
chmanchester: Delete this.

firefox_ui_tests/functional/tabbed_browsing/test_new_tab.py
```diff
@@ -18,9 +18,7 @@ def tearDown(self):
         FirefoxTestCase.tearDown(self)
 
     def test_open_tab_by_newtab_button(self):
```
chmanchester: We can probably get rid of this file altogether.

firefox_puppeteer/base.py
whimboo: Indeed. Should have been `from . import errors`. Thanks.

firefox_puppeteer/base.py
whimboo: The line length is not a problem given that we allow 100 chars now. Also I updated it to keep it in sync with the `UIBaseLib` class, which I have added below. So it's not totally unrelated.

firefox_puppeteer/base.py
whimboo: Ok, I can move it to the top-level but I will have to import it with the full absolute path. This would solve at least your concern here. We can check that sometime later.

firefox_puppeteer/tests/manifest.ini
whimboo: This only works when I keep the above import for BaseLib local.

firefox_puppeteer/base.py
whimboo: Using `type()` here will not work given that this will not cover all the window sub-classes. I would leave it for now and file a follow-up bug to sort this out.

firefox_puppeteer/tests/test_tabs.py
whimboo: Any ui library based on UIBaseLib will need a reference to the window it is contained in, e.g. to retrieve dtds and others. Given that we have wrappers to bring in additional features for windows, there has to be a need to reference the appropriate handle. If you have a proposal please file a bug with some examples, so I could understand what you mean. Thanks.

firefox_puppeteer/tests/test_tabs.py
whimboo: I would add this additionally to the comparison with current_window_handle, just to avoid that we miss broken behavior with the tabs list.

firefox_puppeteer/tests/test_tabs.py
whimboo: Depending on settings in Firefox there is a different behavior in closing tabs. It could be right to left or the other way around. We could combine the proposal from above with this strategy then.

firefox_puppeteer/ui/tabbar.py
whimboo: Nope, it's just an indicator for code blocks. I will add a doc comment for the class.

firefox_puppeteer/ui/tabbar.py
whimboo: We have blocks here to make it easier to find them:

1. Properties for ui elements
2. Properties for helpers
3. Helper methods

I think that this is pretty helpful in case of large classes when you would have to scroll a lot up and down.

firefox_puppeteer/ui/tabbar.py
whimboo: I just wanted to avoid the redefinition of the default values. But lets change it. Not sure if there is a better way for it or not.

firefox_puppeteer/ui/tabbar.py
whimboo: Yes, therefore we need more elements to work on, which will be part of the follow-up bug. Those are not that often used.

firefox_puppeteer/ui/tabbar.py
whimboo: Oh, afaics this came in from you. :) But that opens the question why I have created InvalidValueError, while we have an existing Python error here. I will remove our class and make use of ValueError instead.

firefox_puppeteer/ui/tabbar.py
whimboo: I know and I wasn't that sure what to use here either. :) But sounds like a good compromise, which we should also do for the Window classes.

firefox_puppeteer/ui/windows.py
whimboo: Please watch exactly! We have to pass-in a separate argument to the constructor, which is not possible via a decorator.

firefox_ui_tests/functional/tabbed_browsing/test_new_tab.py
whimboo: Indeed. I will remove it.

whimboo:
I fixed the remaining nits and pushed as https://github.com/mozilla/firefox-ui-tests/commit/37736b3257e761e6c05fa3033edd3ee4670bbd14

PR #57, opened 2015-01-31
Bug 1127823 - Make use of unique parameter types in close_all() of Windows class

galgeek:
I am sympathetic both to Henrik's suggestion that close_all should expect a particular parameter type and to Chris’s distaste for APIs that convert a single element to a list, so I've implemented both here, in case that's what everybody wants after all.

### DIFF
--- a/firefox_puppeteer/tests/test_windows.py
+++ b/firefox_puppeteer/tests/test_windows.py
@@ -119,7 +119,7 @@ class TestWindows(FirefoxTestCase):
         # Check for an unexpected window class
         self.assertRaises(errors.UnexpectedWindowTypeError,
                           win1.open_window, expected_window_class=BaseWindow)
-        self.windows.close_all([BaseWindow(lambda: self.marionette, win1.handle)])
+        self.windows.close_all([win1])
 
     def test_base_window_switch_to_and_focus(self):
         # force BaseWindow instance
--- a/firefox_puppeteer/tests/test_windows.py
+++ b/firefox_puppeteer/tests/test_windows.py
@@ -51,7 +51,7 @@ class TestWindows(FirefoxTestCase):
         self.assertRaises(NoSuchWindowException,
                           self.windows.switch_to, lambda win: False)
 
-        self.windows.close_all(self.browser)
+        self.windows.close_all([self.browser])
         self.browser.switch_to()
 
         self.assertEqual(len(self.windows.all), 1)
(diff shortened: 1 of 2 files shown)


firefox_puppeteer/ui/windows.py
```diff
@@ -60,22 +60,17 @@ def close(self, handle):
         # TODO: Maybe needs to wait as handled via an observer
         return self.marionette.close_chrome_window()
 
-    def close_all(self, exceptions=None):
+    def close_all(self, exceptions=[]):
```
whimboo: Please never use other values than `None` as default for lists and other non-atomic types. This can result in unwanted behavior. So please leave the default value as it was before, and check for it inside the method first.

firefox_puppeteer/ui/windows.py
```diff
@@ -60,22 +60,17 @@ def close(self, handle):
...
         """Closes all open chrome windows.
 
         There is an optional `exceptions` list, which can be used to exclude
         specific chrome windows from being closed.
 
-        :param exceptions: Optional, list or a single entry of handles or
-         :class:`BaseWindow` instances not to close
+        :param exceptions: Optional, list of :class:`BaseWindow` instances
+                           not to close
```
whimboo: We do not want such an indentation. Please follow what the other code in this file is using.

firefox_puppeteer/tests/test_windows.py
```diff
@@ -119,7 +119,7 @@ def closer(window):
         # Check for an unexpected window class
         self.assertRaises(errors.UnexpectedWindowTypeError,
                           win1.open_window, expected_window_class=BaseWindow)
-        self.windows.close_all([win1.handle])
+        self.windows.close_all([BaseWindow(lambda: self.marionette, win1.handle)])
```
whimboo: You can simple pass in win1, which already is a BaseWindow.

firefox_puppeteer/ui/windows.py
```diff
@@ -66,16 +66,13 @@ def close_all(self, exceptions=None):
         There is an optional `exceptions` list, which can be used to exclude
         specific chrome windows from being closed.
 
-        :param exceptions: Optional, list or a single entry of handles or
-         :class:`BaseWindow` instances not to close
+        :param exceptions: Optional, list of :class:`BaseWindow` instances
+         not to close
         """
         handles_to_keep = exceptions or []
```
whimboo: Please rename the variable. What you will have here are `windows_to_keep`.

firefox_puppeteer/ui/windows.py
```diff
@@ -66,16 +66,13 @@ def close_all(self, exceptions=None):
...
-        if not isinstance(handles_to_keep, list):
-            handles_to_keep = [handles_to_keep]
 
-        # Ensure we only have handles and no BaseWindow entries
-        handles_to_keep = [entry.handle if isinstance(entry, BaseWindow) else entry
-                           for entry in handles_to_keep]
+        # Get handles of BaseWindow instances in exceptions
```
whimboo: nit: try to keep blank lines before a comment.

whimboo:
PR merged as https://github.com/mozilla/firefox-ui-tests/commit/b6c5183d1ceeaeacc315d9d56222559ed8504a3c

PR #64, opened 2015-02-04
Bug 1129429 - Convert Mozmill js test testSuggestBookmarks.js to a Firefox UI Test;r=ahal

bobsilverberg:


bobsilverberg:
Thanks @ahal. I added a commit that attempts to address the issues you raised. I did have a flake8 pre-commit hook in place, but it was ignoring `E501`. I fixed that too.

bobsilverberg:
@ahal I think the Travis failure is due to needing ` @skip_under_xvfb`. Do you agree?

bobsilverberg:
Travis is happy and this has @ahal's r+ so this is ready to merge.

firefox_ui_tests/functional/awesomebar/manifest.ini
```diff
@@ -0,0 +1 @@
+[test_star_in_autocomplete.py]
```
whimboo: We decided yesterday to have those tests located under `functional/locationbar`. Can you please do that in a follow-up?

firefox_ui_tests/functional/awesomebar/test_star_in_autocomplete.py
```diff
@@ -0,0 +1,76 @@
...
+
+        # Location bar suggests 'History and Bookmarks'
+        self.prefs.set_pref(self.PREF_LOCATION_BAR_SUGGEST, 0)
+
+    def tearDown(self):
+        FirefoxTestCase.tearDown(self)
+
+        self.prefs.reset_pref(self.PREF_LOCATION_BAR_SUGGEST)
+
+    def clear_history(self):
```
whimboo: It would be good to mark it with a TODO given that this has to be part of the places library. It will make it hard for us to update the tests later and to find all those entries.

firefox_ui_tests/functional/awesomebar/test_star_in_autocomplete.py
```diff
@@ -0,0 +1,76 @@
...
+
+        self.clear_history()
+
+        # Location bar suggests 'History and Bookmarks'
+        self.prefs.set_pref(self.PREF_LOCATION_BAR_SUGGEST, 0)
+
+    def tearDown(self):
+        FirefoxTestCase.tearDown(self)
+
+        self.prefs.reset_pref(self.PREF_LOCATION_BAR_SUGGEST)
```
whimboo: There is no need to call reset_pref() here. We have restore_prefs() in FirefoxTestCase.tearDown(). So please get this removed.

firefox_ui_tests/functional/awesomebar/test_star_in_autocomplete.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def clear_history(self):
+        with self.marionette.using_context('content'):
+            self.marionette.navigate('about:blank')
+
+        self.marionette.execute_script('''
+            let count = gBrowser.sessionHistory.count;
+            gBrowser.sessionHistory.PurgeHistory(count);
+        ''')
```
whimboo: This is not removing the bookmark you added via your test. It's only cleaning up the history.

firefox_ui_tests/functional/awesomebar/test_star_in_autocomplete.py
```diff
@@ -0,0 +1,76 @@
...
+
+    @skip_under_xvfb
+    def test_star_in_autocomplete(self):
+        search_string = 'grants'
+
+        with self.marionette.using_context('content'):
+            self.marionette.navigate(self.marionette.absolute_url('layout/mozilla_grants.html'))
+
+        self.browser.menubar.select('Bookmarks', 'Bookmark This Page')
+        done_button = self.marionette.find_element('id', 'editBookmarkPanelDoneButton')
```
whimboo: We do not wait for the panel to be open here. If we do not do this, it adds a race condition. Maybe add a sleep .5 and add a TODO item.

firefox_ui_tests/functional/awesomebar/test_star_in_autocomplete.py
```diff
@@ -0,0 +1,76 @@
...
+        # The data for autocomplete does not load immediately and we cannot wait for an observer
+        # until the bug 1121731 is complete
+        time.sleep(4)
+        locationbar.clear()
+        locationbar.urlbar.send_keys(search_string)
+        autocomplete_results = locationbar.autocomplete_results
+
+        self.wait_for_condition(lambda mn: locationbar.value == search_string)
+        self.wait_for_condition(lambda mn: autocomplete_results.is_open)
+        self.wait_for_condition(lambda mn: len(autocomplete_results.visible_results) == 1)
```
whimboo: In the Mozmill test we have two entries in the autocomplete. I wonder why we have a single one now.

firefox_ui_tests/functional/awesomebar/test_star_in_autocomplete.py
bobsilverberg: Ah, I see that the current test has a call to `places.restoreDefaultBookmarks()` in teardownModule. Should I just add a TODO to add that when the places module becomes available, or should I use the UI to remove the bookmark at the end of the test?

firefox_ui_tests/functional/awesomebar/test_star_in_autocomplete.py
bobsilverberg: Can we just wait for the element to be visible? That is what we generally do for WebDriver and FxOS Marionette tests.

bobsilverberg: For example:

```
        done_button = self.marionette.find_element('id', 'editBookmarkPanelDoneButton')
        self.wait_for_condition(lambda mn: done_button.is_displayed)
        done_button.click()
```

firefox_ui_tests/functional/awesomebar/test_star_in_autocomplete.py
bobsilverberg: That's not what I'm seeing at http://hg.mozilla.org/qa/mozmill-tests/file/default/firefox/tests/functional/testAwesomeBar/testSuggestBookmarks.js#l80

firefox_ui_tests/functional/awesomebar/test_star_in_autocomplete.py
```diff
@@ -0,0 +1,76 @@
...
+
+        first_result = autocomplete_results.visible_results[0]
+        matching_title = autocomplete_results.get_matching_text(first_result, 'title')[0]
```
whimboo: I think it should be fine to use the first entry instead of the second one, but I would propose that we check all the underlined text for this autocomplete result, and not only the first.

firefox_ui_tests/functional/awesomebar/test_star_in_autocomplete.py
```diff
@@ -0,0 +1,76 @@
...
+        self.assertEqual(
+            search_string.lower(),
+            matching_title.lower(),
+            'The page title matches the highlighted text')
```
whimboo: We had issues in the past on slower machines. So I would love to see a wait_for_condition() call here.

firefox_ui_tests/functional/awesomebar/test_star_in_autocomplete.py
```diff
@@ -0,0 +1,76 @@
...
+        self.assertEqual('bookmark',
+                         first_result.get_attribute('type'),
+                         'The auto-complete result is a bookmark')
```
whimboo: We shouldn't use assertEqual() here but a method which checks that 'bookmark' is contained. This attribute is like the CSS class attribute and can contain other entries too.

firefox_ui_tests/functional/awesomebar/test_star_in_autocomplete.py
whimboo: Also make sure that you close the autocomplete popup here. If it lingers around can break following tests.

firefox_ui_tests/functional/awesomebar/manifest.ini
bobsilverberg: Done

firefox_ui_tests/functional/awesomebar/test_star_in_autocomplete.py
bobsilverberg: Done

firefox_ui_tests/functional/awesomebar/test_star_in_autocomplete.py
bobsilverberg: Done

firefox_ui_tests/functional/awesomebar/test_star_in_autocomplete.py
bobsilverberg: I added this, but maybe it's not ok? See PR #65

firefox_ui_tests/functional/awesomebar/test_star_in_autocomplete.py
bobsilverberg: Done

firefox_ui_tests/functional/awesomebar/test_star_in_autocomplete.py
bobsilverberg: Done

firefox_ui_tests/functional/awesomebar/test_star_in_autocomplete.py
bobsilverberg: Done

firefox_ui_tests/functional/awesomebar/test_star_in_autocomplete.py
whimboo: I think that I can quickly get the places library implemented. All looks pretty straight forward. It will help you all to proceed without the need of workarounds.

firefox_ui_tests/functional/awesomebar/test_star_in_autocomplete.py
whimboo: Sorry, looks like that was a recent change and I haven't updated my local repo for a couple of days. It's fine then.

PR #65, opened 2015-02-04
Bug 1129429 - Convert Mozmill js test testSuggestBookmarks.js to a Firefox UI Test;r=whimboo

bobsilverberg:
Update to address Henrik's comments

Move location to /functional/locationbar
Remove reset_pref from tearDown
Close autocomplete results in tearDown
Add comment to replace clear_history()
Add a wait for editBookmarkPanelDoneButton to be displayed
Check all matches for the title in the autocomplete results
Wait for the title match instead of using an assert
Assert that 'bookmark' is in the 'type' attribute, instead of equal


bobsilverberg:
@whimboo This is a follow up to PR #64

firefox_ui_tests/functional/locationbar/test_star_in_autocomplete.py
```diff
@@ -25,10 +25,13 @@ def setUp(self):
         self.prefs.set_pref(self.PREF_LOCATION_BAR_SUGGEST, 0)
 
     def tearDown(self):
-        FirefoxTestCase.tearDown(self)
+        # Close the autocomplete results
+        self.browser.navbar.locationbar.autocomplete_results.close()
 
-        self.prefs.reset_pref(self.PREF_LOCATION_BAR_SUGGEST)
+        FirefoxTestCase.tearDown(self)
```
whimboo: Better do a try/finally here, so in case of something fails before we call the tearDown of FirefoxTestCase.

firefox_ui_tests/functional/locationbar/test_star_in_autocomplete.py
```diff
@@ -47,6 +50,7 @@ def test_star_in_autocomplete(self):
 
         self.browser.menubar.select('Bookmarks', 'Bookmark This Page')
         done_button = self.marionette.find_element('id', 'editBookmarkPanelDoneButton')
```
whimboo: Please also add a TODO comment here that this need a new UI class for the editBookmarkPanel.

bobsilverberg:
Thanks @whimboo. This has been updated.

firefox_ui_tests/functional/locationbar/test_star_in_autocomplete.py
```diff
@@ -25,10 +25,13 @@ def setUp(self):
...
     def tearDown(self):
-        FirefoxTestCase.tearDown(self)
+        # Close the autocomplete results
+        self.browser.navbar.locationbar.autocomplete_results.close()
 
-        self.prefs.reset_pref(self.PREF_LOCATION_BAR_SUGGEST)
+        FirefoxTestCase.tearDown(self)
 
+    # TODO: Replace with places module method when available
+    # https://bugzilla.mozilla.org/show_bug.cgi?id=1121731
```
whimboo: The module will land soon, so you can change all those lines and make use of the appropriate methods. I will add the dependency to the bug.

firefox_ui_tests/functional/locationbar/test_star_in_autocomplete.py
```diff
@@ -19,24 +21,26 @@ class TestStarInAutocomplete(FirefoxTestCase):
...
-        FirefoxTestCase.tearDown(self)
-
-        self.prefs.reset_pref(self.PREF_LOCATION_BAR_SUGGEST)
-
-    def clear_history(self):
+        # Close the autocomplete results
+        try:
+            self.browser.navbar.locationbar.autocomplete_results.close()
+            self.places.restore_default_bookmarks()
+        except NoSuchElementException:
```
whimboo: When does this exception happening to you? You should really not see it.

firefox_ui_tests/functional/locationbar/test_star_in_autocomplete.py
```diff
@@ -19,24 +21,26 @@ class TestStarInAutocomplete(FirefoxTestCase):
...
+
+    def clean_up_state(self):
         with self.marionette.using_context('content'):
             self.marionette.navigate('about:blank')
 
-        self.marionette.execute_script('''
-            let count = gBrowser.sessionHistory.count;
-            gBrowser.sessionHistory.PurgeHistory(count);
-        ''')
+        self.places.remove_all_history()
```
whimboo: You can get rid of this extra method, and put those lines into setUp() directly.

firefox_ui_tests/functional/locationbar/test_star_in_autocomplete.py
```diff
@@ -45,15 +49,18 @@ def test_star_in_autocomplete(self):
         with self.marionette.using_context('content'):
             self.marionette.navigate(self.marionette.absolute_url('layout/mozilla_grants.html'))
```
whimboo: I would propose you add a `self.test_urls` to setUp() which prepares all that including the call to absolute_url(). See Matt's PR here: https://github.com/mozilla/firefox-ui-tests/pull/70/files#diff-aa5ddb89f02abb509ef9bdd35d53a9a8R25

firefox_ui_tests/functional/locationbar/test_star_in_autocomplete.py
```diff
@@ -45,15 +49,18 @@ def test_star_in_autocomplete(self):
...
 
+        # TODO: Convert to l10n friendly accessor when menu library is available
         self.browser.menubar.select('Bookmarks', 'Bookmark This Page')
+        # TODO: Replace hard-coded selector with library method when one is available
         done_button = self.marionette.find_element('id', 'editBookmarkPanelDoneButton')
```
whimboo: Please use `By.ID` as method. It's available via marionette.By.

firefox_ui_tests/functional/locationbar/test_star_in_autocomplete.py
```diff
@@ -45,15 +49,18 @@ def test_star_in_autocomplete(self):
...
+        self.clean_up_state()
 
         locationbar = self.browser.navbar.locationbar
-        # TODO: Replace with visited observer when places module is available
-        # https://bugzilla.mozilla.org/show_bug.cgi?id=1121731
+        # TODO: Replace with places.wait_for_visited when observers have been implemented
+        # https://bugzilla.mozilla.org/show_bug.cgi?id=1121691
         # The data for autocomplete does not load immediately and we cannot wait for an observer
         # until the bug 1121731 is complete
         time.sleep(4)
```
whimboo: `self.places.wait_for_visited` exists now. So please replace this code with this call. No need for those comments anymore.

whimboo: Oh, you want to make sure this is called when you navigate to the test page. So use marionette.navigate() inside the callback.

whimboo: Make sure that you also open about:blank after bookmarking the page.

firefox_ui_tests/functional/locationbar/test_star_in_autocomplete.py
```diff
@@ -66,11 +73,9 @@ def test_star_in_autocomplete(self):
...
-        self.assertEqual(
-            search_string.lower(),
-            matching_title.lower(),
-            'The page title matches the highlighted text')
-        self.assertEqual('bookmark',
-                         first_result.get_attribute('type'),
-                         'The auto-complete result is a bookmark')
+        matching_titles = autocomplete_results.get_matching_text(first_result, 'title')
+        for title in matching_titles:
+            self.wait_for_condition(lambda mn: search_string.lower() == title.lower())
```
whimboo: Please put the expected value as second inside a comparison, that makes it easier to see what we actually want to test. Also search_string is already lower-cased by definition.

firefox_ui_tests/functional/locationbar/test_star_in_autocomplete.py
bobsilverberg: It wasn't happening. I was just guessing at an exception. I will just do a `try`/`finally` without an `except`, although that seems a bit dangerous to me. What would you put in the `except`?

firefox_ui_tests/functional/locationbar/test_star_in_autocomplete.py
bobsilverberg: Done

firefox_ui_tests/functional/locationbar/test_star_in_autocomplete.py
whimboo: Leave out the except clause. So if something fails it will be printed out, and we know something failed and needs a fix.

firefox_ui_tests/functional/locationbar/test_star_in_autocomplete.py
bobsilverberg: Ok, I've done this. I wasn't sure whether I should loop through the list of urls visiting them, or just use `self.test_urls[0]` as we know there is only one. It feels better to loop through the list even though it is unnecessary, as that will continue to work if for some reason we add more urls to the test later. You'll see that change in the next commit, coming soon.

firefox_ui_tests/functional/locationbar/test_star_in_autocomplete.py
bobsilverberg: Done.

firefox_ui_tests/functional/locationbar/test_star_in_autocomplete.py
bobsilverberg: Done.

firefox_ui_tests/functional/locationbar/test_star_in_autocomplete.py
bobsilverberg: Done

firefox_ui_tests/functional/locationbar/test_star_in_autocomplete.py
```diff
@@ -19,44 +19,48 @@ class TestStarInAutocomplete(FirefoxTestCase):
...
 
     @skip_under_xvfb
     def test_star_in_autocomplete(self):
         search_string = 'grants'
 
+        def visit_urls():
+            for url in self.test_urls:
+                self.marionette.navigate(url)
+
         with self.marionette.using_context('content'):
```
whimboo: The with statement you better have to put into the callback. The helper `wait_for_visited` is working in chrome scope.

firefox_ui_tests/functional/locationbar/test_star_in_autocomplete.py
```diff
@@ -2,7 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
```
whimboo: Can we keep the former name of the test please? `test_suggest_bookmarks.py` is more descriptive here and does not tell how Firefox is showing bookmarks. It might change in the future.

firefox_ui_tests/functional/locationbar/test_star_in_autocomplete.py
```diff
@@ -19,44 +19,48 @@ class TestStarInAutocomplete(FirefoxTestCase):
...
+
+        self.places.remove_all_history()
 
         locationbar = self.browser.navbar.locationbar
-        # TODO: Replace with visited observer when places module is available
-        # https://bugzilla.mozilla.org/show_bug.cgi?id=1121731
-        # The data for autocomplete does not load immediately and we cannot wait for an observer
-        # until the bug 1121731 is complete
-        time.sleep(4)
         locationbar.clear()
```
whimboo: Maybe you can add some comments to the blocks of code which describe briefly what's done?

whimboo:
Bob, there is also a merge conflict, which you will have to solve.

firefox_ui_tests/functional/locationbar/test_star_in_autocomplete.py
bobsilverberg: Done

firefox_ui_tests/functional/locationbar/test_star_in_autocomplete.py
bobsilverberg: Done

firefox_ui_tests/functional/locationbar/test_star_in_autocomplete.py
bobsilverberg: Tbh, I feel the code is pretty self-documenting. It is pretty clear what is going on just by reading the method names. Having said that, I did add a few comments, particularly ones that were in the original test. Let me know if you feel there are still too few, or now too many.

bobsilverberg:
I noticed that Travis failed, and I think it is because only the test was marked with `@skip_under_xvfb` so it was likely attempting the `tearDown`. I moved the marker to the class. We'll see if that makes Travis happy.

whimboo:
Looks great! Thanks for all the updates @bobsilverberg. I will merge it now.

whimboo:
Merged the PR as https://github.com/mozilla/firefox-ui-tests/commit/510a6313f8608e91cb95ad1f501a894227241430.

PR #66, opened 2015-02-04
Bug 1129159 - Convert Mozmill testEscapeAutocomplete

galgeek:
learn more here: https://bugzilla.mozilla.org/show_bug.cgi?id=1129159

### DIFF
--- a/firefox_ui_tests/functional/locationbar/test_escape_autocomplete.py
+++ b/firefox_ui_tests/functional/locationbar/test_escape_autocomplete.py
@@ -24,10 +24,10 @@ class TestEscapeAutocomplete(FirefoxTestCase):
         self.test_string = 'mozilla'
 
         self.locationbar = self.browser.navbar.locationbar
-        self.autocompleteresults = self.locationbar.autocomplete_results
+        self.autocomplete_results = self.locationbar.autocomplete_results
 
     def tearDown(self):
-        self.autocompleteresults.close(force=True)
+        self.autocomplete_results.close(force=True)
         FirefoxTestCase.tearDown(self)
 
     @skip_under_xvfb
@@ -44,12 +44,12 @@ class TestEscapeAutocomplete(FirefoxTestCase):
         self.locationbar.clear()
         self.locationbar.urlbar.send_keys(self.test_string)
         self.assertEqual(self.locationbar.value, self.test_string)
-        self.wait_for_condition(lambda _: self.autocompleteresults.is_open)
+        self.wait_for_condition(lambda _: self.autocomplete_results.is_open)
 
         # Press escape, check location bar value, check autocomplete list closed
         self.locationbar.urlbar.send_keys(self.keys.ESCAPE)
         self.assertEqual(self.locationbar.value, self.test_string)
-        self.wait_for_condition(lambda _: not self.autocompleteresults.is_open)
+        self.wait_for_condition(lambda _: not self.autocomplete_results.is_open)
 
         # Press escape again and check that locationbar returns to the page url
         self.locationbar.urlbar.send_keys(self.keys.ESCAPE)
--- a/firefox_ui_tests/functional/locationbar/test_escape_autocomplete.py
+++ b/firefox_ui_tests/functional/locationbar/test_escape_autocomplete.py
@@ -8,21 +8,12 @@ from firefox_ui_harness.testcase import FirefoxTestCase
 
 class TestEscapeAutocomplete(FirefoxTestCase):
 
-    def tearDown(self):
-        autocompleteresults = self.browser.navbar.locationbar.autocomplete_results
+    def setUp(self):
+        FirefoxTestCase.setUp(self)
 
-        if autocompleteresults.is_open:
-            autocompleteresults.close(force=True)
-            self.wait_for_condition(not autocompleteresults.is_open)
-
-        FirefoxTestCase.tearDown(self)
-
-    @skip_under_xvfb
-    def test_escape_autocomplete(self):
         # Clear complete history so there's no interference from previous entries.
         self.places.remove_all_history()
 
-        # Open some local pages to set up the test environment.
         self.test_urls = [
             'layout/mozilla.html',
             'layout/mozilla_community.html',
@@ -30,29 +21,36 @@ class TestEscapeAutocomplete(FirefoxTestCase):
         self.test_urls = [self.marionette.absolute_url(t)
                           for t in self.test_urls]
 
+        self.test_string = 'mozilla'
+
+        self.locationbar = self.browser.navbar.locationbar
+        self.autocompleteresults = self.locationbar.autocomplete_results
+
+    def tearDown(self):
+        self.autocompleteresults.close(force=True)
+        FirefoxTestCase.tearDown(self)
+
+    @skip_under_xvfb
+    def test_escape_autocomplete(self):
+
+        # Open some local pages
         def load_urls():
             with self.marionette.using_context('content'):
                 for url in self.test_urls:
                     self.marionette.navigate(url)
         self.places.wait_for_visited(self.test_urls, load_urls)
 
-        # Set variables
-        TEST_STRING = 'mozilla'
-
-        locationbar = self.browser.navbar.locationbar
-        autocompleteresults = locationbar.autocomplete_results
-
         # Clear the location bar, type the test string, check that autocomplete list opens
-        locationbar.clear()
-        locationbar.urlbar.send_keys(TEST_STRING)
-        self.assertEqual(locationbar.value, TEST_STRING)
-        self.wait_for_condition(lambda _: autocompleteresults.is_open)
+        self.locationbar.clear()
+        self.locationbar.urlbar.send_keys(self.test_string)
+        self.assertEqual(self.locationbar.value, self.test_string)
+        self.wait_for_condition(lambda _: self.autocompleteresults.is_open)
 
         # Press escape, check location bar value, check autocomplete list closed
-        locationbar.urlbar.send_keys(self.keys.ESCAPE)
-        self.assertEqual(locationbar.value, TEST_STRING)
-        self.assertFalse(autocompleteresults.is_open)
+        self.locationbar.urlbar.send_keys(self.keys.ESCAPE)
+        self.assertEqual(self.locationbar.value, self.test_string)
+        self.wait_for_condition(lambda _: not self.autocompleteresults.is_open)
 
         # Press escape again and check that locationbar returns to the page url
-        locationbar.urlbar.send_keys(self.keys.ESCAPE)
-        self.assertEqual(locationbar.value, self.test_urls[-1])
+        self.locationbar.urlbar.send_keys(self.keys.ESCAPE)
+        self.assertEqual(self.locationbar.value, self.test_urls[-1])
--- a/firefox_ui_tests/functional/locationbar/test_escape_autocomplete.py
+++ b/firefox_ui_tests/functional/locationbar/test_escape_autocomplete.py
@@ -8,17 +8,24 @@ from firefox_ui_harness.testcase import FirefoxTestCase
 
 class TestEscapeAutocomplete(FirefoxTestCase):
 
-    def setUp(self):
-        FirefoxTestCase.setUp(self)
+    def tearDown(self):
+        autocompleteresults = self.browser.navbar.locationbar.autocomplete_results
+        
+        if autocompleteresults.is_open:
+            autocompleteresults.close(force=True)
+            self.wait_for_condition(lambda _: autocompleteresults.is_open == False)
 
+        FirefoxTestCase.tearDown(self)
+
+    @skip_under_xvfb
+    def test_escape_autocomplete(self):
         # Clear complete history so there's no interference from previous entries.
         self.places.remove_all_history()
 
         # Open some local pages to set up the test environment.
         self.test_urls = [
             'layout/mozilla.html',
-            'layout/mozilla_mission.html',
-            'layout/mozilla_grants.html',
+            'layout/mozilla_community.html',
         ]
         self.test_urls = [self.marionette.absolute_url(t)
                           for t in self.test_urls]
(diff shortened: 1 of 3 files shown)


firefox_ui_tests/functional/locationbar/test_escape_autocomplete.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+class TestEscapeAutocomplete(FirefoxTestCase):
+
+    # setUp copied from test_back_forward
+    def setUp(self):
+        FirefoxTestCase.setUp(self)
+
+        with self.marionette.using_context("content"):
+            self.marionette.navigate("about:blank")
```
whimboo: I don't think we need that. We navigate to other urls right away in line 34.

firefox_ui_tests/functional/locationbar/test_escape_autocomplete.py
```diff
@@ -0,0 +1,72 @@
...
+            'layout/mozilla_mission.html',
+            'layout/mozilla_grants.html',
+        ]
+        self.test_urls = [self.marionette.absolute_url(t)
+                          for t in self.test_urls]
+
+        with self.marionette.using_context('content'):
+            for url in self.test_urls:
+                self.marionette.navigate(url)
+            self.assertEquals(self.marionette.get_url(), self.test_urls[-1])
```
whimboo: I wonder if Marionette has an internal check that the requested page has been loaded. Can you please check that?

firefox_ui_tests/functional/locationbar/test_escape_autocomplete.py
```diff
@@ -0,0 +1,72 @@
...
+
+    @skip_under_xvfb
+    def test_escape_autocomplete(self):
+        TEST_STRING = 'mozilla'
+
+        locationbar = self.browser.navbar.locationbar
+        current_url = locationbar.value
+        autocompleteresults = self.browser.navbar.locationbar.autocomplete_results
+
+        # 3. (Focus the locationbar and) delete any locationbar content.
```
whimboo: For the final patch please get rid of those comments.

firefox_ui_tests/functional/locationbar/test_escape_autocomplete.py
```diff
@@ -0,0 +1,72 @@
...
+        locationbar.clear()
+
+        # 4. Type a test string in the locationbar.
+        locationbar.urlbar.send_keys(TEST_STRING)
+
+        # 5. Check that the test string matches locationbar value.
+        self.assertEqual(locationbar.value, TEST_STRING)
```
whimboo: An assert is not necessary. We have unit tests which ensure that send_keys is working.

firefox_ui_tests/functional/locationbar/test_escape_autocomplete.py
```diff
@@ -0,0 +1,72 @@
...
+
+        # 6. Check that the autocomplete list has opened.
+        self.wait_for_condition(lambda _: autocompleteresults.is_open)
+        self.assertTrue(autocompleteresults.is_open)
```
whimboo: You do the assertion already with wait_for_condition. So no extra assert is necessary here.

firefox_ui_tests/functional/locationbar/test_escape_autocomplete.py
whimboo: What can do is to group actions and asserts, and add a comment above those blocks which briefly describes what is done.

galgeek:
Hi @whimboo!

I've updated per your comments, replacing code in setUp with what I think the places library offers. Thank you!

Regarding whether Marionette has an internal check that the requested page has been loaded, "navigate" returns when DomContentLoaded has fired—it's possible for AJAX still to load more (or more strictly speaking, when readyState == "complete" (or readyState == "interactive" for about: urls)). Compare #automation IRC logs just before midnight GMT, 5 February.

whimboo:
> Regarding whether Marionette has an internal check that the requested page has been loaded, "navigate" returns when DomContentLoaded has fired—it's possible for AJAX still to load more 

We do not worry about anything after DOMContentLoaded in this test, so I don't think this is important.

firefox_ui_tests/functional/locationbar/test_escape_autocomplete.py
```diff
@@ -0,0 +1,52 @@
...
+        ]
+        self.test_urls = [self.marionette.absolute_url(t)
+                          for t in self.test_urls]
+
+        def load_urls():
+            with self.marionette.using_context('content'):
+                for url in self.test_urls:
+                    self.marionette.navigate(url)
+        self.places.wait_for_visited(self.test_urls, load_urls)
+
```
whimboo: Please make sure to add a tearDown which ensures that the popup gets closed if not done so yet.

firefox_ui_tests/functional/locationbar/test_escape_autocomplete.py
```diff
@@ -0,0 +1,52 @@
...
+        FirefoxTestCase.setUp(self)
+
+        # Clear complete history so there's no interference from previous entries.
+        self.places.remove_all_history()
+
+        # Open some local pages to set up the test environment.
+        self.test_urls = [
+            'layout/mozilla.html',
+            'layout/mozilla_mission.html',
+            'layout/mozilla_grants.html',
```
whimboo: Please use the original URLs from the mozmill test. We are trying to get some variation in by not only using the same pages across all location bar tests.

firefox_ui_tests/functional/locationbar/test_escape_autocomplete.py
```diff
@@ -0,0 +1,52 @@
...
+        ]
+        self.test_urls = [self.marionette.absolute_url(t)
+                          for t in self.test_urls]
+
+        def load_urls():
+            with self.marionette.using_context('content'):
+                for url in self.test_urls:
+                    self.marionette.navigate(url)
+        self.places.wait_for_visited(self.test_urls, load_urls)
```
whimboo: I would actually put this into the test and not in setUp().

firefox_ui_tests/functional/locationbar/test_escape_autocomplete.py
```diff
@@ -0,0 +1,52 @@
...
+
+    @skip_under_xvfb
+    def test_escape_autocomplete(self):
+        TEST_STRING = 'mozilla'
+
+        locationbar = self.browser.navbar.locationbar
+        current_url = locationbar.value
```
whimboo: nit: I do not see a benefit from having this variable. Please remove it.

firefox_ui_tests/functional/locationbar/test_escape_autocomplete.py
```diff
@@ -0,0 +1,52 @@
...
+        autocompleteresults = self.browser.navbar.locationbar.autocomplete_results
+
+        # Clear the location bar, type the test string, check that autocomplete list opens
+        locationbar.clear()
+        locationbar.urlbar.send_keys(TEST_STRING)
```
whimboo: Constants in Python are lowercase. Please check your PEP8 output.

firefox_ui_tests/functional/locationbar/test_escape_autocomplete.py
```diff
@@ -0,0 +1,52 @@
...
+        self.wait_for_condition(lambda _: autocompleteresults.is_open)
```
whimboo: Please add the check for the location bar value first.

firefox_ui_tests/functional/locationbar/test_escape_autocomplete.py
```diff
@@ -0,0 +1,52 @@
...
+
+        # Press escape, check location bar value, check autocomplete list closed
+        locationbar.urlbar.send_keys(self.keys.ESCAPE)
+        self.assertEqual(locationbar.value, TEST_STRING)
+        self.assertFalse(autocompleteresults.is_open)
+
+        # Press escape again and check that locationbar returns to the page url
+        locationbar.urlbar.send_keys(self.keys.ESCAPE)
+        self.assertEqual(locationbar.value, current_url)
```
whimboo: Better comparing against `self.test_urls[-1]`?

firefox_ui_tests/functional/locationbar/test_escape_autocomplete.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+class TestEscapeAutocomplete(FirefoxTestCase):
+
+    def tearDown(self):
+        autocompleteresults = self.browser.navbar.locationbar.autocomplete_results
+
+        if autocompleteresults.is_open:
+            autocompleteresults.close(force=True)
+            self.wait_for_condition(not autocompleteresults.is_open)
```
whimboo: There is no need to check if the auto-complete is open. Just call `close()` with force set to True. Also this method itself checks if it is closed. So all those 5 lines above can become a single one.

firefox_ui_tests/functional/locationbar/test_escape_autocomplete.py
```diff
@@ -0,0 +1,58 @@
...
+
+        FirefoxTestCase.tearDown(self)
+
+    @skip_under_xvfb
+    def test_escape_autocomplete(self):
+        # Clear complete history so there's no interference from previous entries.
+        self.places.remove_all_history()
```
whimboo: This is setup logic. So please add it to the ´setUp()´ method.

firefox_ui_tests/functional/locationbar/test_escape_autocomplete.py
```diff
@@ -0,0 +1,58 @@
...
+
+        # Open some local pages to set up the test environment.
+        self.test_urls = [
+            'layout/mozilla.html',
+            'layout/mozilla_community.html',
+        ]
+        self.test_urls = [self.marionette.absolute_url(t)
+                          for t in self.test_urls]
```
whimboo: That one too. Looks like I missed to mention that in my last review, but this is setup logic.

firefox_ui_tests/functional/locationbar/test_escape_autocomplete.py
```diff
@@ -0,0 +1,58 @@
...
+
+        def load_urls():
+            with self.marionette.using_context('content'):
+                for url in self.test_urls:
+                    self.marionette.navigate(url)
+        self.places.wait_for_visited(self.test_urls, load_urls)
+
+        # Set variables
```
whimboo: Lets get this comment removed. I think its clear what you do here. Comments are usually helpful to describe a block of actions or something which is not immediately understandable.

firefox_ui_tests/functional/locationbar/test_escape_autocomplete.py
```diff
@@ -0,0 +1,58 @@
...
+        # Clear the location bar, type the test string, check that autocomplete list opens
+        locationbar.clear()
+        locationbar.urlbar.send_keys(TEST_STRING)
+        self.assertEqual(locationbar.value, TEST_STRING)
+        self.wait_for_condition(lambda _: autocompleteresults.is_open)
+
+        # Press escape, check location bar value, check autocomplete list closed
+        locationbar.urlbar.send_keys(self.keys.ESCAPE)
+        self.assertEqual(locationbar.value, TEST_STRING)
+        self.assertFalse(autocompleteresults.is_open)
```
whimboo: We should call `wait_for_condition` here similar to the Mozmill test. There can be delays in closing the popup.

firefox_ui_tests/functional/locationbar/test_escape_autocomplete.py
```diff
@@ -0,0 +1,56 @@
...
+            'layout/mozilla.html',
+            'layout/mozilla_community.html',
+        ]
+        self.test_urls = [self.marionette.absolute_url(t)
+                          for t in self.test_urls]
+
+        self.test_string = 'mozilla'
+
+        self.locationbar = self.browser.navbar.locationbar
+        self.autocompleteresults = self.locationbar.autocomplete_results
```
whimboo: Please keep the underscore in the variable name.

galgeek:
Hi, @whimboo!

Thanks for your feedback!

I've updated the PR.

firefox_ui_tests/functional/locationbar/test_escape_autocomplete.py
```diff
@@ -0,0 +1,56 @@
...
+        self.locationbar = self.browser.navbar.locationbar
+        self.autocomplete_results = self.locationbar.autocomplete_results
+
+    def tearDown(self):
+        self.autocomplete_results.close(force=True)
+        FirefoxTestCase.tearDown(self)
+
+    @skip_under_xvfb
+    def test_escape_autocomplete(self):
+
```
whimboo: nit: There is no need for this empty line.

whimboo:
As what I have seen right now, this PR is missing the manifest entry. So the test would never be executed.

galgeek:
Thanks for your feedback! I've updated the PR.

This test passes locally for me. It looks like it failed on Travis in the teardown, in the autocomplete_results.close, when this test should in fact be skipped.

firefox_ui_tests/functional/locationbar/test_escape_autocomplete.py
```diff
@@ -0,0 +1,57 @@
...
+                          for t in self.test_urls]
+
+        self.test_string = 'mozilla'
+
+        self.locationbar = self.browser.navbar.locationbar
+        self.autocomplete_results = self.locationbar.autocomplete_results
+
+    def tearDown(self):
+        # workaround for marionette bug leaving us in 'content' context here
+        self.marionette.set_context('chrome')
```
whimboo: What is putting is in content scope here if every test in that module is skipped? We are in chrome scope by default, so I don't understand where that should come from.

whimboo: Please get this reverted and put the skip decorator to the whole class. That works perfectly as we have seen on the other PR from Bob.

whimboo:
This PR needs a rebase because it can no longer be merged.

firefox_ui_tests/functional/locationbar/test_escape_autocomplete.py
galgeek: Chris suggested this workaround, noting there's a Marionette bug that leaves us in content context in tearDown.

Since the skip decorator works well applied to the class, that's the better solution here.

whimboo:
The PR has been merged as https://github.com/mozilla/firefox-ui-tests/commit/77e6205e2f8ee0c42ae241b16a141561c099389c

PR #67, opened 2015-02-04
Bug 1131181 - Add focusmanager.testmode: true to default prefs for firefox-ui-tests

galgeek:
Hi @whimboo!

Per today's discussion of focusmanager.testmode...

Not sure whether I ought to file a bug in bugzilla.

### DIFF
--- a/firefox_ui_harness/default_prefs.py
+++ b/firefox_ui_harness/default_prefs.py
@@ -36,7 +36,7 @@ default_prefs = {
     'extensions.showMismatchUI': False,
     'extensions.update.enabled': False,
     'extensions.update.notifyUser': False,
-    'focusmanager.testmode': False,
+    'focusmanager.testmode': True,
     'geo.provider.testing': True,
     'javascript.options.showInConsole': True,
     'security.notification_enable_delay': 0,
--- a/firefox_puppeteer/ui/windows.py
+++ b/firefox_puppeteer/ui/windows.py
@@ -44,7 +44,7 @@ class Windows(BaseLib):
         with self.marionette.using_context('chrome'):
             return self.marionette.execute_script("""
               Cu.import("resource://gre/modules/Services.jsm");
-              var win = Services.wm.getMostRecentWindow("");
+              var win = Services.focus.activeWindow;
               return win.QueryInterface(Ci.nsIInterfaceRequestor)
                         .getInterface(Ci.nsIDOMWindowUtils)
                         .outerWindowID.toString();


galgeek:
Hello, again, @whimboo, and @chmanchester!

I was surprised to see my one line PR adding focusmanager.testmode: True failing Travis.  Tests fail on OS X 10.10.2 similarly.


It looks like Marionette client may not add this pref for the desktop versions of Firefox.

Chris suggested on IRC: "I'd ask whimboo|afk about how the implementation of focused_chrome_window_handle (based on WindowMediator.getMostRecentWindow("")) might interact weirdly with testmode”

I thought I’d check for bugs against getMostRecentWindow and found (maybe you’re already familiar with this?)…

Bug 618470 - nsIWindowMediator::getMostRecentWindow doesn't match what is thought to be focused
https://bugzilla.mozilla.org/show_bug.cgi?id=618470

Bug 528706 - nsIWindowMediator::getMostRecentWindow should not return closed windows
https://bugzilla.mozilla.org/show_bug.cgi?id=528706 

and 

Bug 1029496 - Test failure '[JavaScript Error: "mozmill.wm.getMostRecentWindow(...).document.documentElement is null"' in testPrivateDownloadPanel.js
https://bugzilla.mozilla.org/show_bug.cgi?id=1029496

whimboo:
Can we please get a bug filed as usual for this task? Thanks.

Regarding those failures it would be important to add some print statements which show you which window is currently focused, and if that is different to what we get from currently_focused_window. If that is the case we may have to change the implementation.

firefox_ui_harness/default_prefs.py
```diff
@@ -36,6 +36,7 @@
     'extensions.showMismatchUI': False,
     'extensions.update.enabled': False,
     'extensions.update.notifyUser': False,
+    'focusmanager.testmode': True,
```
whimboo: Please check which other prefs we might miss compared to mozmill, which should be set by Marionette. We might also want to add them here for safety.

Regarding the above preference I would ask David if we even not have to enable it for Firefox Desktop by default.

galgeek: With the addition of focusmanager.testmode: True, we should have recorded in default_prefs everything that we talked about needing and wanting to include from mozmill.

The bug you filed, with links to mozmill's preferences settings, is here: 
https://bugzilla.mozilla.org/show_bug.cgi?id=1123683

David wasn't sure when I asked why focusmanager.testmode wasn't included with the Marionette client. I've filed a bug and will submit a patch for Marionette. David said he's "interested to see how the webapi tests cope with that pref."

galgeek:
Hi @whimboo and @chmanchester!

Here's a debug version of the PR, with some print statements in test_windows and the windows library to show better how the error happens in test_windows, plus the tearDown fix to avoid follow on failures or errors.

The two other tests that produced ERRORs in the earlier run pass fine when run individually from the command line. Nor do they show errors this time, so maybe they, too, were caused previously by the initial test_windows error.

The test_windows error again here is a time out, on line 152 of the file in this commit, "win1.focus()."

whimboo:
So the best here would be to reduce the number of tests you execute by just commenting out the other lines from the manifest files. Then also create a reduced test method in the test_windows.py file which only exercises the code causing us to fail. With that it will be much easier to investigate this problem.

galgeek:
I've moved, temporarily, to facilitate testing, the test that fails with focusmanger.testmode: true to a separate file, firefox_puppeteer/tests/test_windows_focus.py, and added this new test file to the manifest.

In the meantime, I'm not sure this is a priority to work on now.

whimboo:
FYI you do not have to move this test out, you can simply mark it as skipped via `@unittest.skip`.

galgeek:
You can skip a test using

`import unittest`

and then (an argument is required)

`@unittest.skip('your string here: a reason for skipping, perhaps')`

whimboo:
Looks great! I merged this PR as https://github.com/mozilla/firefox-ui-tests/commit/b671f9b253b930005ac6c533fe627b4aa9f6ec75

PR #68, opened 2015-02-05
add OS X support

galgeek:
Hi @whimboo!

A first attempt at configuring Travis multi-os for firefox-ui-tests...


.travis.yml
```diff
@@ -1,38 +1,28 @@
...
+        - "irc.mozilla.org#automation"
+
+matrix:
+    allow_failures:
+      - os: osx
+    exclude:
+      - os: linux
+        compiler: clang
+      - os: osx
+        compiler: gcc
```
whimboo: We don't have compilers as requirement, and you do not define those above. So I think this excludes can be removed.

.travis.yml
```diff
@@ -1,38 +1,28 @@
...
 #      email:
 #          - dev-automation@lists.mozilla.org
     irc:
-        - "irc.mozilla.org#automation"
\ No newline at end of file
+        - "irc.mozilla.org#automation"
+
+matrix:
+    allow_failures:
+      - os: osx
```
whimboo: Maybe adding a comment which type of failures we expect here? Are those known?

.travis/install.sh
```diff
@@ -0,0 +1,46 @@
...
+if [[ "$(uname -s)" == 'Darwin' ]]; then
+    DARWIN=true
+else
+    DARWIN=false
+fi
+
+if [[ "$DARWIN" = true ]]; then
+    brew update
+
+    # xvfb for OS X?
```
whimboo: Please try to only set the xfvb environment variable for linux. We might not need it for OS X. Please test.

.travis/install.sh
```diff
@@ -0,0 +1,46 @@
...
+
+    # flash via command line for OS X?
```
whimboo: You will have to download the DMG and get the package installed. It's not something we need right now, so lets defer flash also for Linux. Please file another bug for it.

.travis/install.sh
```diff
@@ -0,0 +1,46 @@
...
+
+# As long as we have invasive changes lets get the trunk version of marionette-client
+svn checkout https://github.com/mozilla/gecko-dev/trunk/testing/marionette/client
+cd client && sudo python setup.py develop && cd ..
+
+sudo python setup.py develop
+sudo pip install mozdownload pep8
+
+# Run pep8 on all except the checked out marionette-client folder
+pep8 --max-line-length=99 --exclude=client .
```
whimboo: This and the following code is not related to install. Please move it out of the script and back into the travis config. The install.sh file should only contain the commands which are necessary to prepare the system.

.travis/install.sh
```diff
@@ -0,0 +1,46 @@
...
+mozdownload -t daily
+
+if [[ "$DARWIN" = true ]]; then
+    hdiutil mount *firefox*.mac.dmg
+    sudo cp -R /Volumes/Nightly/FirefoxNightly.app /Applications
+    cd ~
+    hdiutil unmount /Volumes/Nightly/
+else
+    tar -xvf *firefox*.tar.bz2
+fi
```
whimboo: Lets make use of mozinstall here. You can install this package from pypi too. It will be x-platform.

whimboo: Maybe move the above install code into run.sh, so we have this all together.

.travis/run.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+set -e
+set -x
+if [[ "$(uname -s)" == "Darwin" ]]; then
```
whimboo: preserve the blank line please

.travis/run.sh
```diff
@@ -0,0 +1,8 @@
...
+    firefox-ui-tests --binary /Applications/FirefoxNightly.app/Contents/MacOS/firefox
+else
+    firefox-ui-tests --binary firefox/firefox
+fi
```
whimboo: I'm ok with it for now. Once our CLI command can handle installers (which will be soon) we can refactor this code.

whimboo:
FYI: For the first commit you do on a new branch lets use the bugzilla summary including the reviewer. That helps with squashing later.

.travis.yml
galgeek: Without the excludes, Travis defaults to run two test sets per os in this environment , one with compiler set to gcc, and one with compiler set to clang.

galgeek: It's better to exclude one by setting the compiler at the top of the file.

.travis.yml
galgeek: I've added a comment. Travis suggests adding this initially.

.travis/install.sh
galgeek: I wasn't able to get  mozinstall to work in .travis.yml or a script.

I've moved the firefox install code into install.sh.

galgeek:
I've updated this PR per the feedback above except where I've noted otherwise.

Travis has had some trouble today, and the Linux build has been failing while the OS X build succeeds. This may be due to Python and pip upgrades for the default Travis Linux builds. I'll check again Friday.

galgeek:
For the multi-site config, the Linux tests continue to fail, while the OS X tests pass.

I've sent this message to support@travis-ci.com (and subscribed to their mailing list for announcements of build updates):

It looks like our biggest problems when we use multi-os, specifying language C, are these errors early in our build:

/usr/local/lib/python2.7/dist-packages/setuptools-12.0.5-py2.7.egg/pkg_resources/__init__.py:2510: PEP440Warning: 'python-debian (0.1.21ubuntu1)' is being parsed as a legacy, non PEP 440, version. You may find odd behavior and sort order. In particular it will be sorted as less than 0.0. It is recommend to migrate to PEP 440 compatible versions.

/usr/local/lib/python2.7/dist-packages/setuptools-12.0.5-py2.7.egg/pkg_resources/__init__.py:2510: PEP440Warning: 'python-apt (0.8.3ubuntu7)' is being parsed as a legacy, non PEP 440, version. You may find odd behavior and sort order. In particular it will be sorted as less than 0.0. It is recommend to migrate to PEP 440 compatible versions.


Our Linux Python worker isn’t showing these errors, though it is showing less alarming PEP440 warnings for one of our own modules, and our tests there are passing.

... Maybe it’s possible for you to update these python-* packages for all C workers?

.travis.yml
```diff
@@ -1,38 +1,47 @@
...
 
-env: DISPLAY=:99 MOZ_XVFB=1
+compiler:
+    - gcc
 
 before_install:
-    - sudo apt-get update
-    - "/sbin/start-stop-daemon --start --quiet --make-pidfile --pidfile /tmp/custom_xvfb_99.pid --background --exec /usr/bin/Xvfb -- :99 -ac -screen 0 1024x768x24"
+    # set up linux only items
+    - ./.travis/preinstall.sh
```
whimboo: I would suggest to name the file equivalent to the travis step.

.travis.yml
```diff
@@ -1,38 +1,47 @@
...
 
-    # As long as we have invasive changes lets get the trunk version of marionette-client
+install:
+    # As long as we have invasive changes, lets get the trunk version of marionette-client
     - svn checkout https://github.com/mozilla/gecko-dev/trunk/testing/marionette/client
-    - "cd client && python setup.py develop && cd .."
+    - "cd client && sudo python setup.py develop && cd .."
 
-    - python setup.py develop
+    - sudo python setup.py develop
```
whimboo: Why sudo in both of the above cases? I don't see why we cannot install all of our packages in the newly created venv.

.travis.yml
```diff
@@ -1,38 +1,47 @@
...
-    - sudo apt-get update
-    - "/sbin/start-stop-daemon --start --quiet --make-pidfile --pidfile /tmp/custom_xvfb_99.pid --background --exec /usr/bin/Xvfb -- :99 -ac -screen 0 1024x768x24"
+    # set up linux only items
+    - ./.travis/preinstall.sh
 
-install:
-    - sudo apt-get install flashplugin-nonfree subversion
+    - sudo -H pip install virtualenv
+    - virtualenv ~/.venv
+    - source ~/.venv/bin/activate
```
whimboo: Why can't we put this all into the appropriate bash script?

.travis.yml
```diff
@@ -1,38 +1,47 @@
...
     - pep8 --max-line-length=99 --exclude=client .
 
-    # Download Firefox Nightly which is compatible with the used version of
-    # marionette client
+    # Download Firefox Nightly which is compatible with the Marionette client version used 
     - mozdownload -t daily
-    - tar -xvf *firefox*.tar.bz2
+
+    # install Firefox Nightly per os
+    - ./.travis/install.sh
```
whimboo: This is before_script, so please update the file name. Also the mozdownload line should be moved into that file. The PEP8 one I would leave here.

.travis/install.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+set -ev
+
+if [ "$TRAVIS_OS_NAME" == "osx" ]; then
+    hdiutil mount *firefox*.mac.dmg
+    sudo cp -R /Volumes/Nightly/FirefoxNightly.app /Applications
+    cd ~
+    hdiutil unmount /Volumes/Nightly/
```
whimboo: This way please: http://mxr.mozilla.org/comm-central/source/mozilla/testing/mozbase/mozinstall/mozinstall/mozinstall.py#240. There is no sudo necessary at all. The mounted DMG is in the user scope.

Also mozinstall works pretty fine for me via `mozinstall *firefox*.dmg` if there is only one single match in this folder. So please update it. Further also export the output path as an environment variable which can be re-used by the run section. That way we would not have to make a distinction between Linux and OS X.

.travis/run.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+set -ev
+
+if [ "$TRAVIS_OS_NAME" == "osx" ]; then
+    firefox-ui-tests --binary /Applications/FirefoxNightly.app/Contents/MacOS/firefox
+else
+    firefox-ui-tests --binary firefox/firefox
+fi
```
whimboo: I don't see that we need this file once we use mozinstall.

.travis.yml
galgeek: It's cleanest to put the lines creating and activating the virtualenv at the top level, in the .travis..yml. They're cross-platform.

I've moved the virtualenv install, needed for linux only, to the before_install.sh script, where I'm using apt-get instead of pip.

galgeek:
Hi, @whimboo!

I've updated the PR per your feedback (thank you!) and a tip from asarih on #travis regarding environment variables.

Both the Linux and the OS X tests are passing again.

.travis.yml
```diff
@@ -1,18 +1,26 @@
...
 
-python:
-    - 2.7
+os:
+    - osx
+    - linux
 
-env: DISPLAY=:99 MOZ_XVFB=1
+compiler:
+    - gcc
```
whimboo: out of interest, which part of the setup routine needs gcc?

.travis.yml
```diff
@@ -1,18 +1,26 @@
...
+
+env:
+    - DISPLAY=$(if [[ $TRAVIS_OS_NAME = 'linux' ]]; then echo ':99'; fi)
+      MOZ_XVFB=$(if [[ $TRAVIS_OS_NAME = 'linux' ]]; then echo 1; fi)
+      BINARY=$(if [[ $TRAVIS_OS_NAME = 'osx' ]]; then echo '/Applications/FirefoxNightly.app/Contents/MacOS/firefox'; else echo 'firefox/firefox'; fi)
```
whimboo: Why can't we define all those variables in the before_install script, where we already do the distinction? It should be just an `export DISPLAY=":99"` in the Linux case, and the same for the others. Also the binary environment variable you should set to the path you get out from mozinstall.

.travis.yml
```diff
@@ -1,18 +1,26 @@
...
 
 before_install:
-    - sudo apt-get update
-    - "/sbin/start-stop-daemon --start --quiet --make-pidfile --pidfile /tmp/custom_xvfb_99.pid --background --exec /usr/bin/Xvfb -- :99 -ac -screen 0 1024x768x24"
+    # set up linux only items
```
whimboo: Please remove this comment. Once we have to do setup stuff for OS X we shouldn't have to do any changes here.

.travis.yml
```diff
@@ -23,16 +31,15 @@ before_script:
     # Run pep8 on all except the checked out marionette-client folder
     - pep8 --max-line-length=99 --exclude=client .
 
-    # Download Firefox Nightly which is compatible with the used version of
-    # marionette client
-    - mozdownload -t daily
-    - tar -xvf *firefox*.tar.bz2
+    # install Firefox Nightly per os
```
whimboo: Same here for the comment.

.travis.yml
```diff
@@ -23,16 +31,15 @@ before_script:
...
+    - ./.travis/before_script.sh
 
 script:
-    - firefox-ui-tests --binary firefox/firefox
+    #- ./.travis/run.sh
```
whimboo: Please remove this obsolete line.

.travis/before_install.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -ev
+
+if [ "$TRAVIS_OS_NAME" == "linux" ]; then
+    /sbin/start-stop-daemon --start --quiet --make-pidfile --pidfile /tmp/custom_xvfb_99.pid \
+         --background --exec /usr/bin/Xvfb -- :99 -ac -screen 0 1024x768x24
+
+    sudo apt-get update -qq
+    sudo apt-get install python-pip python-virtualenv
+fi
```
whimboo: This file looks fine.

.travis/before_script.sh
```diff
@@ -0,0 +1,15 @@
...
+mozdownload -t daily
+
+if [ "$TRAVIS_OS_NAME" == "osx" ]; then
+    hdiutil attach -nobrowse -noautoopen *firefox*.mac.dmg
+    cp -R /Volumes/Nightly/FirefoxNightly.app /Applications
+    cd ~
+    hdiutil unmount /Volumes/Nightly/
+else
+    tar -xvf *firefox*.tar.bz2
+fi
```
whimboo: s already said in my last review I do not have any problems with mozinstall. So we should use that, set the BINARY environment variable via "BINARY=`mozinstall *firefox*`. Then this file is not necessary anymore. Please test this again and if failures occur let me know so I can have a look at. If there is really a blocker we can move this out to a follow-up bug.

.travis.yml
galgeek: Nothing in our build needs gcc.

For Travis multi-os for our Python tests, we need to use one of Travis's C workers, and if you don't specify any compiler, it will create builds for both gcc and clang, for both Linux and OS X. So I added the compiler specification simply to restrict our builds to one each for Linux and OS X.

.travis.yml
galgeek: Environment variables that you want to set on a per-OS basis and use in .travis.yml must go in its env: section for now. That's true for BINARY

I lean toward putting things  in .travis.yml whenever I can since it's the master config file, and this reduces the need to reference other files.

galgeek:
Hi, @whimboo!

Thank you for your feedback!  This update uses mozinstall and, I think, otherwise addresses your concerns.

.travis.yml
whimboo: I see. Lets make a comment out of this information, so that it is clear at this point. Thanks for that info!

.travis.yml
```diff
@@ -1,38 +1,43 @@
...
     - python setup.py develop
-    - pip install mozdownload pep8
+    - pip install mozdownload mozinstall pep8
 
 before_script:
     # Run pep8 on all except the checked out marionette-client folder
     - pep8 --max-line-length=99 --exclude=client .
 
-    # Download Firefox Nightly which is compatible with the used version of
-    # marionette client
```
whimboo: Please leave the comment.

.travis.yml
```diff
@@ -1,38 +1,43 @@
...
+    - linux
 
-env: DISPLAY=:99 MOZ_XVFB=1
+compiler:
+    - gcc
+
+env:
+    - DISPLAY=$(if [[ $TRAVIS_OS_NAME = 'linux' ]]; then echo ':99'; fi)
+      MOZ_XVFB=$(if [[ $TRAVIS_OS_NAME = 'linux' ]]; then echo 1; fi)
+      BINARY=$(if [[ $TRAVIS_OS_NAME = 'osx' ]]; then echo 'FirefoxNightly.app/Contents/MacOS/firefox'; else echo 'firefox/firefox'; fi)
```
whimboo: Ok, then lets figure it out later how to better do it. It's certainly not a blocker right now to get rid of this env variable right now.

.travis.yml
```diff
@@ -1,38 +1,43 @@
...
 before_script:
     # Run pep8 on all except the checked out marionette-client folder
     - pep8 --max-line-length=99 --exclude=client .
 
-    # Download Firefox Nightly which is compatible with the used version of
-    # marionette client
-    - mozdownload -t daily
-    - tar -xvf *firefox*.tar.bz2
+    - "cd && mozdownload -t daily"
+    - mozinstall *-firefox-*
```
whimboo: It looks great that this works now! Thanks for doing it Barbara!

.travis.yml
```diff
@@ -1,38 +1,43 @@
...
 
 script:
-    - firefox-ui-tests --binary firefox/firefox
+    - firefox-ui-tests --binary $BINARY
 
 notifications:
 #      email:
 #          - dev-automation@lists.mozilla.org
```
whimboo: Please make sure to uncomment that part.

galgeek:
Hi @whimboo!

Thank you for your feedback!  I've updated the PR.

whimboo:
Looks great. I got it merged into master as https://github.com/mozilla/firefox-ui-tests/commit/d74bf9bedb8d520ce9eecdc97769ad6d90a4518f. @galgeek in the future please use "r=whimboo" for the commit message given that by the rules it should be the IRC nickname. We handled that a bit loosely in the past. :)

PR #69, opened 2015-02-05
Bug 1121731 - Port places library from mozmill-tests repository

whimboo:
Just a WIP without tests. Those are still to come soon.


whimboo:
@chmanchester please be aware that you have commented on a commit of mine but not on the PR. You may have looked at older code.

firefox_puppeteer/api/places.py
```diff
@@ -0,0 +1,156 @@
...
+            # TODO: In case of a timeout clean-up the registered topic
+            pass
+
+    def wait_for_visited(self, urls, callback):
+        """Waits until all passed-in urls have been visited."""
+        # Bug 1121691: Needs observer handling support with callback first
+        # Until then we have to wait about 4s to ensure the page has been indexed
+        callback()
+        from time import sleep
+        sleep(4)
```
chmanchester: The callback should probably go after the sleep. I don't see tests for this method.

firefox_puppeteer/tests/test_places.py
```diff
@@ -0,0 +1,80 @@
...
+                     ]
+
+    def tearDown(self):
+        try:
+            self.places.restore_default_bookmarks()
+            self.places.remove_all_history()
+        finally:
+            FirefoxTestCase.tearDown(self)
+
+    def get_all_urls_in_history(self):
```
chmanchester: Is this the sort of thing we'd want in a library itself?

firefox_puppeteer/api/places.py
```diff
@@ -0,0 +1,156 @@
...
+                // Remove the pages, then block until we're done or until timeout is reached
+                Services.obs.addObserver(observer, 'places-expiration-finished', false);
+
+                hs.removeAllPages();
+            """, script_timeout=10000)
+        except TimeoutException:
+            # TODO: In case of a timeout clean-up the registered topic
+            pass
+
+    def wait_for_visited(self, urls, callback):
```
chmanchester: For my own benefit, when is this needed? The test below makes me think we can just navigate to pages and then check they're in the history.

firefox_puppeteer/api/places.py
whimboo: No, triggering the navigate() method has to happen first. Then we have to wait 4s until those urls have been visited. Good catch regarding the test. Will do it in a bit.

firefox_puppeteer/tests/test_places.py
whimboo: None of our existing mozmill tests are using that, so I would keep it here for now if that is ok. If needed we could move this later on.

firefox_puppeteer/api/places.py
whimboo: This is not about a page being in history. It somewhat needs to be visited and indexed. This will not happen immediately when you navigate to the page, but in about 4s intervals. If you do not wait that long, the autocomplete in the locationbar will not show up.

firefox_puppeteer/tests/test_places.py
chmanchester: Sure, ok.

firefox_puppeteer/api/places.py
chmanchester: Ok. In that case I don't see any benefit to passing a callback to this function.

firefox_puppeteer/api/places.py
```diff
@@ -0,0 +1,156 @@
...
+
+class Places(BaseLib):
+    """Low-level access to several bookmark and history related actions."""
+
+    bookmark_folders = {'placesRoot': 1,
+                        'bookmarksMenuFolder': 2,
+                        'toolbarFolder': 3,
+                        'tagsFolder': 4,
+                        'unfiledBookmarksFolder': 5,
+                        }
```
chmanchester: I would use a named tuple or class variables for these constants.

firefox_puppeteer/api/places.py
whimboo: We need that once we can use observers as indicated in the comment above. I do not want an API change later, and to update all the tests.

firefox_puppeteer/api/places.py
whimboo: Just to be clear... you mean 5 class variables for each of those constants?

firefox_puppeteer/api/places.py
chmanchester: Looking at the way this is in the mozmill version, can we do something along the lines of:
```javascript
let urls = arguments[0];
let visited = [];
obs = {
  onVisit: function (aURI) {
    let spec = aURI.spec;
    if (urls.indexOf(spec) >= 0 && visited.indexOf(spec) == -1) {
      visited.push(spec);
      if (spec.length == urls.length) {
        historyService.removeObserver(obs);
        marionetteScriptFinished(true);
      }
    }
  }
}
historyService.addObserver(obs, false);
```

whimboo: No because you will have to trigger loading the URLs. And that has to
happen after registering the observer topics.
Am 05.02.2015 23:38 schrieb "chmanchester" <notifications@github.com>:

chmanchester: If it's just a matter of keeping the observer around to unregister it, isn't this an instance we can use execute script with newSandbox=False and stash the observer? I wouldn't push this, but I really think it's preferable to adding a 4 second sleep, which will make developing and running these tests significantly less convenient.

chmanchester: I guess we'd also have to set a property on the sandbox to poll for, as well.

whimboo: As we have discussed a couple of weeks ago, this is basic observer code, which needs a fix for bug 1121691 first. So once that other bug is fixed we can work on updating all the other places where we have to use observers.

chmanchester: During various discussions I've offered workarounds that will be effective for cases like these, and fixed bugs in marionette (with a degree of urgency) to support these cases. If you want to implement this later I guess that's alright, but it doesn't make a lot of sense to me to block all these tests from getting converted for the sake of API compatibility via a four second sleep.

firefox_puppeteer/api/places.py
chmanchester: I mean one class variable that is a named tuple. Dictionaries are good for mapping arbitrary values to arbitrary things, this is more like an enumeration, so I'd like to be able to refer to something like places.bookmark_folders.placesRoot in code.

firefox_puppeteer/api/places.py
```diff
@@ -0,0 +1,161 @@
...
+
+    # Bookmark related helpers #
+
+    def is_bookmarked(self, url):
+        """Checks if the given URL is bookmarked.
+
+        :param url: The URL to Check
+
+        :returns: True, if the URL has been found
```
chmanchester: This comment should mention that the found URL is bookmarked.

firefox_puppeteer/api/places.py
```diff
@@ -0,0 +1,161 @@
...
+
+          return results.length == 1;
+        """, script_args=[url])
+
+    def get_folder_ids_for_url(self, url):
+        """Retrieves the folder ids where the given URL has been bookmarked in.
+
+         :param url: URL of the bookmark
+
+         :returns: List of folder ids
```
chmanchester: Is it correct style to have a line break between params and returns in these comments?

firefox_puppeteer/api/places.py
```diff
@@ -0,0 +1,161 @@
...
+
+              let observer = {
+                observe: function (aSubject, aTopic, aData) {
+                  Services.obs.removeObserver(observer, "bookmarks-restore-success");
+
+                  marionetteScriptFinished(true);
+                }
+              };
+
+              // Fire off an initially import
```
chmanchester: Delete or clarify this comment.

firefox_puppeteer/api/places.py
whimboo: This method is exactly here to unblock everyone to get those tests written and to not force us to change all of them later again! The sleep of 4 seconds would be needed in any case when you want to use the auto-complete dropdown.

Until now I simply have had no time to get started on observers and events. That's the only reason why we don't have them yet.

firefox_puppeteer/api/places.py
whimboo: All the other libraries also have this style, so I just followed it. Personally I find it better with separation, but we can also globally remove that again. I'm open to that.

firefox_puppeteer/api/places.py
whimboo: Sorry, but I still don't see what what you want here. Can you please give me an example? Not sure if its me or you who is confused, but what we need is that when you access e.g `bookmark_folder.places_root` it should be equivalent to `1`, because that is the internal id of this folder in Places.

chmanchester: Something like:

>>> from collections import namedtuple
>>> t = namedtuple('bookmarks_folders', ['places_root', 'bookmarks_menu', 'toolbar', 'tags', 'unfiled'])
>>> bookmarks_folders = t(1, 2, 3, 4, 5)
>>> bookmarks_folders.places_root
1
>>>

whimboo:
PR has been merged as https://github.com/mozilla/firefox-ui-tests/commit/8c2dad20dc465795025565393cb364f9eaa3bca8.

PR #70, opened 2015-02-05
Bug 1129099 - Convert test_access_location_bar to a firefox ui test.;r=c...

mwobensmith:
...hmanchester


firefox_ui_tests/functional/toolbar/test_access_location_bar.py
```diff
@@ -0,0 +1,79 @@
...
+        # Navigate to several urls in order to populate history
+        self.test_urls = [
+            'layout/mozilla_projects.html',
+            'layout/mozilla_mission.html',
+            'layout/mozilla.html']
+        self.test_urls = [self.marionette.absolute_url(t)
+                          for t in self.test_urls]
+
+        for url in self.test_urls:
+            self.location_bar.load_url(url)
```
chmanchester: The mozmill test uses wait_for_visited at this stage.

firefox_ui_tests/functional/toolbar/test_access_location_bar.py
```diff
@@ -0,0 +1,79 @@
...
+        self.url_bar.send_keys(Keys.ARROW_DOWN)
+
+        # Verify that autocomplete is open
+        def auto_complete_open(mn):
+            return self.location_bar.autocomplete_results.is_open
+        self.wait_for_condition(auto_complete_open)
+
+        # Verify that results are displayed in autocomplete
+        def auto_complete_results(mn):
+            self.location_bar_value_before = self.location_bar.value
```
chmanchester: This line looks extraneous.

firefox_ui_tests/functional/toolbar/test_access_location_bar.py
```diff
@@ -0,0 +1,79 @@
...
+
+        # Verify that first item in list is selected
+        def auto_complete_selected_item(mn):
+            self.results = self.location_bar.autocomplete_results.results
+            return self.results.get_attribute('selectedIndex') == '0'
+        self.wait_for_condition(auto_complete_selected_item)
+
+        # Verify that selected item populates location bar with url
+        def check_url_from_selection(mn):
+            return self.location_bar.value == self.marionette.absolute_url('layout/mozilla.html')
```
chmanchester: We may be able to make this check synchronous if it's implied by the wait above.

firefox_ui_tests/functional/toolbar/test_access_location_bar.py
```diff
@@ -0,0 +1,79 @@
...
+        self.wait_for_condition(check_url_from_selection)
+        self.url_bar.send_keys(Keys.ENTER)
+
+        # Verify that selected url is loaded by verifying the title of the page has changed
+        def check_title(mn):
+            self.new_page_title = self.marionette.execute_script("return document.title;")
+            return self.new_page_title == 'Mozilla'
```
chmanchester: A check against the loaded url (self.marionette.get_url()) or the value of the location bar might be closer to the mozmill version.

firefox_ui_tests/functional/toolbar/manifest.ini
```diff
@@ -1,2 +1,3 @@
 [test_back_forward.py]
 [test_home_button.py]
+[test_access_location_bar.py]
```
whimboo: We want to put those tests under functional/locationbar. Please use `git mv` to get the test moved over, and the manifest updated. Please also ensure you sort the tests alphabetical.

firefox_ui_tests/functional/toolbar/test_access_location_bar.py
```diff
@@ -0,0 +1,79 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from marionette import By
+
+from firefox_puppeteer.api.keys import Keys
+from firefox_puppeteer.api.places import Places
+from firefox_puppeteer.ui.windows import BrowserWindow
```
whimboo: There is no need to import those classes. Keys and Places you can reach via self.keys and self.places, and BrowserWindow is not in use in this test.

firefox_ui_tests/functional/toolbar/test_access_location_bar.py
whimboo: I wonder if we should move those lines into the test itself, and only keep declarations in the setUp method. Opening the different locations was also part of the former test.

firefox_ui_tests/functional/toolbar/test_access_location_bar.py
```diff
@@ -0,0 +1,79 @@
...
+        for url in self.test_urls:
+            self.location_bar.load_url(url)
+
+        with self.marionette.using_context('content'):
+            self.marionette.navigate('about:blank')
+
+    @skip_under_xvfb
+    def test_access_location_bar_history(self):
+        # Need to blur url bar or autocomplete won't load - bug 1038614
+        self.marionette.execute_script("arguments[0].blur();", script_args=[self.url_bar])
```
whimboo: Please put the script content into triple double-quotes.

firefox_ui_tests/functional/toolbar/test_access_location_bar.py
```diff
@@ -0,0 +1,79 @@
...
+        # Clear contents of url bar to focus
+        self.location_bar.clear()
+
+        # Arrow down to navigate list of visited sites
+        self.url_bar.send_keys(Keys.ARROW_DOWN)
+
+        # Verify that autocomplete is open
+        def auto_complete_open(mn):
+            return self.location_bar.autocomplete_results.is_open
+        self.wait_for_condition(auto_complete_open)
```
whimboo: This is simple enough. So I would propose that you make use of a lambda function here.  Also try to combine comments and code-blocks, so we have something like an action and checks in a single one.

firefox_ui_tests/functional/toolbar/test_access_location_bar.py
```diff
@@ -0,0 +1,79 @@
...
+
+        # Verify that results are displayed in autocomplete
+        def auto_complete_results(mn):
+            self.location_bar_value_before = self.location_bar.value
+            return len(self.location_bar.autocomplete_results.visible_results) > 1
+        self.wait_for_condition(auto_complete_results)
```
whimboo: lambda method here too?

firefox_ui_tests/functional/toolbar/test_access_location_bar.py
```diff
@@ -0,0 +1,79 @@
...
+
+        # Arrow down again to select first item in list
+        self.url_bar.send_keys(Keys.ARROW_DOWN)
+
+        # Verify that first item in list is selected
+        def auto_complete_selected_item(mn):
+            self.results = self.location_bar.autocomplete_results.results
+            return self.results.get_attribute('selectedIndex') == '0'
```
whimboo: We should work with the visible results here too. Also I would propose you add a new property called `selected_index` to the AutoCompleteResults class, which is doing that for you.

firefox_ui_tests/functional/toolbar/test_access_location_bar.py
```diff
@@ -0,0 +1,79 @@
...
+            return self.location_bar.autocomplete_results.is_open
+        self.wait_for_condition(auto_complete_open)
+
+        # Verify that results are displayed in autocomplete
+        def auto_complete_results(mn):
+            self.location_bar_value_before = self.location_bar.value
+            return len(self.location_bar.autocomplete_results.visible_results) > 1
+        self.wait_for_condition(auto_complete_results)
+
+        # Arrow down again to select first item in list
```
whimboo: Please also add in which order the results are displayed. The comment in the Mozmill test contains more information and such we should keep to make the test understandable.

firefox_ui_tests/functional/toolbar/test_access_location_bar.py
```diff
@@ -0,0 +1,79 @@
...
+        self.url_bar = self.location_bar.urlbar
+
+        # Purge history
+        self.places.remove_all_history()
+
+        # Navigate to several urls in order to populate history
+        self.test_urls = [
+            'layout/mozilla_projects.html',
+            'layout/mozilla_mission.html',
+            'layout/mozilla.html']
```
whimboo: Please use the `mission` page last. All the others also contain `mozilla`, so the test would be a bit flaky.

firefox_ui_tests/functional/toolbar/test_access_location_bar.py
whimboo: We do not have to check for the full URL being displayed in the location bar, but only for the unique term `mission`.

firefox_ui_tests/functional/toolbar/test_access_location_bar.py
```diff
@@ -0,0 +1,79 @@
...
+from firefox_ui_harness.decorators import skip_under_xvfb
+from firefox_ui_harness.testcase import FirefoxTestCase
+
+
+class TestAccessLocationBar(FirefoxTestCase):
+
+    def setUp(self):
+        FirefoxTestCase.setUp(self)
+        self.location_bar = self.browser.navbar.locationbar
+        self.url_bar = self.location_bar.urlbar
```
whimboo: self.url_bar is used only once. So this variable is not that useful.

firefox_ui_tests/functional/toolbar/test_access_location_bar.py
whimboo: That is correct. We do not want to check the window or tab title here but the value as displayed in the locationbar.

mwobensmith: The original test looks for the Mozilla logo in the page, which currently does not exist in the codebase and appears as a broken image. Hence, I chose the page title to verify that the proper HTML page was retrieved. The test that you are asking me to write was already performed in comment 71.

firefox_ui_tests/functional/toolbar/test_access_location_bar.py
mwobensmith: Please make a decision - keep this in the setUp method or move to test.

firefox_ui_tests/functional/toolbar/test_access_location_bar.py
mwobensmith: It is, I will remove.

firefox_ui_tests/functional/toolbar/manifest.ini
mwobensmith: The test is in the current directory because that's what you requested when we asked you during the training. I will move it.

firefox_ui_tests/functional/toolbar/test_access_location_bar.py
mwobensmith: > We should work with the visible results here too. 

This is vague. What are you asking me to do? 

RE: add a property to your framework. This is my first test. I think you are asking a lot.

firefox_ui_tests/functional/toolbar/test_access_location_bar.py
mwobensmith: I do not understand this comment. What difference does the order make in this test?

firefox_ui_tests/functional/toolbar/test_access_location_bar.py
mwobensmith: self.url_bar is used 4 more times in this test after the initial declaration.

firefox_ui_tests/functional/toolbar/test_access_location_bar.py
mwobensmith: s/comment 71/line 71

firefox_ui_tests/functional/toolbar/test_access_location_bar.py
mwobensmith: Original test uses a mozmill method "locationBar.contains" to check this string. I don't see this functionality in the current framework. Without this method, we can still do a string compare on the entire URL, and I don't see why it's an issue.

firefox_ui_tests/functional/toolbar/test_access_location_bar.py
whimboo: If the image is missing, please add it from http://hg.mozilla.org/qa/mozmill-tests/file/tip/data/images.

firefox_ui_tests/functional/toolbar/test_access_location_bar.py
whimboo: Please move it into the test.

firefox_ui_tests/functional/toolbar/test_access_location_bar.py
whimboo: You are using `autocomplete_results.results` here. It's different compare to line 57 where `autocomplete_results.visible_results` is used. 

Adding a property to the class is pretty simple. Trust me. :) We already have a couple of those, so you only have to copy the block and update the property name and function content. The latter is what you have above. You can also @chmanchester when I'm not around, also regarding a test for it in test_toolbars.py.

firefox_ui_tests/functional/toolbar/test_access_location_bar.py
whimboo: This is the order in how pages are getting loaded, and the awesome bar suggests autocomplete entries in a reversed order from loading - means newest first. By re-ordering those pages you are changing the test logic.

firefox_ui_tests/functional/toolbar/test_access_location_bar.py
whimboo: So leave it in.

firefox_ui_tests/functional/toolbar/test_access_location_bar.py
whimboo: You will have `self.assertIn(a, b)` which checks if a is contained in b. Keep in mind that the locationbar does not show the `http://` prefix, so this test should fail. I wondered why it is not and after a check I can see it. The reason is that `self.location_bar.value` returns the real URL but not the content as shown in the locationbar. For your case I would suggest to leave what you have.

@chmanchester I would suggest we get a property like `self.location_bar.text` implemented as a follow-up. What do you think?

chmanchester: This is the implementation in mozmill, which also has the same behavior: http://hg.mozilla.org/qa/mozmill-tests/file/default/firefox/lib/toolbars.js#l719

We can add an attribute for just the displayed test if that seems valuable, but I don't think it makes sense to do as a part of this test.

firefox_ui_tests/functional/toolbar/test_access_location_bar.py
mwobensmith: Current test logic of checking page title is functionally the same as checking for a loaded image. Both verify that the correct page is loaded. I respectfully ask that you fix this issue on your end, as you will clearly have many other problems if you have not migrated your old codebase in its entirety, including this images directory.

firefox_ui_tests/functional/toolbar/test_access_location_bar.py
mwobensmith: I've been asked to migrate a test case. I do not have time to work on your framework. Please add the property and I will update my test to use it. Thank you.

chmanchester: Granted we end up with a test that provides equivalent coverage, I don't think we can consider adding something to a library as a strict prerequisite to test conversion.

However, if we insist on a division between converting tests and working on libraries in every case, I think we lose a lot. The nature of the mozmill tests is that they are tightly tied to mozmill libraries (and mozmill libraries don't receive a lot of testing outside of the tests that consumer them) -- in many cases, I don't think considering the conversion of a test and the conversion of a library as separate work will make a lot of sense.

mwobensmith:
I've incorporated as many of the requested changes into the test as I could. I am now handing it off to you for completion as I do not have time to proceed further.

whimboo:
@mwobensmith thanks for working on this test. Lets see who could actually pick it up to finish it off. @galgeek would you be interested in?

firefox_ui_tests/functional/toolbar/test_access_location_bar.py
whimboo: It's not that easy to find a solution for that problem  @chmanchester. I can agree that there might be lesser time for someone or also less of degree in working on libraries. But as you say we cannot convert everything ahead of time. So the question is to find the right set of people who can help with the transition of the tests. As it looks like QA people do not have the appropriate time.

galgeek:
@whimboo, I could take it on. My list is increasing. You may want to help me prioritize.

whimboo:
This has been merged about 2 weeks ago via PR #80.

PR #72, opened 2015-02-05
Updated test to open browser from menu

rbillings:
https://bugzilla.mozilla.org/show_bug.cgi?id=1129652

### DIFF
--- a/firefox_puppeteer/tests/test_windows.py
+++ b/firefox_puppeteer/tests/test_windows.py
@@ -77,11 +77,14 @@ class TestWindows(FirefoxTestCase):
                           win1.send_shortcut, 'l', acel=True)
 
     def test_base_window_open_close(self):
+        def setUp(self):
+            FirefoxTestCase.setUp(self)
+
         # force BaseWindow instance
         win1 = BaseWindow(lambda: self.marionette, self.browser.handle)
 
         # Open a new window (will be focused), and check states
-        win2 = win1.open_window()
+        win2 = self.browser.open_browser(trigger='menu')
 
         # force BaseWindow instance
         win2 = BaseWindow(lambda: self.marionette, win2.handle)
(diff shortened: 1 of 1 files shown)


firefox_puppeteer/tests/test_windows.py
```diff
@@ -77,11 +77,14 @@ def test_base_window_basics(self):
...
     def test_base_window_open_close(self):
+        def setUp(self):
+            FirefoxTestCase.setUp(self)
+
         # force BaseWindow instance
         win1 = BaseWindow(lambda: self.marionette, self.browser.handle)
 
         # Open a new window (will be focused), and check states
-        win2 = win1.open_window()
+        win2 = self.browser.open_browser(trigger='menu')
```
whimboo: Please don't change this line but only the test for the browser window. BaseWindow does not support any kind of trigger beside a callback.

firefox_puppeteer/tests/test_windows.py
```diff
@@ -171,7 +174,7 @@ def test_browser_window_basic(self):
 
     def test_browser_window_open_close(self):
         # open and close a new browser windows by menu
-        win2 = self.browser.open_browser()
+        win2 = self.browser.open_browser(trigger='menu')
         self.assertEquals(win2, self.windows.current)
         self.assertFalse(self.browser.is_private)
         win2.close()
```
whimboo: Do the same for close() and it will be fine.

firefox_puppeteer/tests/test_windows.py
whimboo: Your comment for the commit mentions that you have removed the menu trigger, but it is still present. :) So please revert this line to what it was before. We don't want to operate on a browser window in a base window test.

whimboo:
@rbillings thanks for the update. It looks fine now. Maybe you can add a comment in the future when you have pushed new commits. Reason is that github doesn't sent out any emails to notify reviewers. :/

I will get the branch squashed and merged in a moment.

whimboo:
Merged as: https://github.com/mozilla/firefox-ui-tests/commit/f30fd0fea8f79ff2cdb8bbece2d6c66937e30cf9. Thanks!

PR #74, opened 2015-02-06
Bug 1131260 - Mozmill conversion: Test the location bar go button

rbillings:


firefox_puppeteer/tests/test_toolbars.py
```diff
@@ -28,6 +28,15 @@ def test_focus_and_clear(self):
         locationbar.clear()
         self.assertEqual(locationbar.value, '')
 
+    def test_gobutton(self):
```
whimboo: For unit level tests of properties we usually only check some of its attributes. Mainly the `localName`. See https://github.com/rbillings/firefox-ui-tests/blob/gobutton/firefox_puppeteer/tests/test_toolbars.py#L48 as reference.

So the test as written here is way more complex and should really end-up in firefox_ui_tests/functional.

firefox_puppeteer/tests/test_toolbars.py
```diff
@@ -28,6 +28,15 @@ def test_focus_and_clear(self):
...
+        # This replaces http://hg.mozilla.org/qa/mozmill-tests/file/default/firefox/tests/functional/testAwesomeBar/testGoButton.js
```
whimboo: This line is too long and travis fails because of it. Maybe do it like Bob here:
https://github.com/rbillings/firefox-ui-tests/blob/gobutton/firefox_ui_tests/functional/awesomebar/test_star_in_autocomplete.py#L11

firefox_puppeteer/tests/test_toolbars.py
```diff
@@ -28,6 +28,15 @@ def test_focus_and_clear(self):
...
+        input_text = 'mozilla.org/'
+        locationbar = self.browser.navbar.locationbar
+        locationbar.urlbar.send_keys(input_text)
+        go_button = self.browser.navbar.go_button
+        self.assertEqual(locationbar.value, 'mozilla.org/')
+        self.assertFalse(go_button.is_displayed())
```
whimboo: The above lines do not match the test we have in Mozmill:

* it doesn't check that the button is hidden if no input has been given
* it does not check the functionality of the button by clicking it, that the target page has been loaded, and the button is hidden again

rbillings:
oh crud - it looks like all of the interim changes went in along with my fixes

whimboo:
What you wanna do is a rebase against your local master branch once it has been updated for latest upstream master changes. Then those additional commits will go away.

firefox_puppeteer/ui/toolbars.py
```diff
@@ -31,6 +31,14 @@ def forward_button(self):
         return self.marionette.find_element(By.ID, 'forward-button')
 
     @property
+    def go_button(self):
+        """ Provides access to the go button from the navbar ui.
```
whimboo: nit: this button is inside the location bar. I think its fine to remove all behind `button`.

whimboo: So this property should be part of the LocationBar class but not NavBar class.

firefox_puppeteer/ui/toolbars.py
```diff
@@ -137,6 +145,14 @@ def get_contextmenu_entry(self, action):
         return found[0] if len(found) else None
 
     @property
+    def go_button(self):
+        """Provides access to the DOM element go button.
+
+        :returns: Reference to the go button.
+        """
+        return self.marionette.find_element(By.ID, 'go_button')
```
whimboo: I'm not sure what this ID refers to. I think you accidentally put in the property name of the `go_button` you added to the NavBar class. This wont work because the property is not part of the DOM.

firefox_ui_tests/functional/locationbar/test_go_button.py
```diff
@@ -0,0 +1,36 @@
...
+from firefox_ui_harness.testcase import FirefoxTestCase
+
+from marionette_driver import By
+
+
+class TestGoButton(FirefoxTestCase):
+    """ This replaces
+    This replaces http://hg.mozilla.org/qa/mozmill-tests/file/default/firefox/tests/functional/testAwesomeBar/testGoButton.js
+    Check a go button displays after text has been added to the awesome bar, takes user to correct results.
+    """
```
whimboo: You can drop this docstring. It doesn't give us any value.

firefox_ui_tests/functional/locationbar/test_go_button.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def setUp(self):
+        FirefoxTestCase.setUp(self)
+
+    def test_gobutton(self):
+        # no input, go button should not display
+        go_button = self.browser.navbar.go_button
```
whimboo: As mentioned before the go button is part of the location bar class. So this needs an update.

firefox_ui_tests/functional/locationbar/test_go_button.py
```diff
@@ -0,0 +1,36 @@
...
+        self.assertFalse(go_button.is_displayed())
```
whimboo: The `is_displayed()` method is not working reliable and we got told to not use it. So here it would be better to use `button.get_attribute('foobar')` whereby foobar is the name of the js property which returns the visibility state of the button.

firefox_ui_tests/functional/locationbar/test_go_button.py
```diff
@@ -0,0 +1,36 @@
...
+
+        # add input, go button displays, click go button
+        input_text = 'mozilla.org/'
```
whimboo: This test is part of the functional testrun, which means we do not allow any remote website here. Please use the same URLs as in the Mozmill test. Those pages already exist. See https://github.com/mozilla/firefox-ui-tests/blob/master/firefox_ui_tests/functional/locationbar/test_access_locationbar.py#L18 how to define that in the setUp method.

firefox_ui_tests/functional/locationbar/test_go_button.py
```diff
@@ -0,0 +1,36 @@
...
+        locationbar = self.browser.navbar.locationbar
+        locationbar.urlbar.send_keys(input_text)
+        go_button = self.marionette.find_element(By.ID, 'gobutton')
```
whimboo: You already have a reference to the go button from above. For chrome elements we should always have a property in a ui library, so that there is no need to use `find_element` in a test. This only adds complexity if IDs are getting changed.

firefox_ui_tests/functional/locationbar/test_go_button.py
```diff
@@ -0,0 +1,36 @@
...
+        go_button.click()
```
whimboo: You are not checking if the button is displayed before clicking on it. Doing that would give us a better vision what's failing in case of the target page doesn't load.

firefox_ui_tests/functional/locationbar/test_go_button.py
```diff
@@ -0,0 +1,36 @@
...
+
+        # landing page matches input, go button no longer displays
+        self.assertEqual(locationbar.value, 'mozilla.org/')
```
whimboo: This cannot be done with `assertEqual` because you don't know how long it takes to actually load the page. Check other [existing tests](https://github.com/mozilla/firefox-ui-tests/pull/133/files#diff-d9746bcb1da6dc49a500e72e4924fe61R34) in our repository how to use the `Wait(...).until(...)` method to ensure the target page is shown.

whimboo: Also please do not check against the hard-coded value but the URL as setup in setUp.

firefox_ui_tests/functional/locationbar/test_go_button.py
```diff
@@ -0,0 +1,36 @@
...
+        self.assertFalse(go_button.is_displayed())
+
+    def tearDown(self):
+        FirefoxTestCase.tearDown(self)
```
whimboo: If there is no special code in this method, it can be omitted. Python will automatically call the tearDown method from the parent class.

rbillings:
I got really behind at the end of the quarter. I'll address the comments when I return from PTO after 4/15.

firefox_puppeteer/ui/toolbars.py
```diff
@@ -12,179 +12,187 @@
 
 class NavBar(BaseLib):
     """Provides access to the DOM elements contained in the
-    navigation bar as well as the location bar."""
-
+        navigation bar as well as the location bar."""
+
```
whimboo: Not sure what's going on with all those doc changes. They introduced a broken indentation across the whole file. Please revert all of those and add the one you need for the go button with the same indentation.

firefox_puppeteer/ui/toolbars.py
```diff
@@ -12,179 +12,187 @@
...
-        self.keys = Keys(self.get_marionette)
-
-    @use_class_as_property('ui.toolbars.AutocompleteResults')
+        
+            def __init__(self, *args, **kwargs):
+                BaseLib.__init__(self, *args, **kwargs)
+                    # TODO: A "utility" module that sets up the client directly would be
+                    # useful here.
+                    self.l10n = L10n(self.get_marionette)
+                        self.keys = Keys(self.get_marionette)
```
whimboo: Same here. This is even code with broken indenation, which even Python should complain about. This definitely needs a fix.

firefox_ui_tests/functional/locationbar/test_go_button.py
```diff
@@ -0,0 +1,31 @@
...
+from firefox_ui_harness.testcase import FirefoxTestCase
+
+from marionette_driver import By, Wait
+
+
+class TestGoButton(FirefoxTestCase):
+
+    def setUp(self):
+        FirefoxTestCase.setUp(self)
+        self.url = self.marionette.absolute_url('layout/mozilla.html')
```
whimboo: Please name it `self.test_url`. That's the name we use across all of our tests.

firefox_ui_tests/functional/locationbar/test_go_button.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def test_gobutton(self):
+        # no input, go button should not display
+        go_button = self.browser.locationbar.go_button
+        self.assertEqual(go_button.element.get_attribute('state'), 'disabled')
```
whimboo: The state will not be clear because you don't have the initial page load anymore. Keep in mind that you don't know how the former test left the browser.

firefox_ui_tests/functional/locationbar/test_go_button.py
```diff
@@ -0,0 +1,31 @@
...
+
+        # add input, go button displays, click go button
+        url = self.marionette.absolute_url('layout/mozilla.html')
+        urlbar.send_keys(url)
```
whimboo: Use `self.test_url` here.

firefox_ui_tests/functional/locationbar/test_go_button.py
```diff
@@ -0,0 +1,31 @@
...
+        self.assertEqual(go_button.element.get_attribute('state'), 'active')
+        go_button.click()
+
+        # landing page matches input, go button no longer displays
+        target_url = self.browser.get_final_url(url)
+        Wait(self.marionette).until(lambda mn: mn.get_url() == target_url)
```
whimboo: We do not have a redirect for our local test page, so the call to `get_final__url()` is not necessary. Even it would introduce a bug which doesn't let us check anymore if the click action on the go button actually loaded the page. That is because the method itself loads the target url. Line 29 should be enough when compared to `self.test_url`.

firefox_ui_tests/functional/locationbar/test_go_button.py
```diff
@@ -0,0 +1,31 @@
...
+        self.assertEqual(self.marionette.get_url(), self.url)
+        self.assertFalse(go_button.is_displayed())
```
whimboo: Here you are using `is_displayed' while for all other instances you are getting the state attribute. We should be consistent. So please use the state attribute here.

firefox_ui_tests/functional/locationbar/test_go_button.py
```diff
@@ -0,0 +1,31 @@
...
+        # add input, go button displays, click go button
+        url = self.marionette.absolute_url('layout/mozilla.html')
+        urlbar.send_keys(url)
+        self.assertEqual(go_button.element.get_attribute('state'), 'active')
+        go_button.click()
+
+        # landing page matches input, go button no longer displays
+        target_url = self.browser.get_final_url(url)
+        Wait(self.marionette).until(lambda mn: mn.get_url() == target_url)
+        self.assertEqual(self.marionette.get_url(), self.url)
```
whimboo: With the above change this line won't be necessary anymore.

firefox_ui_tests/functional/locationbar/test_go_button.py
whimboo: Oh and given that Marionette cannot correctly check for a page load triggered externally yet, we should keep the element check for the organization or whatever is on that page.

rbillings:
I haven't been keeping up with this and don't have time to complete it. It should be a quick test for someone else to complete.

PR #77, opened 2015-02-09
Travis multi-os: test skipping flash with master config

galgeek:
flash install pulls in updates to dependencies, including one or two that produce PEP440 errors in multi-os tests


galgeek:
It looks like those packages have already been updated for Travis' Linux Python workers.

PR #83, opened 2015-02-13
Bug 1123758-Provide access to more members of Services.appinfo in firefox puppeteer libraries. r=whimboo

heisenberg-:


firefox_puppeteer/api/appinfo.py
```diff
@@ -6,17 +6,37 @@
...
+    def __getattr__(self, attr):
+        with self.marionette.using_context('chrome'):
+            value = self.marionette.execute_script("""
+              return Services.appinfo[arguments[0]];
+            """, script_args=[attr])
+
+            if value:
+                return value
+            else:
+                raise AttributeError("'"+self.__class__.__name__+"' has no attribute '"+attr+"'")
```
whimboo: nit: please put blanks between operators and operands, so it becomes `"'" + `.

firefox_ui_harness/decorators.py
```diff
@@ -10,8 +10,11 @@
 
 def skip_if_e10s(target):
     def wrapper(self, *args, **kwargs):
-        if self.appinfo.browserTabsRemoteAutostart:
-            raise SkipTest("Skipping due to e10s")
+        try:
+            if self.appinfo.browserTabsRemoteAutostart:
```
whimboo: If this fails then there is something wrong in retrieving this property via the appinfo class. There should be no need to do any changes here.

heisenberg-: Earlier JS script was returning null value only, but now for null values we are raising AttributeError.

whimboo: Can I ask what kind of Firefox build are you using for testing your patch? Is it the latest Nightly build? Getting this property out of the appinfo object should not throw an exception but return false.

heisenberg-: I ran it on travis ci. You can see the results above showing failures.

heisenberg-:
Nope, it is returning null.

firefox_puppeteer/api/appinfo.py
```diff
@@ -6,17 +6,38 @@
...
+    """
+
+    def __getattr__(self, attr):
+        with self.marionette.using_context('chrome'):
+            value = self.marionette.execute_script("""
+              return Services.appinfo[arguments[0]];
+            """, script_args=[attr])
+
+            if value is None:
+                return value
```
whimboo: Small mistake with big failing result. Please check this if condition and it will be clear why you see this failure. :)

firefox_puppeteer/api/appinfo.py
```diff
@@ -6,17 +6,38 @@
...
+            else:
+                raise AttributeError(
+                    "'" + self.__class__.__name__ + "' has no attribute '" + attr + "'")
```
whimboo: Please format it with the `format` method like `"{}" has no attribute "{}".format(self.__class__.__name__, attr)`.

firefox_puppeteer/api/appinfo.py
heisenberg-: Actually, i am backing down for a few days from here. I will learn git first. I was making changes here in the git text editor itself. 
And I don't have access to IRC. I work on an institutional network. So we need to work out something else.

firefox_puppeteer/api/appinfo.py
```diff
@@ -6,17 +6,38 @@
...
+        with self.marionette.using_context('chrome'):
+            value = self.marionette.execute_script("""
+              return Services.appinfo[arguments[0]];
+            """, script_args=[attr])
+
+            if value is not None:
+                return value
+            else:
+                raise AttributeError(
+                    "{} has no attribute {}".format(self.__class__.__name__, attr))
```
whimboo: It looks like that you were able to learn all the necessary bits in git quiet fast. That's good to see. The update of this PR looks pretty good!

There is only one nit here, where you make use of double quotes instead of single quotes. But let me fix that when I merge your patch.

heisenberg-:
Actually, i am modifying in the git text editor itself. Still have to learn git.
And I don't have access to IRC network, we have to figure out something else.

whimboo:
@heisenberg- I have sent you an email so we can continue discuss git problems there. Thanks for letting me know about those.

heisenberg-:
Actually i am reading a book about git. I think i will complete it 2-3 days. 
Most of the doubts got cleared when i learnt about 3 tree arch. in git.

whimboo:
The PR has been rebased against latest master and the commit message updated. I merged it as https://github.com/mozilla/firefox-ui-tests/commit/7b2ccef3632b21251c0ea0cee20ee17843ae2185.

PR #84, opened 2015-02-13
Bug 1132943 - Implement class for PageInfoWindow

whimboo:
This PR is not ready yet for review. It just visualizes the current state of the ui module.

### DIFF
--- a/firefox_puppeteer/api/l10n.py
+++ b/firefox_puppeteer/api/l10n.py
@@ -40,8 +40,8 @@ class L10n(BaseLib):
 
         with self.marionette.using_context('chrome'):
             value = self.marionette.execute_script("""
-                var parser = Cc["@mozilla.org/xmlextras/domparser;1"]
-                              .createInstance(Ci.nsIDOMParser);
+                var parser = Components.classes["@mozilla.org/xmlextras/domparser;1"]
+                             .createInstance(Components.interfaces.nsIDOMParser);
                 var doc = parser.parseFromString(arguments[0], "text/xml");
                 var node = doc.querySelector("elem[id='entity']");
 
--- a/firefox_puppeteer/base.py
+++ b/firefox_puppeteer/base.py
@@ -2,7 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
-from . import errors
+from marionette_driver.marionette import HTMLElement
 
 
 class BaseLib(object):
@@ -28,16 +28,29 @@ class BaseLib(object):
 class UIBaseLib(BaseLib):
     """A base class for all UI element wrapper classes inside a chrome window."""
 
-    def __init__(self, marionette_getter, window):
-        BaseLib.__init__(self, marionette_getter)
-
+    def __init__(self, marionette_getter, window, element):
         # importing globally doesn't work
         from .ui.windows import BaseWindow
-        if not isinstance(window, BaseWindow):
-            raise errors.UnexpectedWindowTypeError('Not a valid BaseWindow: "%s"' %
-                                                   window)
+
+        assert isinstance(window, BaseWindow)
+        assert isinstance(element, HTMLElement)
+
+        BaseLib.__init__(self, marionette_getter)
         self._window = window
+        self._element = element
+
+    @property
+    def element(self):
+        """Returns the reference to the underlying DOM element.
+
+        :returns: Reference to the DOM element
+        """
+        return self._element
 
     @property
     def window(self):
+        """Returns the reference to the chrome window.
+
+        :returns: :class:`BaseWindow` instance of the chrome window.
+        """
         return self._window
--- a/firefox_puppeteer/ui/pageinfo/deck.py
+++ b/firefox_puppeteer/ui/pageinfo/deck.py
@@ -1,0 +1,202 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from marionette_driver import By, Wait
+
+from ...base import UIBaseLib
+
+
+class Deck(UIBaseLib):
+
+    def _create_panel_for_id(self, panel_id):
+        """Creates an instance of :class:`Panel` for the specified panel id.
+
+        :param panel_id: The ID of the panel to create an instance of.
+
+        :returns: :class:`Panel` instance
+        """
+        mapping = {'feedPanel': FeedPanel,
+                   'generalPanel': GeneralPanel,
+                   'mediaPanel': MediaPanel,
+                   'permPanel': PermissionsPanel,
+                   'securityPanel': SecurityPanel
+                   }
+
+        panel = self.element.find_element(By.ID, panel_id)
+        return mapping.get(panel_id, Panel)(lambda: self.marionette, self.window, panel)
+
+    # Properties for visual elements of the deck #
+
+    @property
+    def feed(self):
+        """The :class:`FeedPanel` instance for the feed panel.
+
+        :returns: :class:`FeedPanel` instance.
+        """
+        return self._create_panel_for_id('feedPanel')
+
+    @property
+    def general(self):
+        """The :class:`GeneralPanel` instance for the general panel.
+
+        :returns: :class:`GeneralPanel` instance.
+        """
+        return self._create_panel_for_id('generalPanel')
+
+    @property
+    def media(self):
+        """The :class:`MediaPanel` instance for the media panel.
+
+        :returns: :class:`MediaPanel` instance.
+        """
+        return self._create_panel_for_id('mediaPanel')
+
+    @property
+    def panels(self):
+        """List of all the :class:`Panel` instances of the current deck.
+
+        :returns: List of :class:`Panel` instances.
+        """
+        panels = self.marionette.execute_script("""
+          let deck = arguments[0];
+          let panels = [];
+
+          for (let index = 0; index < deck.children.length; index++) {
+            panels.push(deck.children[index].id);
+          }
+
+          return panels;
+        """, script_args=[self.element])
+
+        return [self._create_panel_for_id(panel) for panel in panels]
+
+    @property
+    def permissions(self):
+        """The :class:`PermissionsPanel` instance for the permissions panel.
+
+        :returns: :class:`PermissionsPanel` instance.
+        """
+        return self._create_panel_for_id('permPanel')
+
+    @property
+    def security(self):
+        """The :class:`SecurityPanel` instance for the security panel.
+
+        :returns: :class:`SecurityPanel` instance.
+        """
+        return self._create_panel_for_id('securityPanel')
+
+    # Properties for helpers when working with the deck #
+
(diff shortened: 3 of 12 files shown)


whimboo:
I pushed some more updates to this PR which show how the deck could be used. It's still not clear right now how to best separate out the tabs which are used for selecting the appropriate panel. Maybe we can bind it to the appropriate category class for those cases when we need its label etc?

Otherwise I had to do a couple of changes for the UIBaseLib so it accepts a parent element and not only a window. This helps in better navigating in the hierarchy. If all that works fine, we can file a bug later to update all the other UI classes.

firefox_puppeteer/api/l10n.py
```diff
@@ -40,8 +40,8 @@ def get_entity(self, dtd_urls, entity_id):
 
         with self.marionette.using_context('chrome'):
             value = self.marionette.execute_script("""
-                var parser = Cc["@mozilla.org/xmlextras/domparser;1"]
-                              .createInstance(Ci.nsIDOMParser);
+                var parser = Components.classes["@mozilla.org/xmlextras/domparser;1"]
+                              .createInstance(Components.interfaces.nsIDOMParser);
```
whimboo: This has to be changed because `Cc` and `Ci` is not available for window types other then browser.

firefox_puppeteer/base.py
```diff
@@ -28,15 +28,31 @@ def get_marionette(self):
...
+
+        self._element = element
+
+    @property
+    def element(self):
+        return self._element
+
+    @property
+    def parent(self):
+        return self._parent
```
whimboo: If this is all fine we will have a follow-up bug to convert the remaining UI classes to UIBaseLib.

firefox_puppeteer/ui/pageinfo/deck.py
```diff
@@ -0,0 +1,208 @@
...
+
+    @property
+    def panels(self):
+        """List of all the :class:`Panel` instances of the current deck.
+
+        :returns: List of :class:`Panel` instances.
+        """
+        panels = self.element.find_elements(By.TAG_NAME, 'vbox')
+        return [self._create_panel_for_id(panel.get_attribute('id')) for panel in panels
+                if panel.get_attribute('id').endswith('Panel')]
```
chmanchester: Maybe we should have a list of panels we're looking for than just looping through and checking ids -- does anything guarantee a vbox with an id ending in "Panel" is what we mean by panel?

firefox_puppeteer/base.py
```diff
@@ -28,16 +28,44 @@ def get_marionette(self):
 class UIBaseLib(BaseLib):
     """A base class for all UI element wrapper classes inside a chrome window."""
 
-    def __init__(self, marionette_getter, window):
-        BaseLib.__init__(self, marionette_getter)
-
+    def __init__(self, marionette_getter, parent, element):
```
chmanchester: I'm a little skeptical of adding "parent" to all our ui classes. Some elements don't have a parent, but moreover I think this opens us up to confusion by what's specified here and what's the actual parent element in the dom (which could probably get out of sync in some cases).

firefox_puppeteer/ui/pageinfo/deck.py
```diff
@@ -0,0 +1,208 @@
...
+        :returns: True, if the panel is selected.
+        """
+        return self.parent.selected_panel == self
+
+    def select(self):
+        """Selects this panel.
+
+        :returns: :class:`Panel` instance to the this panel.
+        """
+        return self.parent.select(self)
```
chmanchester: These two methods are the only places "parent" is used meaningfully. As long as we're likely always have a reference to our "deck" we can just use its methods.  I tend to think any place a hierarchy needs to be inverted in code is a code smell.

firefox_puppeteer/ui/windows.py
```diff
@@ -99,6 +99,9 @@ def create_window_instance(self, handle, expected_class=None):
 
         if window_type == 'navigator:browser':
             window = BrowserWindow(lambda: self.marionette, handle)
+        elif window_type == 'Browser:page-info':
+            from .pageinfo.window import PageInfoWindow
```
chmanchester: Are the local imports by choice or working around some other bug?

firefox_puppeteer/ui/pageinfo/deck.py
chmanchester: Looking at pageinfo.xul, this is just the children of mainDeck.

whimboo: As implemented all panels will be a `vbox`. So what would you suggest? Using a selector for the child elements of mainDeck?

firefox_puppeteer/base.py
whimboo: Every element in the DOM has a parent, except for the top level root node. That's why the check for the `BaseWindow` instance. I never thought that we copy the DOM 1:1, and I don't think that this is our intention. We are building up a tree of UI elements, which are important for our testing. The DOM is not relevant here IMHO.

firefox_puppeteer/ui/pageinfo/deck.py
whimboo: There will be coming up some other classes for other window types, which will also have to use parent. IMO it's way clearer to have a streamlined class design as lots of exceptions in various classes you would have to obey.

firefox_puppeteer/ui/windows.py
whimboo: This is still the same issue as we talked about on another PR a couple of weeks ago. But I will check it again later.

firefox_puppeteer/ui/pageinfo/deck.py
chmanchester: Yes.

firefox_puppeteer/base.py
chmanchester: I think this is confusing if in some cases our "parent" is the parent in the DOM and in some cases it isn't.

firefox_puppeteer/ui/pageinfo/deck.py
chmanchester: I don't see how accessing an element by its parent rather than storing a reference to its parent is going to invite exceptions. As long as we can keep our references to the hierarchy going in one direction things will be simpler so I think we should.

firefox_puppeteer/ui/windows.py
chmanchester: This should be on file so someone can look into it.

whimboo: Tested and it's the same problem with maybe the cyclic import between the pageinfo window module and windows.py. We would have to refactor those window classes later, so we might be able to get rid of those local imports.

firefox_puppeteer/ui/pageinfo/deck.py
whimboo: I don't see how to use By.SELECTOR to get back child elements for a parent node. I always would have to specify the type of DOM node.

Coming back to the other idea with a hard-coded panel list, it would block us from being able to allow others from working with those dynamically added panels, e.g. via an addon. I wanted to make sure that we keep this option.

firefox_puppeteer/base.py
whimboo: So what now? I don't see a way that we could or even should duplicate the DOM regarding the parent element. We have our own class structure the test creators are working with. They do not work with the DOM.

firefox_puppeteer/ui/pageinfo/deck.py
whimboo: This will never work for e.g. window classes. Keep in mind that each instance of whatever type has to know about the window it lives in, and e.g. how to find properties, dtd entities. Also I do not understand what you mean with "accessing an element by its parent"? If you are talking about the DOM, its parent will not match our parent, and even doesn't necessarily mean it is the parent we want. So please explain what you have in mind.

firefox_puppeteer/ui/windows.py
whimboo: We can file a but once this PR got merged.

firefox_puppeteer/ui/pageinfo/deck.py
chmanchester: I just mean calling deck.select() or deck.selected_panel rather than self.select() or self.selected_panel

firefox_puppeteer/ui/pageinfo/deck.py
chmanchester: I would get the list of children from JS.

firefox_puppeteer/base.py
chmanchester: My proposal is not to add "parent" to our classes at all, at least not at this time.

firefox_puppeteer/ui/pageinfo/deck.py
whimboo: Sorry, I still don't see what you are talking about here. This method does not call `self.select`, but calls `select()` for the parent element, which in this case is the deck.

firefox_puppeteer/ui/pageinfo/deck.py
whimboo: We have a ui testing framework and have to limit ourselves as much as possible from retrieving elements via JS objects. Those pieces are tested en-mass via browser-chrome etc, so I also see only lots of overlap.

firefox_puppeteer/ui/pageinfo/deck.py
chmanchester: I mean the equivalent functionality will be achieved by removing every call to <panel>.select and <panel>.selected and replacing each with a call to deck.select(<panel>) or comparing <panel> == deck.selected_panel. By "parent" above I mean the deck.

firefox_puppeteer/ui/pageinfo/deck.py
chmanchester: In the future some vbox with an id ending in "Panel" could be added to an unrelated part of the dom that will erroneously get added to this list. This is just a limitation of marionette's selectors that we can work around with a small amount of JavaScript so we don't have to use imprecise heuristics to retrieve elements.

firefox_puppeteer/ui/pageinfo/deck.py
whimboo: We clarified that in a Vidyo call, and decided to rename `parent` to `window`. It will be fixed in one of the next commits I push.

firefox_puppeteer/ui/pageinfo/deck.py
whimboo: We decided to use the JS object here to get the panels via `deck.children[x]`. A fix will be part of the next commit.

firefox_puppeteer/ui/pageinfo/deck.py
```diff
@@ -0,0 +1,202 @@
...
+          let panels = [];
+
+          for (let index = 0; index < deck.children.length; index++) {
+            panels.push(deck.children[index]);
+          }
+
+          return panels;
+        """, script_args=[self.element])
+
+        return [self._create_panel_for_id(panel.get_attribute('id')) for panel in panels]
```
chmanchester: You can just get the ids from JS.

whimboo: Ups, that's totally true. No need to return full DOM nodes when we can simply return the id with the `execute_script` call. I will update it.

PR #85, opened 2015-02-17
Bug 1132940 - Convert security module from Mozmill tests to Marionette. r=chmanchester

whimboo:


firefox_puppeteer/ui/tabbar.py
```diff
@@ -236,6 +237,40 @@ def tab_element(self):
...
+        """
+        self.switch_to()
+
+        cert = self.marionette.execute_script("""
+          var securityUI = arguments[0].linkedBrowser.securityUI;
+          var status = securityUI.QueryInterface(Ci.nsISSLStatusProvider)
+                                 .SSLStatus;
+
+          return status ? status.serverCert : null;
+        """, script_args=[self.tab_element])
```
chmanchester: Can't you do this in the security module by using gBrowser.selectedBrowser ?

firefox_puppeteer/api/security.py
```diff
@@ -0,0 +1,31 @@
...
+
+    # Security related helpers #
+
+    def get_address_from_certificate(self, certificate):
+        """Retrieves the address of the organization from the certificate information.
+
+        The returned address may be `None` in case of no address found, or a dictionary
+        with the following entries: `city`, `country`, `postal_code`, `state`, `street`.
+
+        :param tab: A JSON blob representing the certificate, which can usually be
```
chmanchester: :param certificate:

firefox_puppeteer/ui/tabbar.py
chmanchester: If that's doesn't seem like a good idea, what about passing a tab element as a parameter to security module methods?

whimboo: >> +        cert = self.marionette.execute_script("""
>> +          var securityUI = arguments[0].linkedBrowser.securityUI;
>> +          var status = securityUI.QueryInterface(Ci.nsISSLStatusProvider)
>> +                                 .SSLStatus;
>> +
>> +          return status ? status.serverCert : null;
>> +        """, script_args=[self.tab_element])
> 
> Can't you do this in the security module by using gBrowser.selectedBrowser ?

We could but I don't want to force that the tab has to be selected
first. It should perfectly be doable to get the certificate for a loaded
page in a background tab.

chmanchester: The I'd say we should pass in the tab element.

whimboo: This might actually be a good idea. Lemme do that instead. It would indeed be cleaner to have this in the security module.

whimboo: We will loose the location information, but maybe we can also get this via the tab element.

firefox_puppeteer/api/security.py
```diff
@@ -0,0 +1,54 @@
...
+
+    # Security related helpers #
+
+    def get_address_from_certificate(self, certificate):
+        """Retrieves the address of the organization from the certificate information.
+
+        The returned address may be `None` in case of no address found, or a dictionary
+        with the following entries: `city`, `country`, `postal_code`, `state`, `street`.
+
+        :param certificate: A JSON blob representing the certificate, which can usually be
```
chmanchester: s/blob/object/ ?

firefox_puppeteer/api/security.py
```diff
@@ -0,0 +1,54 @@
...
+                           ',postalCode=(?P<postal_code>.+?),STREET=(?P<street>.+?)'
+                           ',serial')
+        results = regex.search(certificate['subjectName'])
+
+        return results.groupdict() if results else results
+
+    def get_certificate_for_page(self, tab_element):
+        """The SSL certificate assiciated with the loaded web page in the given tab.
+
+        :returns: Certificate details as JSON blob.
```
chmanchester: Document the param

whimboo: Ups. Thanks for the note. Simply missed it.

whimboo:
PR has been merged as https://github.com/mozilla/firefox-ui-tests/commit/ad8b37f657cd68b71ed46b4b51bf3f1b7b5531ce

PR #86, opened 2015-02-17
Bug 1132936 - Create unit tests for IdentityPopup class in toolbars.py. r=whimboo

heisenberg--:


firefox_puppeteer/tests/test_toolbars.py
```diff
@@ -135,3 +137,71 @@ def test_matching_text(self):
             self.assertTrue(len(all_matches) > 0)
             for match_fragment in all_matches:
                 self.assertIn(match_fragment, (input_text, input_text.upper()))
+
+
+class TestIdentityPopup(FirefoxTestCase):
+
+    def test_box(self):
+        self.assertIsNotNone(
+            self.browser.navbar.locationbar.identity_popup.box)
```
whimboo: For all those tests please check not for notNone but for the exact type of element like: 'self.assertEqual('textbox', urlbar.get_attribute('localName'))'. Also you can put all in a single line, we have a limit of 100 chars per line in this repository. The same applies to all the other test methods in this file.

whimboo: And maybe you want to use a `setUp` method where you can set `self.identity_popup = self.browser.navbar.locationbar.identity_popup`, and to use it as shortcut in your tests.

firefox_puppeteer/tests/test_toolbars.py
```diff
@@ -135,3 +137,71 @@ def test_matching_text(self):
...
+
+    def test_host(self):
+        self.assertIsNotNone(
+            self.browser.navbar.locationbar.identity_popup.host)
+
+    def test_is_open(self):
+        self.assertEquals(
+            self.browser.navbar.locationbar.identity_popup.is_open, False)
+
+        url = 'https://ssl-ev.mozqa.com'
```
whimboo: Please use a test page loaded from the local web server or a data uri, which AFAIR is also used in this test file.

whimboo: Reason for that is that our unit tests should touch as less as possible the external network.

whimboo:
@heisenberg-- do you have an update for us? Will you be able to finish this PR off?

heisenberg--:
@whimboo Sorry, but I can't contribute for a while. I am caught into something, and can't spend even an hour.

whimboo:
I see. So thanks for your feedback. @galgeek will pick it up then. I'm closing this PR now.

PR #88, opened 2015-02-17
Bug 1132386: Observe bookmarks-restore-failed in restoring the bookmarks

gbrmachado:


firefox_puppeteer/api/places.py
```diff
@@ -88,14 +88,22 @@ def restore_default_bookmarks(self):
 
               let observer = {
                 observe: function (aSubject, aTopic, aData) {
-                  Services.obs.removeObserver(observer, "bookmarks-restore-success");
+                  if ( aTopic === "bookmarks-restore-success") {
+                    Services.obs.removeObserver(observer, "bookmarks-restore-success");
```
whimboo: With this change we will not remove all registered observer topics. So what you wanna do is to call `removeObserver` for both registered topics at the beginning of this method.

firefox_puppeteer/api/places.py
```diff
@@ -88,14 +88,22 @@ def restore_default_bookmarks(self):
...
 
-                  marionetteScriptFinished(true);
+                    marionetteScriptFinished(true);
+                  }
+                  if ( aTopic == "bookmarks-restore-failed") {
+                    Services.obs.removeObserver(observer, "bookmars-restore-failed");
+
+                    marionetteScriptFinished(false);
```
whimboo: You can get rid of the above two if blocks if you do a simple compare of the topic with `bookmarks-restore-success` and take this as argument for marionetteScriptFinished.

gbrmachado:
Hi. I tried to resync this branch with the master one(in order to get all the news commits). However, it seems that it messed all this branch up :P 
Do you think that it would be better to take this branch out to create another pull request?(with a different branch)

whimboo:
What you usually should do is to first switch to your local master branch and pull the latest changes from the remote master. Once this is done, you run a `git rebase master` from your integration branch. Have you done those steps? Did you get any merge conflict?

whimboo:
@gbrmachado any reason that you closed this PR? I was hoping that we can solve the problems you were seeing here.

gbrmachado:
Sorry, I think that would be easier to create another branch. But now I realised that would be better to use this situation in order to learn how to use the git. :(

whimboo:
Right, but don't worry. I assume we will hit a similar situation in the near future. We are all still working too close on code that merge conflicts can happen all the time.

PR #90, opened 2015-02-18
Bug 1133752 - Convert Mozmill software-update.js library to Python/Marionette

bobsilverberg:
@whimboo This is not yet complete, and I'm not sure how good it is, as I was mostly hacking my way through it, but everything (except `active_update`) seems to be working.

I'm sure both the code and tests need quite a bit of work, but it's a start. I'd appreciate it if you could take a look at it when you have time and provide any feedback you may have.

### DIFF
--- a/firefox_puppeteer/api/software_update.py
+++ b/firefox_puppeteer/api/software_update.py
@@ -1,0 +1,496 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import os
+import re
+
+import mozinfo
+
+from ..base import BaseLib
+from .appinfo import AppInfo
+from .prefs import Preferences
+
+
+class ActiveUpdate(BaseLib):
+
+    def __getattr__(self, attr):
+        value = self.marionette.execute_script("""
+          let ums = Components.classes['@mozilla.org/updates/update-manager;1']
+                    .getService(Components.interfaces.nsIUpdateManager);
+          return ums.activeUpdate[arguments[0]];
+        """, script_args=[attr])
+
+        if value:
+            return value
+        else:
+            raise AttributeError('{} has no attribute {}'.format(self.__class__.__name__,
+                                                                 attr))
+
+    @property
+    def exists(self):
+        """Checks if there is an active update.
+
+        :returns: True if there is an active update
+        """
+        active_update = self.marionette.execute_script("""
+          let ums = Components.classes['@mozilla.org/updates/update-manager;1']
+                    .getService(Components.interfaces.nsIUpdateManager);
+          return ums.activeUpdate;
+        """)
+
+        return bool(active_update)
+
+    def get_patch_at(self, patch_index):
+        """Use nsIUpdate.getPatchAt to return a patch from an update.
+
+        :returns: JSON data for an nsIUpdatePatch object
+        """
+        return self.marionette.execute_script("""
+          let ums = Components.classes['@mozilla.org/updates/update-manager;1']
+                    .getService(Components.interfaces.nsIUpdateManager);
+          return ums.activeUpdate.getPatchAt(arguments[0]);
+        """, script_args=[patch_index])
+
+    @property
+    def patch_count(self):
+        """Get the patchCount from the active update.
+
+        :returns: The patch count
+        """
+        return self.marionette.execute_script("""
+          let ums = Components.classes['@mozilla.org/updates/update-manager;1']
+                    .getService(Components.interfaces.nsIUpdateManager);
+          return ums.activeUpdate.patchCount;
+        """)
+
+    @property
+    def selected_patch(self):
+        """Get the selected patch for the active update.
+
+        :returns: JSON data for the selected patch
+        """
+        return self.marionette.execute_script("""
+          let ums = Components.classes['@mozilla.org/updates/update-manager;1']
+                    .getService(Components.interfaces.nsIUpdateManager);
+          return ums.activeUpdate.selectedPatch;
+    """)
+
+
+class MARChannels(BaseLib):
+    """Class to handle the allowed MAR channels as listed in update-settings.ini."""
+    INI_SECTION = 'Settings'
+    INI_OPTION = 'ACCEPTED_MAR_CHANNEL_IDS'
+
+    def __init__(self, marionette_getter):
+        BaseLib.__init__(self, marionette_getter)
+
+        self._ini_file_path = self.marionette.execute_script("""
+          Components.utils.import('resource://gre/modules/Services.jsm');
+
+          let file = Services.dirsvc.get('GreD', Components.interfaces.nsIFile);
+          file.append('update-settings.ini');
+
+          return file.path;
+        """)
+
+    @property
+    def config_file_path(self):
+        """The path to the update-settings.ini file."""
+        return self._ini_file_path
+
+    @property
+    def config_file_contents(self):
+        """The contents of the update-settings.ini file."""
+        with open(self.config_file_path) as f:
+            return f.read()
+
+    @property
+    def channels(self):
+        """The channels as found in the ACCEPTED_MAR_CHANNEL_IDS option
+        of the update-settings.ini file.
+
+        :returns: A set of channel names
+        """
+        channels = self.marionette.execute_script("""
+          Components.utils.import("resource://gre/modules/FileUtils.jsm");
+          let iniFactory = Components.classes['@mozilla.org/xpcom/ini-processor-factory;1']
+                           .getService(Components.interfaces.nsIINIParserFactory);
+
+          let file = new FileUtils.File(arguments[0]);
+          let parser = iniFactory.createINIParser(file);
+
+          return parser.getString(arguments[1], arguments[2]);
+        """, script_args=[self.config_file_path, self.INI_SECTION, self.INI_OPTION])
+        return set(channels.split(','))
+
+    @channels.setter
+    def channels(self, channels):
+        """Set the channels in the update-settings.ini file.
+
+        :param channels: A set of channel names
+        """
+        new_channels = ','.join(channels)
+        self.marionette.execute_script("""
+          Components.utils.import("resource://gre/modules/FileUtils.jsm");
+          let iniFactory = Components.classes['@mozilla.org/xpcom/ini-processor-factory;1']
+                           .getService(Components.interfaces.nsIINIParserFactory);
+
+          let file = new FileUtils.File(arguments[0]);
+
+          let writer = iniFactory.createINIParser(file)
+                       .QueryInterface(Components.interfaces.nsIINIParserWriter);
+
+          writer.setString(arguments[1], arguments[2], arguments[3]);
+          writer.writeFile(null, Components.interfaces.nsIINIParserWriter.WRITE_UTF16);
+        """, script_args=[self.config_file_path, self.INI_SECTION, self.INI_OPTION, new_channels])
+
+    def add_channels(self, channels):
+        """Add channels to the update-settings.ini file.
+
+        :param channels: A set of channel names to add
+        """
+        self.channels = self.channels | channels
+
+    def remove_channels(self, channels):
+        """Remove channels from the update-settings.ini file.
+
+        :param channels: A set of channel names to remove
+        """
+        self.channels = self.channels - channels
+
+
+class SoftwareUpdate(BaseLib):
+    """The SoftwareUpdate API adds support for an easy access to the update process."""
+    PREF_APP_DISTRIBUTION = 'distribution.id'
+    PREF_APP_DISTRIBUTION_VERSION = 'distribution.version'
+    PREF_APP_UPDATE_URL = 'app.update.url'
+    PREF_DISABLED_ADDONS = 'extensions.disabledAddons'
+
+    def __init__(self, marionette_getter):
+        BaseLib.__init__(self, marionette_getter)
+
+        self.app_info = AppInfo(marionette_getter)
+        self.prefs = Preferences(marionette_getter)
+
+        self._update_channel = UpdateChannel(marionette_getter)
+        self._mar_channels = MARChannels(marionette_getter)
+        self._active_update = ActiveUpdate(marionette_getter)
+
+    @property
+    def ABI(self):
+        """Get the customized ABI for the update service.
+
+        :returns: ABI version
+        """
+        abi = self.app_info.XPCOMABI
+        if mozinfo.isMac:
+            abi += self.marionette.execute_script("""
+              let macutils = Components.classes['@mozilla.org/xpcom/mac-utils;1']
+                             .getService(Components.interfaces.nsIMacUtils);
+              if (macutils.isUniversalBinary) {
+                return '-u-' + macutils.architecturesInBinary;
+              }
+              return '';
+            """)
+
+        return abi
+
+    @property
+    def active_update(self):
+        """ Holds a reference to an :class:`ActiveUpdate` object."""
+        return self._active_update
+
+    @property
+    def allowed(self):
+        """Check if the user has permissions to run the software update
+
+        :returns: Status if the user has the permissions
+        """
+        return self.marionette.execute_script("""
+          let aus = Components.classes['@mozilla.org/updates/update-service;1']
+                    .getService(Components.interfaces.nsIApplicationUpdateService);
+          return aus.canCheckForUpdates && aus.canApplyUpdates;
+        """)
+
+    @property
+    def build_info(self):
+        """Return information of the current build version
+
+        :returns: A dictionary of build information
+        """
+        return {
+            'buildid': self.app_info.appBuildID,
+            'channel': self.update_channel.channel,
+            'disabled_addons': self.prefs.get_pref(self.PREF_DISABLED_ADDONS),
+            'locale': self.app_info.locale,
+            'mar_channels': self.mar_channels.channels,
+            'url_aus': self.get_update_url(True),
+            'user_agent': self.app_info.user_agent,
+            'version': self.app_info.version
+        }
+
+    @property
+    def is_complete_update(self):
+        """Return true if the offered update is a complete update
+
+        :returns: True if the offered update is a complete update
+        """
+        # Throw when isCompleteUpdate is called without an update. This should
+        # never happen except if the test is incorrectly written.
+        assert self.active_update.exists, 'An active update has been found'
+
+        patch_count = self.active_update.patch_count
+        assert patch_count == 1 or patch_count == 2,\
+            'An update must have one or two patches included'
+
+        # Ensure Partial and Complete patches produced have unique urls
+        if patch_count == 2:
+            patch0_URL = self.active_update.get_patch_at(0)['URL']
+            patch1_URL = self.active_update.get_patch_at(1)['URL']
+            assert patch0_URL != patch1_URL,\
+                'Partial and Complete download URLs are different'
+
+        return self.active_update.selected_patch['type'] == 'complete'
+
+    @property
+    def mar_channels(self):
+        """ Holds a reference to a :class:`MARChannels` object."""
+        return self._mar_channels
+
+    @property
+    def os_version(self):
+        """Returns information about the OS version
+
+        :returns: The OS version
+        """
+        return self.marionette.execute_script("""
+          Components.utils.import("resource://gre/modules/Services.jsm");
+
+          let osVersion;
+          try {
+            osVersion = Services.sysinfo.getProperty("name") + " " +
+                        Services.sysinfo.getProperty("version");
+          }
+          catch (ex) {
+          }
+
+          if (osVersion) {
+            try {
+              osVersion += " (" + Services.sysinfo.getProperty("secondaryLibrary") + ")";
+            }
+            catch (e) {
+              // Not all platforms have a secondary widget library,
+              // so an error is nothing to worry about.
+            }
+            osVersion = encodeURIComponent(osVersion);
+          }
+          return osVersion;
+        """)
+
+    @property
+    def patch_info(self):
+        """ Returns information of the active update in the queue."""
+        info = {'channel': self.update_channel.channel}
+
+        if (self.active_update.exists):
+            info['buildid'] = self.active_update.buildID
+            info['is_complete'] = self.is_complete_update
+            info['size'] = self.active_update.selected_patch['size']
+            info['type'] = self.update_type
+            info['url_mirror'] = \
+                self.active_update.selected_patch['finalURL'] or 'n/a'
+            info['version'] = self.active_update.appVersion
+
+        return info
+
+    @property
+    def staging_directory(self):
+        """ Returns the path to the updates staging directory."""
+        return self.marionette.execute_script("""
+          let aus = Components.classes['@mozilla.org/updates/update-service;1']
+                    .getService(Components.interfaces.nsIApplicationUpdateService);
+          return aus.getUpdatesDirectory().path;
+        """)
+
+    @property
+    def update_channel(self):
+        """ Holds a reference to an :class:`UpdateChannel` object."""
+        return self._update_channel
+
+    @property
+    def update_type(self):
+        """Returns the type of the active update."""
+        return self.active_update.type
+
+    def assert_update_applied(self, update_data):
+        """Checks if an update has been applied correctly.
+
+        :param update_data: All the data collected during the update process
+        """
+        # Get the information from the last update
+        info = update_data['updates'][update_data['update']['index']]
+
+        # The upgraded version should be identical with the version given by
+        # the update and we shouldn't have run a downgrade
+        check = self.marionette.execute_script("""
+          Components.utils.import("resource://gre/modules/Services.jsm");
+
+          return  Services.vc.compare(arguments[0], arguments[1]);
+        """, script_args=[info['build_post']['version'], info['build_pre']['version']])
+
+        assert check >= 0, 'The version of the upgraded build is higher or equal'
+
+        # If there was an update, the post build id should be identical with the patch
+        if info['patch']['buildid']:
+            assert info['build_post']['buildid'] == info['patch']['buildid'], \
+                'The post buildid is equal to the buildid of the update'
+
+        # If a target build id has been given, check if it matches the updated build
+        info['target_buildid'] = update_data['update'].get('target_buildid')
+        if info['target_buildid']:
+            assert info['build_post']['buildid'] == info['target_buildid'], \
+                'Post buildid matches target buildid of the update patch'
+
+        # An upgrade should not change the builds locale
+        assert info['build_post']['locale'] == info['build_pre']['locale'],\
+            'The locale of the updated build has not been changed'
+
+        # Check that no application-wide add-ons have been disabled
+        assert info['build_post']['disabled_addons'] == info['build_pre']['disabled_addons'],\
+            'No application-wide add-ons have been disabled by the update'
+
+    def force_fallback(self):
+        """Update the update.status file and set the status to 'failed:6'"""
+        with open(os.path.join(self.staging_directory, 'update.status'), 'w') as f:
+            f.write('failed: 6\n')
+
+    def init_update_tests(self, fallback=False):
+        """Initialize all the data used by update tests.
+
+        :param fallback: It is a fallback update
+        """
+        # Prepare persisted object for update results
+        # If an update fails the post build will be the same as the pre build.
+        persisted = {
+            'update': {
+                'index': 0,
+                'staging_path': self.staging_directory},
+            # Create results object with information of the unmodified pre build
+            # TODO: Lets change to persisted.update.results once the issue is fixed:
+            # https://github.com/mozilla/mozmill-automation/issues/175
+            'updates': [
+                {
+                    'build_pre': self.build_info,
+                    'build_post': self.build_info,
+                    'fallback': fallback,
+                    'patch': {},
+                    'success': False}]}
+
+        # If requested modify the default update channel. It will be active
+        # after the next restart of the application
+        if persisted['update']['channel']:
+            # Backup the original content and the path of the channel-prefs.js file
+            persisted['update']['default_update_channel'] = {
+                'content': self.update_channel.file_contents,
+                'path': self.update_channel.file_path}
+            self.update_channel.default_channel = persisted['update']['channel']
+
+        # If requested modify the list of allowed MAR channels
+        if persisted['update']['allowed_mar_channels']:
+            # Backup the original content and the path of the update-settings.ini file
+            persisted['update']['default_mar_channels'] = {
+                'content': self.mar_channels.config_file_contents,
+                'path': self.mar_channels.config_file_path}
+
+    def get_update_url(self, force=False):
+        """Retrieve the AUS update URL the update snippet is retrieved from
+
+        :param force: Boolean flag to force an update check
+
+        :returns: The URL of the update snippet
+        """
+        url = self.prefs.get_pref(self.PREF_APP_UPDATE_URL, '')
+
+        if not url:
+            return None
+
+        # get the next two prefs from the default branch
+        dist = self.prefs.get_pref(self.PREF_APP_DISTRIBUTION, True) or 'default'
+        dist_version = self.prefs.get_pref(self.PREF_APP_DISTRIBUTION_VERSION, True) or 'default'
+
+        # Not all placeholders are getting replaced correctly by formatURL
+        url = url.replace('%PRODUCT%', self.app_info.name)
+        url = url.replace('%BUILD_ID%', self.app_info.appBuildID)
+        url = url.replace('%BUILD_TARGET%', self.app_info.OS + '_' + self.ABI)
+        url = url.replace('%OS_VERSION%', self.os_version)
+        url = url.replace('%CHANNEL%', self.update_channel.channel)
+        url = url.replace('%DISTRIBUTION%', dist)
+        url = url.replace('%DISTRIBUTION_VERSION%', dist_version)
+
+        url = self.marionette.execute_script("""
+          Components.utils.import("resource://gre/modules/Services.jsm");
+          return Services.urlFormatter.formatURL(arguments[0]);
+        """, script_args=[url])
+
+        if force:
+            if '?' in url:
+                url += '&'
+            else:
+                url += '?'
+            url += 'force=1'
+        return url
+
+
+class UpdateChannel(BaseLib):
+    """Class to handle the update channel as listed in channel-prefs.js"""
+    REGEX_UPDATE_CHANNEL = re.compile(r'("app\.update\.channel", ")([^"].*)(?=")')
+
+    def __init__(self, marionette_getter):
+        BaseLib.__init__(self, marionette_getter)
+
+        self.prefs = Preferences(marionette_getter)
+
+        self.file_path = self.marionette.execute_script("""
+          Components.utils.import('resource://gre/modules/Services.jsm');
+
+          let file = Services.dirsvc.get('PrfDef', Components.interfaces.nsIFile);
+          file.append('channel-prefs.js');
+
+          return file.path;
+        """)
+
+    @property
+    def file_contents(self):
+        """The contents of the channel-prefs.js file."""
+        with open(self.file_path) as f:
+            return f.read()
+
+    @property
+    def channel(self):
+        """The name of the update channel as stored in the
+        app.update.channel pref."""
+        return self.prefs.get_pref('app.update.channel', True)
+
+    @property
+    def default_channel(self):
+        """Get the default update channel
+
+        :returns: Current default update channel
+        """
+        matches = re.search(self.REGEX_UPDATE_CHANNEL, self.file_contents).groups()
+        assert len(matches) == 2, 'Update channel value has been found'
+
+        return matches[1]
+
+    @default_channel.setter
+    def default_channel(self, channel):
+        """Set default update channel.
+
+        :param channel: New default update channel
+        """
+        assert channel, 'Update channel has been specified'
+        new_content = re.sub(
+            self.REGEX_UPDATE_CHANNEL, r'\1' + channel, self.file_contents)
+        with open(self.file_path, 'w') as f:
+            f.write(new_content)
--- a/firefox_puppeteer/docs/api/software_update.rst
+++ b/firefox_puppeteer/docs/api/software_update.rst
@@ -1,0 +1,27 @@
+.. py:currentmodule:: firefox_puppeteer.api.software_update
+
+SoftwareUpdate
+==============
+
+The SoftwareUpdate class provides helpers for update tests.
+
+.. autoclass:: SoftwareUpdate
+   :members:
+
+ActiveUpdate
+------------
+
+.. autoclass:: ActiveUpdate
+   :members:
+
+MARChannels
+-----------
+
+.. autoclass:: MARChannels
+   :members:
+
+UpdateChannel
+-------------
+
+.. autoclass:: UpdateChannel
+   :members:
--- a/firefox_puppeteer/tests/test_software_update.py
+++ b/firefox_puppeteer/tests/test_software_update.py
@@ -1,0 +1,122 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import os
+
+from firefox_ui_harness.testcase import FirefoxTestCase
+from firefox_puppeteer.api.software_update import SoftwareUpdate
+
+
+class TestSoftwareUpdate(FirefoxTestCase):
+
+    def setUp(self):
+        FirefoxTestCase.setUp(self)
+        self.software_update = SoftwareUpdate(self.get_marionette)
+        self.saved_mar_channels = self.software_update.mar_channels.channels
+        self.software_update.mar_channels.channels = set(['expected', 'channels'])
+
+    def tearDown(self):
+        try:
+            self.software_update.mar_channels.channels = self.saved_mar_channels
+        finally:
+            FirefoxTestCase.tearDown(self)
+
+    def test_abi(self):
+        self.assertTrue(self.software_update.ABI)
+
+    def test_allowed(self):
+        self.assertTrue(self.software_update.allowed)
+
+    def test_build_info(self):
+        build_info = self.software_update.build_info
+        self.assertEqual(build_info['disabled_addons'], None)
+        self.assertIn('Mozilla/', build_info['user_agent'])
+        self.assertEqual(build_info['mar_channels'], set(['expected', 'channels']))
+        self.assertTrue(build_info['version'])
+        self.assertTrue(build_info['buildid'].isdigit())
+        self.assertTrue(build_info['locale'])
+        self.assertIn('force=1', build_info['url_aus'])
+        self.assertEqual(build_info['channel'], self.software_update.update_channel.channel)
+
+    def test_force_fallback(self):
+        try:
+            self.software_update.force_fallback()
+            with open(
+                    os.path.join(
+                            self.software_update.staging_directory, 'update.status'), 'r')as f:
+                content = f.read()
+            self.assertEqual(content, 'failed: 6\n')
+        finally:
+            os.remove(os.path.join(self.software_update.staging_directory, 'update.status'))
+
+    def test_get_update_url(self):
+        update_url = self.software_update.get_update_url()
+        self.assertIn('Firefox', update_url)
+        self.assertNotIn('force=1', update_url)
+        update_url = self.software_update.get_update_url(True)
+        self.assertIn('Firefox', update_url)
+        self.assertIn('force=1', update_url)
+
+    def test_os_version(self):
+        self.assertTrue(self.software_update.os_version)
+
+    def test_staging_directory(self):
+        self.assertTrue(self.software_update.staging_directory)
+
+
+class TestUpdateChannel(FirefoxTestCase):
+
+    def setUp(self):
+        FirefoxTestCase.setUp(self)
+        self.software_update = SoftwareUpdate(self.get_marionette)
+        self.saved_channel = self.software_update.update_channel.default_channel
+        self.software_update.update_channel.default_channel = 'expected_channel'
+
+    def tearDown(self):
+        try:
+            self.software_update.update_channel.default_channel = self.saved_channel
+        finally:
+            FirefoxTestCase.tearDown(self)
+
+    def test_update_channel_channel(self):
+        self.assertEqual(self.software_update.update_channel.channel, self.saved_channel)
+
+    def test_update_channel_default_channel(self):
+        self.assertEqual(self.software_update.update_channel.default_channel, 'expected_channel')
+
+    def test_update_channel_set_default_channel(self):
+        self.software_update.update_channel.default_channel = 'new_channel'
+        self.assertEqual(self.software_update.update_channel.default_channel, 'new_channel')
+
+
+class TestMARChannels(FirefoxTestCase):
+
+    def setUp(self):
+        FirefoxTestCase.setUp(self)
+        self.software_update = SoftwareUpdate(self.get_marionette)
+        self.saved_mar_channels = self.software_update.mar_channels.channels
+        self.software_update.mar_channels.channels = set(['expected', 'channels'])
+
+    def tearDown(self):
+        try:
+            self.software_update.mar_channels.channels = self.saved_mar_channels
+        finally:
+            FirefoxTestCase.tearDown(self)
+
+    def test_mar_channels_channels(self):
+        self.assertEqual(self.software_update.mar_channels.channels, set(['expected', 'channels']))
+
+    def test_mar_channels_set_channels(self):
+        self.software_update.mar_channels.channels = set(['a', 'b', 'c'])
+        self.assertEqual(self.software_update.mar_channels.channels, set(['a', 'b', 'c']))
+
+    def test_mar_channels_add_channels(self):
+        self.software_update.mar_channels.add_channels(set(['some', 'new', 'channels']))
+        self.assertEqual(
+            self.software_update.mar_channels.channels,
+            set(['expected', 'channels', 'some', 'new']))
+
+    def test_mar_channels_remove_channels(self):
+        self.software_update.mar_channels.remove_channels(set(['expected']))
+        self.assertEqual(self.software_update.mar_channels.channels, set(['channels']))
--- a/firefox_puppeteer/docs/index.rst
+++ b/firefox_puppeteer/docs/index.rst
@@ -44,6 +44,7 @@ future. Each library is available from an instance of the FirefoxTestCase class.
    api/places
    api/prefs
    api/security
+   api/software_update
    api/utils
 
 
--- a/firefox_puppeteer/tests/manifest.ini
+++ b/firefox_puppeteer/tests/manifest.ini
@@ -4,6 +4,7 @@
 [test_places.py]
 [test_prefs.py]
 [test_security.py]
+[test_software_update.py]
 [test_tabbar.py]
 [test_toolbars.py]
 [test_windows.py]


bobsilverberg:
@whimboo This is now complete, in that all of the methods have been converted. I would not be surprised if there are some problems with `init_update_tests` and `assert_update_applied` as I'm not really sure how info is going to get populated and passed around, but hopefully this PR is a good start.

I'm going to turn my attention back to about-window.js now.

firefox_puppeteer/__init__.py
```diff
@@ -79,6 +79,14 @@ def security(self):
         See the :class:`~api.security.Security` reference.
         """
 
+    @use_class_as_property('api.software_update.SoftwareUpdate')
+    def software_update(self):
+        """
+        Provides support for an easy access to the update process.
+
+        See the :class:`~api.software_update.SoftwareUpdate` reference.
+        """
```
whimboo: I think that we don't need to expose the software update module to Puppeteer. We only have 2 tests at all using it. They can import the necessary bits when necessary.

whimboo: Please make sure to add the appropriate sphinx files so that the module gets documented.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,544 @@
...
+class MARChannels(object):
+    """Class to handle the allowed MAR channels as listed in update-settings.ini
+    """
+
+    INI_SECTION = 'Settings'
+    INI_OPTION = 'ACCEPTED_MAR_CHANNEL_IDS'
+
+    REGEX_UPDATE_CHANNEL_PREF = re.compile(r'("app\.update\.channel", ")([^"].*)(?=")')
+
+    def __init__(self, software_update):
```
whimboo: This class should be based on BaseLib instead of object, so it would require the marionette getter as parameter to the __init__ method.

whimboo: Also to add, this class needs nothing from the software_update class and should stay independent from it. If you have to pass data lets get a property added.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,544 @@
...
+
+        self.marionette = software_update.marionette
+        self.prefs = software_update.prefs
+
+        self.ini_file_path = self.marionette.execute_script("""
+            Cu.import('resource://gre/modules/Services.jsm');
```
whimboo: nit: For JS code we have an indentation of 2 chars.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,544 @@
...
+            file.append('update-settings.ini');
+
+            return file.path;
+
+            """)
+
+    @property
+    def config(self):
+        config = ConfigParser.ConfigParser()
+        config.readfp(open(self.ini_file_path))
```
whimboo: What happens in case of a failure in readfp? Is the file getting closed? I think it would be better to work with `with` here, or?

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,544 @@
...
+
+            var file = Services.dirsvc.get('GreD', Ci.nsIFile);
+            file.append('update-settings.ini');
+
+            return file.path;
+
+            """)
+
+    @property
+    def config(self):
```
whimboo: This property is for internal use. I would prefix its name with an underscore.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,544 @@
...
+        config = ConfigParser.ConfigParser()
+        config.readfp(open(self.ini_file_path))
+        return config
+
+    @property
+    def config_file_path(self):
+        return self.ini_file_path
+
+    @property
+    def config_file_contents(self):
```
whimboo: This property is also for internal use only. There is no need that we have to expose it. Lets mark it as private.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,544 @@
...
+        with open(self.ini_file_path) as f:
+            return f.read()
+
+    @property
+    def channels(self):
+        channels = self.config.get(self.INI_SECTION, self.INI_OPTION)
+        return channels.split(',')
+
+    def set_channels(self, channels):
```
whimboo: Make this a setter property instead. It simplifies the usage of this class.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,544 @@
...
+        channels = ','.join(channels)
```
whimboo: I think you can directly put this into the call to `set`.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,544 @@
...
+        new_config = self.config
+        new_config.set(self.INI_SECTION, self.INI_OPTION, channels)
+        with open(self.ini_file_path, 'wb') as f:
+            new_config.write(f)
```
whimboo: Have you checked that this produces the correct output? When I was implementing this class originally I have seen that the Python config library added blanks around the `=` sign for this key. This will break reading the ini file in Firefox.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,544 @@
...
+
+    def add_channels(self, channels):
+        new_channels = ','.join(self.channels + channels)
+        new_config = self.config
+        new_config.set(self.INI_SECTION, self.INI_OPTION, new_channels)
+        with open(self.ini_file_path, 'wb') as f:
+            new_config.write(f)
```
whimboo: This is duplicating the code from `set_channels`. Please call this method with the final list of channels instead.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,544 @@
...
+
+    def remove_channels(self, channels):
+        current_channels = self.channels
+        for channel in channels:
+            if channel in current_channels:
+                current_channels.remove(channel)
```
whimboo: Have you done list comprehensions yet? It's a very powerful future we can use here.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,544 @@
...
+        new_channels = ','.join(current_channels)
+        new_config = self.config
+        new_config.set(self.INI_SECTION, self.INI_OPTION, new_channels)
+        with open(self.ini_file_path, 'wb') as f:
+            new_config.write(f)
```
whimboo: Same as above. Please call `set_channels` instead of duplicating the code.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,544 @@
...
+
+
+class UpdateChannel(object):
+    """Class to handle the update channel as listed in channel-prefs.js
+    """
+
+    REGEX_UPDATE_CHANNEL_PREF = re.compile(r'("app\.update\.channel", ")([^"].*)(?=")')
+
+    def __init__(self, software_update):
```
whimboo: Same as for the other class. Please make BaseLib the base class.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,544 @@
...
+
+        self.marionette = software_update.marionette
+        self.prefs = software_update.prefs
+
+        self.config_file_path = self.marionette.execute_script("""
```
whimboo: Maybe just `file_path`? It's not really a config file.

firefox_puppeteer/api/software_update.py
whimboo: And may name it `_content`?

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,544 @@
...
+
+            var file = Services.dirsvc.get('PrfDef', Ci.nsIFile);
+            file.append('channel-prefs.js');
+
+            return file.path;
+
+            """)
+
+    @property
+    def config_file_contents(self):
```
whimboo: Maybe just `_content`?

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,544 @@
...
+    def channel(self):
+        return self.prefs.get_pref('app.update.channel', True)
+
+    @property
+    def default_channel(self):
+        """Get the default update channel
+
+        :returns: Current default update channel
+        """
+        matches_list = re.findall(self.REGEX_UPDATE_CHANNEL_PREF, self.config_file_contents)
```
whimboo: We only have a single entry here. What would be the benefit of `findall` compared to `search`?

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,544 @@
...
+        assert len(matches_list) == 1, 'Update channel value has been found'
+        matches_tuple = matches_list[0]
+        assert len(matches_tuple) == 2, 'Update channel value has been found'
```
whimboo: Additionally to the above question I think `search` would make it easier here.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,544 @@
...
+
+        return matches_tuple[1]
+
+    def set_default_channel(self, channel):
```
whimboo: Please make it also a setter property.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,544 @@
...
+        new_config.set(self.INI_SECTION, self.INI_OPTION, new_channels)
+        with open(self.ini_file_path, 'wb') as f:
+            new_config.write(f)
+
+
+class UpdateChannel(object):
+    """Class to handle the update channel as listed in channel-prefs.js
+    """
+
+    REGEX_UPDATE_CHANNEL_PREF = re.compile(r'("app\.update\.channel", ")([^"].*)(?=")')
```
whimboo: nit: It's not a pref but a pref_name. But I would simply drop this suffix.

firefox_puppeteer/tests/test_software_update.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def test_staging_directory(self):
+        self.assertTrue(self.software_update.staging_directory)
+
+    def test_update_url(self):
+        self.assertIn('Firefox', self.software_update.update_url())
+        self.assertIn('Firefox', self.software_update.update_url(True))
+        self.assertIn('force=1', self.software_update.update_url(True))
+
+    def test_update_channel_channel(self):
```
whimboo: Each class in the software_update.py file should get it's own test class in this file.

firefox_puppeteer/api/software_update.py
bobsilverberg: This is actually used in `init_update_tests`, which is in `SoftwareUpdate` so I think it needs to remain public.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,544 @@
...
+        self.prefs = Preferences(marionette_getter)
+
+        self.update_channel = UpdateChannel(self)
+        self.mar_channels = MARChannels(self)
+        self.active_update = ActiveUpdate(self)
+
+        self._aus = self.marionette.execute_script("""
+            return Cc['@mozilla.org/updates/update-service;1']
+                .getService(Ci.nsIApplicationUpdateService);
+            """)
```
whimboo: Lets not return the service here. It will be JSON string but no real service anymore, and you won't be able to convert it back. We can access properties from it but no methods. So lets simply add this code to all the methods where it is used.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,544 @@
...
+        """Get the customized ABI for the update service.
+
+        :returns: ABI version
+        """
+
+        abi = self.app_info.XPCOMABI
+        if mozinfo.isMac:
+            abi += self.marionette.execute_script("""
+                let macutils = Cc['@mozilla.org/xpcom/mac-utils;1']
+                    .getService(Ci.nsIMacUtils);
```
whimboo: For a continuation line the dot should be aligned with the `C`.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,544 @@
...
+                if (macutils.isUniversalBinary)
+                    return '-u-' + macutils.architecturesInBinary;
```
whimboo: Please always use brackets for blocks in JS.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,544 @@
...
+
+    def __init__(self, marionette_getter):
+        BaseLib.__init__(self, marionette_getter)
+
+        self.app_info = AppInfo(marionette_getter)
+        self.prefs = Preferences(marionette_getter)
+
+        self.update_channel = UpdateChannel(self)
+        self.mar_channels = MARChannels(self)
+        self.active_update = ActiveUpdate(self)
```
whimboo: Those would have to appear in the sphinx docs. Not sure how you would do that. I propose you assign the instances to private variables and create a property for each of those.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,544 @@
...
+        return abi
+
+    @property
+    def allowed(self):
+        """Check if the user has permissions to run the software update
+
+        :returns: Status if the user has the permissions
+        """
+
+        return self._aus['canCheckForUpdates'] and self._aus['canApplyUpdates']
```
whimboo: As mentioned above lets work on the real aus service object.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,544 @@
...
+
+        # The upgraded version should be identical with the version given by
+        # the update and we shouldn't have run a downgrade
+        check = self.marionette.execute_script("""
+            Cu.import("resource://gre/modules/Services.jsm");
+
+            return  Services.vc.compare(arguments[0], arguments[1]);
+        """, script_args=[info['build_post']['version'], info['build_pre']['version']])
+
+        assert check >= 0, 'The version of the upgraded build is higher or equal'
```
whimboo: Maybe it is better to use assert methods from the unittest class here.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,544 @@
...
+    def build_info(self):
+        """Return information of the current build version
+
+        :returns: A dictionary of build information
+        """
+
+        return {
+            'buildid': self.app_info.appBuildID,
+            'channel': self.update_channel.channel,
+            'disabled_addons': self.prefs.get_pref('PREF_DISABLED_ADDONS'),
```
whimboo: The parameter in `get_pref` please without the quotes. :)

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,544 @@
...
+    @property
+    def allowed(self):
+        """Check if the user has permissions to run the software update
+
+        :returns: Status if the user has the permissions
+        """
+
+        return self._aus['canCheckForUpdates'] and self._aus['canApplyUpdates']
+
+    def assert_update_applied(self, update_data):
```
whimboo: I would suggest we also keep the following order in a class: properties, and then methods. If it's not such a hassle to move some parts around here.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,544 @@
...
+        :returns: nsIUpdate object
+        """
+
+        return self.marionette.execute_script("""
+            let ums = Cc['@mozilla.org/updates/update-manager;1']
+                .getService(Ci.nsIUpdateManager);
+            let update_index = arguments[0];
+            return ums.getUpdateAt(update_index);
+
+        """, script_args=[update_index])
```
whimboo: Same as for the aus service above. We cannot return a JS object. It will be a JSON string and cannot be used anymore. :S

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,544 @@
...
+                    'success': False}]}
+
+        # If requested modify the default update channel. It will be active
+        # after the next restart of the application
+        if persisted['update']['channel']:
+            # Backup the original content and the path of the channel-prefs.js file
+            persisted['update']['default_update_channel'] = {
+                'content': self.update_channel.config_file_contents,
+                'path': self.update_channel.config_file_path}
+            self.update_channel.set_default_channel(persisted['update']['channel'])
```
whimboo: Given that the UpdateChannel class lives in Python now we won't have to store the defaults anymore, but simply ensure to restore the file with its original content at the end of the test. Maybe we can call this method `setUp` and add a new `tearDown` method to do that?

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,544 @@
...
+
+        # If requested modify the list of allowed MAR channels
+        if persisted['update']['allowed_mar_channels']:
+            # Backup the original content and the path of the update-settings.ini file
+            persisted['update']['default_mar_channels'] = {
+                'content': self.mar_channels.config_file_contents,
+                'path': self.mar_channels.config_file_path}
```
whimboo: Same here. We can directly restore the content at the end of the test.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,544 @@
...
+        :returns: True if the offered update is a complete update
+        """
+
+        # Throw when isCompleteUpdate is called without an update. This should
+        # never happen except if the test is incorrectly written.
+        patch_count = self.active_update.patch_count
+        assert patch_count, 'An active update has been found'
+
+        assert patch_count == 1 or patch_count == 2,\
+            'An update must have one or two patches included'
```
whimboo: Can we use the `unittest` equivalent methods here too?

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,544 @@
...
+    def is_complete_update(self):
+        """Return true if the offered update is a complete update
+
+        :returns: True if the offered update is a complete update
+        """
+
+        # Throw when isCompleteUpdate is called without an update. This should
+        # never happen except if the test is incorrectly written.
+        patch_count = self.active_update.patch_count
+        assert patch_count, 'An active update has been found'
```
whimboo: You have added `active_update.exists` so I think it is what we should use here.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,544 @@
...
+
+        return info
+
+    @property
+    def staging_directory(self):
+        return self.marionette.execute_script("""
+            let aus = Cc['@mozilla.org/updates/update-service;1']
+                .getService(Ci.nsIApplicationUpdateService);
+            var dir = aus.getUpdatesDirectory();
+            return dir.path;
```
whimboo: I would directly do a `return aus.getUpdatesDirectory().path` here.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,544 @@
...
+        """)
+
+    @property
+    def update_type(self):
+        """Returns the type of the active update."""
+        return self.active_update.type
+
+    def update_url(self, force=False):
```
whimboo: This is not a property, so please add the `get_` prefix to the method name.

firefox_puppeteer/api/software_update.py
bobsilverberg: You are correct. My tests didn't catch that. I guess we cannot use `ConfigParser`, which is too bad. :(

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,544 @@
...
+
+        :param force: Boolean flag to force an update check
+
+        :returns: The URL of the update snippet
+        """
+
+        url = self.prefs.get_pref(self.PREF_APP_UPDATE_URL, '')
+        # get the next two prefs from the default branch
+        dist = self.prefs.get_pref(self.PREF_APP_DISTRIBUTION, True) or 'default'
+        dist_version = self.prefs.get_pref(self.PREF_APP_DISTRIBUTION_VERSION, True) or 'default'
```
whimboo: Move those 3 lines after the following if block, so we don't run this code if url is empty.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,544 @@
...
+
+        if not url:
+            return None
+
+        # Not all placeholders are getting replaced correctly by formatURL
+        url = re.sub(r'%PRODUCT%', self.app_info.name, url)
```
whimboo: I would suggest we keep using the string replace method here, which is much faster than any regex.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,544 @@
...
+        url = re.sub(r'%BUILD_TARGET%', self.app_info.OS + '_' + self.ABI, url)
+        url = re.sub(r'%OS_VERSION%', self.os_version, url)
+        url = re.sub(r'%CHANNEL%', self.update_channel.channel, url)
+        url = re.sub(r'%DISTRIBUTION%', dist, url)
+        url = re.sub(r'%DISTRIBUTION_VERSION%', dist_version, url)
+
+        url = self.marionette.execute_script("""
+            Cu.import("resource://gre/modules/Services.jsm");
+
+            var url = arguments[0];
```
whimboo: Please lets use `let` in all the places.

whimboo: In this case we could even get rid of this line.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,544 @@
...
+        if force:
+            if '?' in url:
+                url += '&'
+            else:
+                url += '?'
+            url += 'force=1'
+        return url
+
+
+class ActiveUpdate(object):
```
whimboo: Same here for BaseLib as super class.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,544 @@
...
+
+    def __getattr__(self, attr):
+        value = self.marionette.execute_script("""
+            let ums = Cc['@mozilla.org/updates/update-manager;1']
+                .getService(Ci.nsIUpdateManager);
+            var active_update = ums.activeUpdate;
+            return active_update[arguments[0]];
+        """, script_args=[attr])
+
+        if value is not None:
```
whimboo: `if value` should be enough.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,544 @@
...
+
+        :returns: nsIUpdatePatch object
+        """
+
+        return self.marionette.execute_script("""
+            let ums = Cc['@mozilla.org/updates/update-manager;1']
+                .getService(Ci.nsIUpdateManager);
+            let active_update = ums.activeUpdate;
+            let patch_index = arguments[0];
+            return active_update.getPatchAt(patch_index);
```
whimboo: I checked DXR and so far we really only use properties so we should be fine in returning the JSON data for the interface.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,544 @@
...
+        """Get the patchCount from the active update
+
+        :returns: The patch count
+        """
+
+        return self.marionette.execute_script("""
+            let ums = Cc['@mozilla.org/updates/update-manager;1']
+                .getService(Ci.nsIUpdateManager);
+            let active_update = ums.activeUpdate;
+            return active_update.patchCount;
```
whimboo: Let directly return without the extra variable.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,544 @@
...
+        """Get the finalURL of the selected patch for the active update
+
+        :returns: The patch size
+        """
+
+        return self.marionette.execute_script("""
+            let ums = Cc['@mozilla.org/updates/update-manager;1']
+                .getService(Ci.nsIUpdateManager);
+            let active_update = ums.activeUpdate;
+            return active_update.selectedPatch.finalURL;
```
whimboo: Same here. return directly.

whimboo: Hm, maybe we can also have a `selected_patch` property only and return the JSON object? That way we have only a single method and using that property will become easier.

bobsilverberg:
@whimboo I have added a commit that addresses the comments for MarChannels. I plan on adding a commit for each class, as I address them, You may want to wait for all of my updates before re-reviewing. I will add a comment when they are all complete.

firefox_puppeteer/api/software_update.py
bobsilverberg: Good point, `search` is better.

firefox_puppeteer/api/software_update.py
whimboo: Lets keep it for now until we can make use of a Pythonic INI reader and writer. We can refactor this out later.

firefox_puppeteer/api/software_update.py
whimboo: > You are correct. My tests didn't catch that. I guess we cannot use
> |ConfigParser|, which is too bad. :(

Please file a new follow-up bug for that. Maybe we can find a hacker who
wants to take it.

firefox_puppeteer/api/software_update.py
bobsilverberg: That is not what is currently being done in `prefs` (e.g., [1])  or `windows` [2]. I can make that change if you wish, but I wanted to check with you first for consistency.

[1] https://github.com/mozilla/firefox-ui-tests/blob/master/firefox_puppeteer/api/prefs.py#L153
[2] https://github.com/mozilla/firefox-ui-tests/blob/master/firefox_puppeteer/ui/windows.py#L318

firefox_puppeteer/api/software_update.py
bobsilverberg: It turns out I never need to call this `get_update_at` anyway. I think I added it while testing some things, so I will just remove the entire method.

bobsilverberg:
@whimboo I have addressed all of your review comments, except for the one about `init_update_tests` as I have to admit I'm not really sure what is going on in that method and what actually needs to be done. If you can re-review, then maybe we can have a short meeting to discuss what's left, including `init_update_tests`.

firefox_puppeteer/api/software_update.py
whimboo: >> +        assert check >= 0, 'The version of the upgraded build is higher or equal'
> 
> That is not what is currently being done in |prefs| (e.g., [1]) or
> |windows| [2]. I can make that change if you wish, but I wanted to check
> with you first for consistency.
> 
> [1]
> https://github.com/mozilla/firefox-ui-tests/blob/master/firefox_puppeteer/api/prefs.py#L153
> [2]
> https://github.com/mozilla/firefox-ui-tests/blob/master/firefox_puppeteer/ui/windows.py#L318

In those other libs we use `assert` for checking parameters. But I see
your point. @chmanchester what do you think? Thinking more about it, it
may not a good idea to base library modules on the unittest module.

chmanchester: This is somewhere between a library and a test, I think either would be acceptable.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,491 @@
...
+        """
+
+        return self.marionette.execute_script("""
+          let ums = Cc['@mozilla.org/updates/update-manager;1']
+                    .getService(Ci.nsIUpdateManager);
+          return ums.activeUpdate.selectedPatch;
+        """)
+
+
+class MARChannels(BaseLib):
```
whimboo: Mind moving this class up so we have an alphabetical sorting?

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,491 @@
...
+    INI_SECTION = 'Settings'
+    INI_OPTION = 'ACCEPTED_MAR_CHANNEL_IDS'
+
+    def __init__(self, marionette_getter):
+        BaseLib.__init__(self, marionette_getter)
+
+        self.ini_file_path = self.marionette.execute_script("""
+          Cu.import('resource://gre/modules/Services.jsm');
+
+          let file = Services.dirsvc.get('GreD', Ci.nsIFile);
```
whimboo: I would advise to not use the shortcuts for `Components.*` here. It will fail if this code is executed from a window which does not have those defined. Please change that across the file.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,491 @@
...
+
+class MARChannels(BaseLib):
+    """Class to handle the allowed MAR channels as listed in update-settings.ini
+    """
+
+    INI_SECTION = 'Settings'
+    INI_OPTION = 'ACCEPTED_MAR_CHANNEL_IDS'
+
+    def __init__(self, marionette_getter):
+        BaseLib.__init__(self, marionette_getter)
```
whimboo: Please add a docstring here.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,491 @@
...
+
+          let file = Services.dirsvc.get('GreD', Ci.nsIFile);
+          file.append('update-settings.ini');
+
+          return file.path;
+          """)
+
+    @property
+    def config_file_path(self):
+        return self.ini_file_path
```
whimboo: Same here and other places for docstrings.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,491 @@
...
+    def channels(self):
+
+        channels = self.marionette.execute_script("""
+          Cu.import('resource://gre/modules/Services.jsm');
+
+          let iniFactory = Cc['@mozilla.org/xpcom/ini-processor-factory;1']
+                           .getService(Ci.nsIINIParserFactory);
+
+          let file = Services.dirsvc.get('GreD', Ci.nsIFile);
+          file.append('update-settings.ini');
```
whimboo: We have the path as `self.ini_file_path` so please push this in here as a script_arg.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,491 @@
...
+        return channels.split(',')
+
+    @channels.setter
+    def channels(self, channels):
+        """Set the channels in the ini file
+
+        :param channels: a comma delimited string of channel names
+        """
+
+        self.marionette.execute_script("""
```
whimboo: nit: There is not a blank line after a docstring.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,491 @@
...
+          Cu.import('resource://gre/modules/Services.jsm');
+
+          let iniFactory = Cc['@mozilla.org/xpcom/ini-processor-factory;1']
+                           .getService(Ci.nsIINIParserFactory);
+
+          let file = Services.dirsvc.get('GreD', Ci.nsIFile);
+          file.append('update-settings.ini');
```
whimboo: Same here as above for the file path.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,491 @@
...
+
+          let writer = iniFactory.createINIParser(file)
+                       .QueryInterface(Ci.nsIINIParserWriter);
+
+          writer.setString(arguments[0], arguments[1], arguments[2]);
+          writer.writeFile(null, Ci.nsIINIParserWriter.WRITE_UTF16);
+          """, script_args=[self.INI_SECTION, self.INI_OPTION, channels])
+
+    def add_channels(self, channels):
+        new_channels = ','.join(self.channels) + ',' + channels
```
whimboo: This would add duplicates to the list. I would suggest that we pass in an array of channels here and not a string, so we can work with a Set().

whimboo: Also we should do the stringification inside the channels getter and setter.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,491 @@
...
+        self.channels = new_channels
+
+    def remove_channels(self, channels):
+        current_channels = self.channels
+        new_channels = [channel for channel in current_channels if channel not in channels]
+        new_channels = ','.join(new_channels)
```
whimboo: Same here as for `add_channels`.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,491 @@
...
+          writer.writeFile(null, Ci.nsIINIParserWriter.WRITE_UTF16);
+          """, script_args=[self.INI_SECTION, self.INI_OPTION, channels])
+
+    def add_channels(self, channels):
+        new_channels = ','.join(self.channels) + ',' + channels
+        self.channels = new_channels
+
+    def remove_channels(self, channels):
+        current_channels = self.channels
+        new_channels = [channel for channel in current_channels if channel not in channels]
```
whimboo: When looking at this it reminds me of a comment from @chmanchester and that we should better use `Set() - Set()` here. Maybe we can hold `self.channels` as a set on that class directly.

firefox_puppeteer/docs/api/software_update.rst
```diff
@@ -0,0 +1,27 @@
...
+ActiveUpdate
+------------
+
+.. autoclass:: ActiveUpdate
+   :members:
+
+MARChannels
+-----------
+
+.. autoclass:: ActiveUpdate
```
whimboo: c&p failure.

firefox_puppeteer/docs/api/software_update.rst
```diff
@@ -0,0 +1,27 @@
...
+   :members:
+
+UpdateChannel
+-------------
+
+.. autoclass:: ActiveUpdate
```
whimboo: c&p failure.

firefox_puppeteer/docs/api/software_update.rst
```diff
@@ -0,0 +1,27 @@
...
+
+The SoftwareUpdate class provides helpers for update tests.
+
+.. autoclass:: SoftwareUpdate
+   :members:
+
+ActiveUpdate
+------------
+
+.. autoclass:: ActiveUpdate
```
whimboo: c&p failure.

firefox_puppeteer/tests/test_software_update.py
```diff
@@ -0,0 +1,108 @@
...
+        self.assertTrue(build_info['buildid'].isdigit())
+        self.assertTrue(build_info['locale'])
+        self.assertIn('force=1', build_info['url_aus'])
+        self.assertEqual(self.software_update.update_channel.channel, build_info['channel'])
+
+    def test_force_fallback(self):
+        self.software_update.force_fallback()
+        with open(os.path.join(self.software_update.staging_directory, 'update.status'), 'r') as f:
+            content = f.read()
+        self.assertEqual(content, 'failed: 6\n')
```
whimboo: To prevent follow-up failures you might want to remove this file in tearDown or maybe better have a try/finally here.

firefox_puppeteer/tests/test_software_update.py
```diff
@@ -0,0 +1,108 @@
...
+    def test_os_version(self):
+        self.assertTrue(self.software_update.os_version)
+
+    def test_staging_directory(self):
+        self.assertTrue(self.software_update.staging_directory)
+
+    def test_get_update_url(self):
+        self.assertIn('Firefox', self.software_update.get_update_url())
+        self.assertIn('Firefox', self.software_update.get_update_url(True))
+        self.assertIn('force=1', self.software_update.get_update_url(True))
```
whimboo: Maybe caching the call to `get_update_url()`? Should we also add `assertNotIn()` for `force=1` if not forced?

firefox_puppeteer/tests/test_software_update.py
```diff
@@ -0,0 +1,108 @@
...
+from firefox_puppeteer.api.software_update import SoftwareUpdate
+
+
+class TestSoftwareUpdate(FirefoxTestCase):
+
+    def setUp(self):
+        FirefoxTestCase.setUp(self)
+        self.software_update = SoftwareUpdate(self.get_marionette)
+        self.saved_mar_channels = self.software_update.mar_channels.channels
+        self.software_update.mar_channels.channels = 'expected,channels'
```
whimboo: MAR channel logic is not needed in this test class.

firefox_puppeteer/tests/test_software_update.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def setUp(self):
+        FirefoxTestCase.setUp(self)
+        self.software_update = SoftwareUpdate(self.get_marionette)
+        self.saved_channel = self.software_update.update_channel.default_channel
+        self.software_update.update_channel.default_channel = 'expected_channel'
+
+    def tearDown(self):
+        self.software_update.update_channel.default_channel = self.saved_channel
+        FirefoxTestCase.tearDown(self)
```
whimboo: I would suggest a try/finally here so we can ensure that the super `tearDown` call gets executed.

firefox_puppeteer/tests/test_software_update.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def test_update_channel_channel(self):
+        self.assertEqual(self.saved_channel, self.software_update.update_channel.channel)
+
+    def test_update_channel_default_channel(self):
+        self.assertEqual('expected_channel', self.software_update.update_channel.default_channel)
+
+    def test_update_channel_set_default_channel(self):
+        self.software_update.update_channel.default_channel = 'new_channel'
+        self.assertEqual('new_channel', self.software_update.update_channel.default_channel)
```
whimboo: Please flip so that the reference comes as second argument.

firefox_puppeteer/tests/test_software_update.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def setUp(self):
+        FirefoxTestCase.setUp(self)
+        self.software_update = SoftwareUpdate(self.get_marionette)
+        self.saved_mar_channels = self.software_update.mar_channels.channels
+        self.software_update.mar_channels.channels = 'expected,channels'
+
+    def tearDown(self):
+        self.software_update.mar_channels.channels = self.saved_mar_channels
+        FirefoxTestCase.tearDown(self)
```
whimboo: Same for try/finally here.

firefox_puppeteer/tests/test_software_update.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def test_mar_channels_channels(self):
+        self.assertEqual(['expected', 'channels'], self.software_update.mar_channels.channels)
```
whimboo: Please flip for those tests too.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,491 @@
...
+        self.active_update = ActiveUpdate(marionette_getter)
+
+    @property
+    def ABI(self):
+        """Get the customized ABI for the update service.
+
+        :returns: ABI version
+        """
+
+        abi = self.app_info.XPCOMABI
```
whimboo: nit: no blank line after a docstring. Please also check other instances.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,491 @@
...
+
+    def __init__(self, marionette_getter):
+        BaseLib.__init__(self, marionette_getter)
+
+        self.app_info = AppInfo(marionette_getter)
+        self.prefs = Preferences(marionette_getter)
+
+        self.update_channel = UpdateChannel(marionette_getter)
+        self.mar_channels = MARChannels(marionette_getter)
+        self.active_update = ActiveUpdate(marionette_getter)
```
whimboo: Those properties would not appear in the sphinx docs. Maybe we can tag them with a special flag?

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,491 @@
...
+    @property
+    def allowed(self):
+        """Check if the user has permissions to run the software update
+
+        :returns: Status if the user has the permissions
+        """
+
+        aus_dict = self.marionette.execute_script("""
+          return Cc['@mozilla.org/updates/update-service;1']
+            .getService(Ci.nsIApplicationUpdateService);
```
whimboo: nit: please align the dot with `C`

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,491 @@
...
+          """)
+
+        return aus_dict['canCheckForUpdates'] and aus_dict['canApplyUpdates']
```
whimboo: Maybe better to directly do the check in JS and return a boolean via `execute_script`?

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,491 @@
...
+                self.active_update.selected_patch['finalURL'] or 'n/a'
+            info['version'] = self.active_update.appVersion
+
+        return info
+
+    @property
+    def staging_directory(self):
+        return self.marionette.execute_script("""
+          let aus = Cc['@mozilla.org/updates/update-service;1']
+                   .getService(Ci.nsIApplicationUpdateService);
```
whimboo: nit: indentation.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,491 @@
...
+                    'fallback': fallback,
+                    'patch': {},
+                    'success': False}]}
+
+        # If requested modify the default update channel. It will be active
+        # after the next restart of the application
+        if persisted['update']['channel']:
+            # Backup the original content and the path of the channel-prefs.js file
+            persisted['update']['default_update_channel'] = {
+                'content': self.update_channel.file_contents,
```
whimboo: As we talked in the meeting yesterday please file a follow-up bug so we do copies of the files instead.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,491 @@
...
+        url = url.replace('%CHANNEL%', self.update_channel.channel)
+        url = url.replace('%DISTRIBUTION%', dist)
+        url = url.replace('%DISTRIBUTION_VERSION%', dist_version)
+
+        url = self.marionette.execute_script("""
+          Cu.import("resource://gre/modules/Services.jsm");
+
+          return Services.urlFormatter.formatURL(arguments[0]);
+
+          """, script_args=[url])
```
whimboo: nit: indentation issue. Align it with `u` from url. Maybe drop the blank lines in the JS code.

whimboo:
I made a couple of comments for the current code. It all looks fine code-wise in nearly all cases. When working on the next commit please also rebase against master. Looks like there are merge conflicts.

firefox_puppeteer/api/software_update.py
bobsilverberg: I'm sorry, @whimboo, but I cannot tell what the context of this comment was. I'm not sure what the "Same here." refers to. Can you be more explicit about what needs to change in here? I'm still not totally clear about what we're planning to do to get rid of `persisted` and if it means we need more major changes in here.

firefox_puppeteer/api/software_update.py
bobsilverberg: I'm not entirely sure what you mean. Do you mean replace `Ci.` with `Components.interfaces.`? Do I also need an import for that before calling it?

firefox_puppeteer/api/software_update.py
bobsilverberg: A docstring for `__init__`? None of the other classes have docstrings for their `__init__` method. I would think that would be somewhat superfluous. Or do you mean a comment to describe the code below, that sets the path for the ini file?

firefox_puppeteer/api/software_update.py
bobsilverberg: I could add properties for them, but that seems a bit redundant. Do you know how to flag them to be picked up for the docs, or is that something you want me to figure out?

firefox_puppeteer/api/software_update.py
whimboo: Good question! I assume it was also related to restoring the content of the file. With the new bug to be filed we can cover both cases. So just file the bug as mentioned above for the update-settings.ini file.

firefox_puppeteer/api/software_update.py
whimboo: Correct. `Ci.` will only work inside a browser window but will most likely fail if used in other chrome window types. Given that you are in chrome scope no import or such things are necessary.

firefox_puppeteer/api/software_update.py
whimboo: You are right. We should do a follow-up bug for all those cases. No need to do it now.

firefox_puppeteer/api/software_update.py
whimboo: I don't know about the details, but I would propose we make the instances private by adding a _ prefix to them. Then only add a getter property for each of them. That way we can ensure that no outer code can mess up with those instances.

firefox_puppeteer/tests/test_software_update.py
bobsilverberg: I included it because `build_info` includes the MARChannels, so I wanted to be able to set an expected value for that in order to do the assert. I can remove it and just use `assertTrue` if you would prefer that.

firefox_puppeteer/api/software_update.py
bobsilverberg: Done. https://bugzilla.mozilla.org/show_bug.cgi?id=1138946

firefox_puppeteer/tests/test_software_update.py
whimboo: > I included it because |build_info| includes the MARChannels, so I wanted
> to be able to set an expected value for that in order to do the assert.
> I can remove it and just use |assertTrue| if you would prefer that.

I see. That makes sense then. So lets leave it as is. Thanks.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,499 @@
...
+from ..base import BaseLib
+from .appinfo import AppInfo
+from .prefs import Preferences
+
+
+class ActiveUpdate(BaseLib):
+
+    def __getattr__(self, attr):
+        value = self.marionette.execute_script("""
+          let ums = Cc['@mozilla.org/updates/update-manager;1']
```
whimboo: You still have `Cc` in your patch which would also break us. Please also use `Components.classes` everywhere.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,499 @@
...
+
+        if value:
+            return value
+        else:
+            raise AttributeError('{} has no attribute {}'.format(self.__class__.__name__,
+                                                                 attr))
+
+    @property
+    def exists(self):
+        """Is there an active update?
```
whimboo: nit: We don't use questions everywhere else. So just `Checks if there is an active update`.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,499 @@
...
+
+class MARChannels(BaseLib):
+    """Class to handle the allowed MAR channels as listed in update-settings.ini."""
+    INI_SECTION = 'Settings'
+    INI_OPTION = 'ACCEPTED_MAR_CHANNEL_IDS'
+
+    def __init__(self, marionette_getter):
+        BaseLib.__init__(self, marionette_getter)
+
+        self.ini_file_path = self.marionette.execute_script("""
```
whimboo: With the `config_file_path` property on this class we should make this property private.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,499 @@
...
+
+          return parser.getString(arguments[1], arguments[2]);
+        """, script_args=[self.ini_file_path, self.INI_SECTION, self.INI_OPTION])
+        return channels.split(',')
+
+    @channels.setter
+    def channels(self, channels):
+        """Set the channels in the update-settings.ini file.
+
+        :param channels: an array of channel names
```
whimboo: nit: Start the description with a capital letter.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,499 @@
...
+          writer.writeFile(null, Components.interfaces.nsIINIParserWriter.WRITE_UTF16);
+        """, script_args=[self.ini_file_path, self.INI_SECTION, self.INI_OPTION, new_channels])
+
+    def add_channels(self, channels):
+        """Add channels to the update-settings.ini file.
+
+        :param channels: an array of channel names to add
+        """
+        new_channels = set(self.channels) | set(channels)
+        self.channels = new_channels
```
whimboo: As mentioned in my last review cycle you can make `self.channels` to be a Set by default. So `channels` can return a Set instead of an array. That would make it easier across the file to use it. Maybe also assign it directly?

firefox_puppeteer/tests/test_software_update.py
```diff
@@ -0,0 +1,122 @@
...
+        self.assertIn('force=1', build_info['url_aus'])
+        self.assertEqual(build_info['channel'], self.software_update.update_channel.channel)
+
+    def test_force_fallback(self):
+        try:
+            self.software_update.force_fallback()
+            with open(
+                    os.path.join(
+                            self.software_update.staging_directory, 'update.status'), 'r')as f:
+                content = f.read()
```
whimboo: The indentation of the lines above look a bit strange but for a test it will be fine.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,499 @@
...
+            patch0_URL = self.active_update.get_patch_at(0)['URL']
+            patch1_URL = self.active_update.get_patch_at(1)['URL']
+            assert patch0_URL != patch1_URL,\
+                'Partial and Complete download URLs are different'
+
+        return self.active_update.selected_patch['type'] == 'complete'
+
+    @property
+    def mar_channels(self):
+        """ Holds a reference to a MARChannels object."""
```
whimboo: nit: you may want to linkify the class name like ":class:MARChannels". This might also apply to some other methods.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,499 @@
...
+
+        :returns: Status if the user has the permissions
+        """
+        allowed = self.marionette.execute_script("""
+          let aus = Cc['@mozilla.org/updates/update-service;1']
+                    .getService(Components.interfaces.nsIApplicationUpdateService);
+          return aus.canCheckForUpdates && aus.canApplyUpdates;
+        """)
+
+        return bool(allowed)
```
whimboo: `execute_script` already returns a boolean value, so you can directly return it.

whimboo:
I went through the code again by only watching those areas I have commented on earlier, and you applied fixes. I hope you didn't change something else. If you did please let me know, otherwise I have to do a full review again. 

Most of the remaining changes can quickly be done, and as you see the number of comments decreases rapidly. So I think that we can get it merged by tomorrow. Thanks a lot Bob!

firefox_puppeteer/api/software_update.py
bobsilverberg: Sorry. This is the first time I've ever worked with any of this Firefox API stuff, so I'm not really sure what I'm doing. Hopefully `Cc` is the last one that needs to be changed.

bobsilverberg:
@whimboo I have only been making changes specified by your comments, so that should be fine. Also, I always ensure to make all of the changes for a review cycle in a single commit, so you can always look at the diff for the commit to see exactly what has changed since the last review.

I have now addressed the most recent set of comments and pushed a new commit.

firefox_puppeteer/api/software_update.py
whimboo: > Sorry. This is the first time I've ever worked with any of this Firefox
> API stuff, so I'm not really sure what I'm doing. Hopefully |Cc| is the
> last one that needs to be changed.

No worries Bob! You are doing great on all that. Yes, `Cc` was the last
shotcut which needed to be ripped off. Thanks.

firefox_puppeteer/api/software_update.py
```diff
@@ -0,0 +1,496 @@
...
+        :param channels: A set of channel names to add
+        """
+        self.channels = self.channels | channels
+
+    def remove_channels(self, channels):
+        """Remove channels from the update-settings.ini file.
+
+        :param channels: A set of channel names to remove
+        """
+        self.channels = self.channels - channels
```
whimboo: Much cleaner and with more performance!  This looks way better.

whimboo:
This PR has been merged as https://github.com/mozilla/firefox-ui-tests/commit/2c3c1df786fe8843fb89c3c85f3eab7e8ecb73cc

PR #92, opened 2015-02-19
Bug 1132699 - Convert Mozmill testSafeBrowsingWarningPages

galgeek:
https://bugzilla.mozilla.org/show_bug.cgi?id=1132699

This test requires the same permissions library required by the other SafeBrowsing test, but I've not included it here.  I'll create another PR for it.


galgeek:
I've added the permissions library here to facilitate testing.

firefox_puppeteer/__init__.py
```diff
@@ -47,6 +47,14 @@ def keys(self):
         See the :class:`~api.keys.Keys` reference.
         """
 
+    @use_class_as_property('api.permissions.Permissions')
+    def perms(self):
+        """
+        Provides an api for interacting with permissions.
```
whimboo: The description comes right after the triple quotes.

firefox_puppeteer/__init__.py
```diff
@@ -47,6 +47,14 @@ def keys(self):
...
+
+        See the :class:`~api.permissions.Permissions` reference.
```
whimboo: You will also have to add the appropriate documents under docs/api for the new library. Let me know which PR you want to use for that implementation so we don't have to share that information between the two for safebrowsing.

firefox_puppeteer/api/permissions.py
```diff
@@ -0,0 +1,14 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from ..base import BaseLib
+
+
+class Permissions(BaseLib):
+    def remove(self, host, permission):
```
whimboo: This needs documentation of the parameters, and a unit test. So maybe it would be wiser to file a separate bug for its implementation.

firefox_puppeteer/api/permissions.py
```diff
@@ -0,0 +1,14 @@
...
+        with self.marionette.using_context('chrome'):
```
whimboo: We are in chrome context by default. So no need to switch it here.

firefox_puppeteer/api/permissions.py
```diff
@@ -0,0 +1,14 @@
...
+            self.marionette.execute_script("""
+            Cu.import("resource://gre/modules/Services.jsm");
+            Services.perms.remove(arguments[0], arguments[1]);
```
whimboo: Please indent by 2 char for js code.

firefox_ui_tests/remote/security/test_safe_browsing_warning_pages.py
```diff
@@ -0,0 +1,116 @@
...
+            # Malware URL
+            "https://www.itisatrap.org/firefox/its-an-attack.html"
+            ]
+
+        self.prefs.set_pref('browser.safebrowsing.enabled', True)
+        self.prefs.set_pref('browser.safebrowsing.malware.enabled', True)
+
+        # Give the browser a little time, because SafeBrowsing.jsm takes a
+        # while between start up and adding the example urls to the db.
+        time.sleep(2)
```
whimboo: Is this really necessary? I thought the problem initially was that both prefs were not set to `True`.

firefox_ui_tests/remote/security/test_safe_browsing_warning_pages.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def tearDown(self):
+        self.perms.remove('www.itisatrap.org', 'safe-browsing')
+        self.prefs.restore_all_prefs
```
whimboo: That's done automatically by FirefoxTestCase.tearDown. Also that would be a method to call.

firefox_ui_tests/remote/security/test_safe_browsing_warning_pages.py
```diff
@@ -0,0 +1,116 @@
...
+        try:
+            self.windows.close_all([self.browser])
```
whimboo: I don't see that this test is opening new windows. So this line is not necessary. Instead put the call to `self.perms.remove` in here.

firefox_ui_tests/remote/security/test_safe_browsing_warning_pages.py
```diff
@@ -0,0 +1,116 @@
...
+        with self.marionette.using_context("content"):
+            for unsafe_page in self.urls:
+                # Load a test page and test the get me out button
+                self.marionette.navigate(unsafe_page)
+                self.check_get_me_out_of_here_button(unsafe_page)
+
+                # Load the test page again and test the report button
+                self.marionette.navigate(unsafe_page)
+
+                self.check_report_button(unsafe_page)
```
whimboo: nit: kill the empty line above.

firefox_ui_tests/remote/security/test_safe_browsing_warning_pages.py
```diff
@@ -0,0 +1,116 @@
...
+
+                # Load the test page again and test the ignore warning button
+                self.marionette.navigate(unsafe_page)
+                self.check_ignore_warning_button(unsafe_page)
+
+    def check_get_me_out_of_here_button(self, unsafe_page):
+        get_me_out_of_here_button = self.marionette.find_element(By.ID, "getMeOutButton")
```
whimboo: We are only testing a single button in this method, so the variable can simply be `button`. That makes the line way shorter.

firefox_ui_tests/remote/security/test_safe_browsing_warning_pages.py
```diff
@@ -0,0 +1,116 @@
...
+
+        # This isn't clickable by the time we get here and needs a delay.
+        time.sleep(1)
```
whimboo: Doesn't Marionette throw an exception when loading such a page? Isn't it an error page?

firefox_ui_tests/remote/security/test_safe_browsing_warning_pages.py
```diff
@@ -0,0 +1,116 @@
...
+        get_me_out_of_here_button.click()
+
+        homepage = self.prefs.get_pref('browser.startup.homepage',
+                                       interface='nsIPrefLocalizedString')
```
whimboo: As said on another PR already this needs a property as best on the browser class, which retrieves the default homepage.

firefox_ui_tests/remote/security/test_safe_browsing_warning_pages.py
```diff
@@ -0,0 +1,116 @@
...
+        self.wait_for_condition(lambda mn: homepage in mn.get_url())
+
+    def check_report_button(self, unsafe_page):
+        report_button = self.marionette.find_element(By.ID, "reportButton")
+
+        # This isn't clickable by the time we get here and needs a delay.
+        time.sleep(1)
```
whimboo: Same as above.

firefox_ui_tests/remote/security/test_safe_browsing_warning_pages.py
```diff
@@ -0,0 +1,116 @@
...
+                # mozmill: utils.formatUrlPref("app.support.baseURL") + "phishing-malware"
+                # the code commented out below produces this:
+                # https://support.mozilla.org/1/firefox/38.0a1/Darwin/en-US/phishing-malware
+                # which fails to match the page we're sent to...
+                # url = self.marionette.execute_script("""
+                #    Cu.import("resource://gre/modules/Services.jsm");
+                #    locale = Services.prefs.getCharPref("general.useragent.locale", "");
+                #    return (Services.urlFormatter.formatURLPref("app.support.baseURL")
+                #                                               + "phishing-malware");
+                #    """)
```
whimboo: This is because it's a redirect and we do not know the exact final location for it. Here we should be safe in assume that it will be support.mozilla.org, but it will not be the case in general. With your changes we do no longer check that the correct page on SUMO has been opened.

firefox_ui_tests/remote/security/test_safe_browsing_warning_pages.py
```diff
@@ -0,0 +1,116 @@
...
+                url = 'https://support.mozilla.org'
+
+            else:
+                url = self.marionette.execute_script("""
+                   Cu.import("resource://gre/modules/Services.jsm");
+                   locale = Services.prefs.getCharPref("general.useragent.locale", "");
+                   return (Services.urlFormatter.formatURLPref(
+                            "browser.safebrowsing.malware.reportURL") + arguments[0]);
```
whimboo: I'm sure this is also a redirect, so we can fail with the `wait_for_condition` call below.

firefox_ui_tests/remote/security/test_safe_browsing_warning_pages.py
```diff
@@ -0,0 +1,116 @@
...
+        # This isn't clickable by the time we get here and needs a delay.
+        time.sleep(1)
+        ignore_warning_button.click()
+
+        def find_main_feature_el(mn):
+            try:
+                mn.find_element("id", "main-feature")
+                return True
+            except:
+                return False
```
whimboo: I wonder if there is a better way to wait until an element is available.

firefox_ui_tests/remote/security/test_safe_browsing_warning_pages.py
```diff
@@ -0,0 +1,116 @@
...
+    def check_ignore_warning_button(self, unsafe_page):
+        ignore_warning_button = self.marionette.find_element(By.ID, "ignoreWarningButton")
+
+        # This isn't clickable by the time we get here and needs a delay.
+        time.sleep(1)
+        ignore_warning_button.click()
+
+        def find_main_feature_el(mn):
+            try:
+                mn.find_element("id", "main-feature")
```
whimboo: By.ID

firefox_ui_tests/remote/security/test_safe_browsing_warning_pages.py
```diff
@@ -0,0 +1,116 @@
...
+                return True
+            except:
+                return False
+
+        self.wait_for_condition(find_main_feature_el)
+        self.assertRaises(NoSuchElementException, self.marionette.find_element,
+                          'id', 'ignoreWarningButton')
```
whimboo: By.ID

firefox_ui_tests/remote/security/test_safe_browsing_warning_pages.py
```diff
@@ -0,0 +1,116 @@
...
+        self.assertEquals(self.marionette.get_url(), unsafe_page)
```
whimboo: Can we be sure there is no redirect? Please test with other locales of Firefox here.

firefox_ui_tests/remote/security/test_safe_browsing_warning_pages.py
```diff
@@ -0,0 +1,116 @@
...
+        self.perms.remove('www.itisatrap.org', 'safe-browsing')
```
whimboo: This is tearDown logic, so you can remove it from the test.

firefox_ui_tests/remote/security/test_safe_browsing_warning_pages.py
chmanchester: The entries are added by SafeBrowsing.init, which is called by a setTimeout in browser code:

https://hg.mozilla.org/mozilla-central/annotate/b74938ef94bc/browser/base/content/browser.js#l1194

galgeek: We set both to false in the harness's default_prefs.py

firefox_puppeteer/__init__.py
galgeek: Hi, @whimboo!

I wonder whether we want to maintain a small permissions library, or should we add this to the utils library created in PR #100?

For whatever it's worth here, here's another utils.js function that we need for this PR, utils.assertLoadedUrlEqual, which enables us to check that URLs are equal after a redirect.

firefox_ui_tests/remote/security/test_safe_browsing_warning_pages.py
galgeek: The safe browsing tests load about:blocked, not about:error, and there's no exception thrown.

firefox_ui_tests/remote/security/test_safe_browsing_warning_pages.py
galgeek: Replaced with self.wait_for_condition(expected.element_present(By.ID, 'main-feature'))

firefox_ui_tests/remote/security/test_safe_browsing_warning_pages.py
galgeek: No, it needs to run between the Phishing URL test and the Malware URL test.

It fits better in the main test function. I'll add it there and remote the tearDown.

firefox_ui_tests/remote/security/test_safe_browsing_warning_pages.py
whimboo: Oh! So the init method itself gets called with a 2s timeout. Adding the moz entries will then happen in 2s+Xms here:
https://dxr.mozilla.org/mozilla-central/source/toolkit/components/url-classifier/SafeBrowsing.jsm?from=SafeBrowsing.jsm#199

Given that there is no observer we could listen for we should at least use 3s as sleep here. With 2s only we could run into race conditions.

firefox_puppeteer/__init__.py
whimboo: Lets keep it in utils.js then for now.

firefox_ui_tests/remote/security/test_safe_browsing_warning_pages.py
whimboo: I see. Then please update the comment so it applies to what we see for about:error pages.

firefox_puppeteer/api/permissions.py
galgeek: We don't seem to be in chrome context when this function gets run. It fails without the line setting context, like so:

avascriptException: ReferenceError: Cu is not defined
stacktrace:
	execute_script @utils.py, line 21
	inline javascript, line 1
	src: "          Cu.import("resource://gre/modules/Services.jsm");"

chmanchester: Libraries should use with_context so they don't make an assumption about the caller's context.

whimboo: This is not done by every library. I thought we wanted to have clear context switches in the test itself. I feel we should have a chat about that to find a proper solution, so we can file a mentored bug for refactoring.

chmanchester: Bug 1088905 has the relevant discussion, cases like these are why we added the context manager. We should add this as we find conflicts. It's going to make more sense for functions like these that do relatively isolated things. For more involved interactions with chrome like the tabs library I don't think we'll find too many callers in content scope, anyway.

firefox_ui_tests/remote/security/test_safe_browsing_warning_pages.py
galgeek: Actually, it fits best in the ignore warning check, since that's where the permission gets set in the first place, so I've left it here.

firefox_puppeteer/api/utils.py
```diff
@@ -10,6 +10,23 @@
...
+        https://dxr.mozilla.org/mozilla-central/source/browser/modules/SitePermissions.jsm#144
+        and elsewhere.
+
+        :param host: The web host whose permission will be removed.
+        :param permission: The type of permission to be removed.
+        """
+
+        with self.marionette.using_context('chrome'):
+            self.marionette.execute_script("""
+              Cu.import("resource://gre/modules/Services.jsm");
```
whimboo: Please don't use the shortcut for `Components.utils`, given that it will fail if you call this method if not a brower window is currently selected.

firefox_ui_tests/remote/security/manifest.ini
```diff
@@ -1,3 +1,4 @@
 [test_security_notification.py]
 [test_unknown_issuer.py]
 [test_ssl_disabled_error_page.py]
+[test_safe_browsing_warning_pages.py]
```
whimboo: It still needs sorting.

firefox_ui_tests/remote/security/test_safe_browsing_warning_pages.py
```diff
@@ -0,0 +1,101 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import time
+
+from marionette_driver import (
+    By,
+    expected
+)
```
whimboo: Lets use a single line as long as we do not exceed the line.

firefox_ui_tests/remote/security/test_safe_browsing_warning_pages.py
```diff
@@ -0,0 +1,101 @@
...
+        # Wait for the DOM to receive events
+        time.sleep(1)
+        button.click()
+
+        self.wait_for_condition(expected.element_present(By.ID, 'main-feature'))
+        self.wait_for_condition(expected.element_stale(button))
+        self.assertEquals(self.marionette.get_url(), self.browser.get_loaded_url(unsafe_page))
+
+        # Clean up by removing safe browsing permission for unsafe page
+        self.utils.remove_perms('www.itisatrap.org', 'safe-browsing')
```
whimboo: You have to ensure to remove the set permissions in teardown! If the assert before would fail, you do not remove it here.

firefox_ui_tests/remote/security/test_safe_browsing_warning_pages.py
```diff
@@ -0,0 +1,101 @@
...
+    def check_report_button(self, unsafe_page):
+        button = self.marionette.find_element(By.ID, "reportButton")
+
+        # Wait for the DOM to receive events
+        time.sleep(1)
+        button.click()
+
+        # Wait for page load so we can record and verify the url
+        self.wait_for_condition(lambda mn: mn.execute_script(
+            "return document.readyState == 'complete'"))
```
whimboo: This causes a race-condition if the page load is a bit delayed. So it would return immediately. Wait for an element of the target page for now instead.

firefox_ui_tests/remote/security/test_safe_browsing_warning_pages.py
```diff
@@ -0,0 +1,101 @@
...
+        report_url = self.marionette.get_url()
+
+        with self.marionette.using_context('chrome'):
+            if 'its-a-trap' in unsafe_page:
+                url = self.marionette.execute_script("""
+                   Cu.import("resource://gre/modules/Services.jsm");
+                   return (Services.urlFormatter.formatURLPref("app.support.baseURL")
+                                                              + "phishing-malware");
```
whimboo: There is no need for the outer brackets. Same for the execute_script call below.

firefox_ui_tests/remote/security/test_safe_browsing_warning_pages.py
```diff
@@ -0,0 +1,101 @@
...
+        button = self.marionette.find_element(By.ID, "reportButton")
+
+        # Wait for the DOM to receive events
+        time.sleep(1)
+        button.click()
+
+        # Wait for page load so we can record and verify the url
+        self.wait_for_condition(lambda mn: mn.execute_script(
+            "return document.readyState == 'complete'"))
+        report_url = self.marionette.get_url()
```
whimboo: I don't see the need for this variable.

whimboo: Well, ok we need it, but please move it down to line 86.

firefox_ui_tests/remote/security/test_safe_browsing_warning_pages.py
```diff
@@ -0,0 +1,101 @@
...
+
+                # Load the test page again and test the ignore warning button
+                self.marionette.navigate(unsafe_page)
+                self.check_ignore_warning_button(unsafe_page)
+
+    def check_get_me_out_of_here_button(self, unsafe_page):
+        button = self.marionette.find_element(By.ID, "getMeOutButton")
+
+        # Wait for the DOM to receive events
+        time.sleep(1)
```
whimboo: Please move the `sleep` calls up to the main test method right after each `navigate()` call. That way we can see that it belongs to the page load.

firefox_ui_tests/remote/security/test_safe_browsing_warning_pages.py
```diff
@@ -0,0 +1,101 @@
...
+        button.click()
+
+        self.wait_for_condition(lambda mn: self.browser.default_homepage in mn.get_url())
+
+    def check_report_button(self, unsafe_page):
+        button = self.marionette.find_element(By.ID, "reportButton")
+
+        # Wait for the DOM to receive events
+        time.sleep(1)
```
whimboo: Same as above.

firefox_ui_tests/remote/security/test_safe_browsing_warning_pages.py
```diff
@@ -0,0 +1,101 @@
...
+                   """, script_args=[unsafe_page])
+
+        # check that report_url matches the loaded url that we expect
+        self.assertEquals(report_url, self.browser.get_loaded_url(url))
+
+    def check_ignore_warning_button(self, unsafe_page):
+        button = self.marionette.find_element(By.ID, 'ignoreWarningButton')
+
+        # Wait for the DOM to receive events
+        time.sleep(1)
```
whimboo: Same here.

firefox_ui_tests/remote/security/test_safe_browsing_warning_pages.py
whimboo: Or maybe wait for an existing element become stale. That might work better because we know the source page but not the final redirected web page.

whimboo: And if the element is stale, the above wait condition should work.

firefox_ui_tests/remote/security/test_safe_browsing_warning_pages.py
```diff
@@ -0,0 +1,101 @@
...
+        button.click()
+
+        self.wait_for_condition(expected.element_present(By.ID, 'main-feature'))
+        self.wait_for_condition(expected.element_stale(button))
```
whimboo: Shouldn't we wait for the target page being loaded, similar as above?

galgeek:
Hi, @whimboo!

Thanks for your feedback. I've updated my code. Assuming we'll merge the safe browsing notification bar PR soon, I'll wait to rebase before I update this PR.

firefox_ui_tests/remote/security/test_safe_browsing_warning_pages.py
galgeek: The ignore warning button doesn't seem to do much, at least, with current Nightly, apart from activating the safe browsing notification bar. Waiting for the "main-feature" element is the easy substitute for detecting the notification bar itself.

firefox_ui_tests/remote/security/test_safe_browsing_warning_pages.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def check_report_button(self, unsafe_page):
+        button = self.marionette.find_element(By.ID, "reportButton")
+        button.click()
+
+        # Wait for the button to become stale, then wait for page load
+        # so we can verify the url
+        self.wait_for_condition(expected.element_stale(button))
+        self.wait_for_condition(lambda mn: mn.execute_script(
+            "return document.readyState == 'complete'"))
```
whimboo: We don't have to wait for the `complete` state. `marionette.navigates` already returns for `interactive` AFAIK. It means we have the DOM present but don't have to wait for JS and CSS to be loaded.

Another option would be to wait for an element of the target page but I don't think that's a good idea because we don't have that page under our control. 

So please add a TODO item here for a proper waitForPageLoad method. Lets file a bug if not done yet, for the reference.

firefox_ui_tests/remote/security/test_safe_browsing_warning_pages.py
```diff
@@ -0,0 +1,105 @@
...
+    def check_report_button(self, unsafe_page):
+        button = self.marionette.find_element(By.ID, "reportButton")
+        button.click()
+
+        # Wait for the button to become stale, then wait for page load
+        # so we can verify the url
+        self.wait_for_condition(expected.element_stale(button))
+        # TODO: replace with waitforPageLoad replacement per Bug 1140470
+        self.wait_for_condition(lambda mn: mn.execute_script(
+            "return document.readyState == 'complete'"))
```
whimboo: This is still `complete` and not `interactive` as mentioned in my last review. I don't see why that change won't work.

For the comment please use `Bug 1140470: %description%`.

firefox_ui_tests/remote/security/test_safe_browsing_warning_pages.py
```diff
@@ -0,0 +1,105 @@
...
+
+        # Get the base URL to check; this will result in a redirect.
+        with self.marionette.using_context('chrome'):
+            if 'its-a-trap' in unsafe_page:
+                url = self.marionette.execute_script("""
+                   Components.utils.import("resource://gre/modules/Services.jsm");
+                   return Services.urlFormatter.formatURLPref("app.support.baseURL")
+                                                              + "phishing-malware";
+                   """)
```
whimboo: nit: The triple quotes should be aligned with `u` from line 81. Same for the else block.

firefox_ui_tests/remote/security/test_safe_browsing_warning_pages.py
```diff
@@ -0,0 +1,103 @@
...
+        self.wait_for_condition(expected.element_stale(button))
+        # TODO: Bug 1140470: use replacement for mozmill's waitforPageLoad
+        self.wait_for_condition(lambda mn: mn.execute_script(
+            "return document.readyState == 'complete'"))
+
+        # Get the base URL to check; this will result in a redirect.
+        with self.marionette.using_context('chrome'):
+            if 'its-a-trap' in unsafe_page:
+                url = self.marionette.execute_script("""
+                    Components.utils.import("resource://gre/modules/Services.jsm");
```
whimboo: As usual we are still using 2 chars of indentation for JS code and not 4.

firefox_ui_tests/remote/security/test_safe_browsing_warning_pages.py
```diff
@@ -0,0 +1,103 @@
...
+    def check_report_button(self, unsafe_page):
+        button = self.marionette.find_element(By.ID, "reportButton")
+        button.click()
+
+        # Wait for the button to become stale, then wait for page load
+        # so we can verify the url
+        self.wait_for_condition(expected.element_stale(button))
+        # TODO: Bug 1140470: use replacement for mozmill's waitforPageLoad
+        self.wait_for_condition(lambda mn: mn.execute_script(
+            "return document.readyState == 'complete'"))
```
whimboo: And this is still `complete` again. Please change that, I mentioned it already twice.

nit: please also use triple-quotes.

galgeek:
Thanks again for your feedback! I've updated the PR.

Testing using `document.readystate == 'intereactive'`, instead of `'complete'`,  fails in the second "malware" test run. (I started a comment earlier to this effect that looks like it didn't get entered, sorry!)

whimboo:
Ok, so it shouldn't be a blocker. Waiting a single time for a `complete` page load shouldn't harm us too much, given that we already save lots of time with `navigate`. I will land it soon.

whimboo:
Merged as https://github.com/mozilla/firefox-ui-tests/commit/e0f8f91ff1899aebf52b062da36659e3233d7270

PR #94, opened 2015-02-19
Temporary fix for failing test_base_window_switch_to_and_focus

galgeek:
Sets focusmanager.testmode False for now (since the pref got added to Marionette)


chmanchester:
This looks like what's needed. Can you file a bug for this change?

PR #96, opened 2015-02-20
Bug 1132709 - Convert Mozmill testUnknownIssuer

galgeek:
https://bugzilla.mozilla.org/show_bug.cgi?id=1132709

### DIFF
--- a/firefox_ui_tests/remote/security/test_unknown_issuer.py
+++ b/firefox_ui_tests/remote/security/test_unknown_issuer.py
@@ -1,0 +1,39 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import time
+
+from marionette import By
+from marionette.errors import MarionetteException
+
+from firefox_ui_harness.testcase import FirefoxTestCase
+
+
+class TestUnknownIssuer(FirefoxTestCase):
+
+    def setUp(self):
+        FirefoxTestCase.setUp(self)
+
+        self.url = 'https://ssl-unknownissuer.mozqa.com'
+
+    def test_unknown_issuer(self):
+        with self.marionette.using_context('content'):
+            # Go to a site that has a cert with an unknown issuer
+            self.assertRaises(MarionetteException, self.marionette.navigate, self.url)
+
+            # Wait for the DOM to receive events
+            time.sleep(1)
+
+            # Check the link in cert_domain_link
+            link = self.marionette.find_element(By.ID, 'cert_domain_link')
+            self.assertEquals(link.get_attribute('textContent'),
+                              'ssl-selfsigned-unknownissuer.mozqa.com')
+
+            # Verify the "Get Me Out Of Here!" and "Add Exception" buttons appear
+            self.assertIsNotNone(self.marionette.find_element(By.ID, 'getMeOutOfHereButton'))
+            self.assertIsNotNone(self.marionette.find_element(By.ID, 'exceptionDialogButton'))
+
+            # Verify the error code is correct
+            text = self.marionette.find_element(By.ID, 'technicalContentText')
+            self.assertIn('sec_error_unknown_issuer', text.get_attribute('textContent'))
--- a/firefox_ui_tests/functional/manifest.ini
+++ b/firefox_ui_tests/functional/manifest.ini
@@ -1,6 +1,5 @@
 [include:keyboard_shortcuts/manifest.ini]
 [include:locationbar/manifest.ini]
 [include:private_browsing/manifest.ini]
-[include:security/manifest.ini]
 [include:tabbed_browsing/manifest.ini]
 [include:toolbar/manifest.ini]
--- a/firefox_ui_tests/remote/security/manifest.ini
+++ b/firefox_ui_tests/remote/security/manifest.ini
@@ -1,1 +1,2 @@
 [test_security_notification.py]
+[test_unknown_issuer.py]


firefox_ui_tests/remote/security/test_unknown_issuer.py
```diff
@@ -0,0 +1,47 @@
...
+
+from firefox_ui_harness.testcase import FirefoxTestCase
+
+
+class TestUnknownIssuer(FirefoxTestCase):
+
+    def setUp(self):
+        FirefoxTestCase.setUp(self)
+
+        self.test_data = 'https://ssl-unknownissuer.mozqa.com'
```
whimboo: To be consistent with currently landed tests for Marionette please name this `self.url`.

firefox_ui_tests/remote/security/test_unknown_issuer.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def tearDown(self):
+        try:
+            self.windows.close_all([self.browser])
```
whimboo: This test is not opening new windows. So there is no need to call this method.

firefox_ui_tests/remote/security/test_unknown_issuer.py
```diff
@@ -0,0 +1,47 @@
...
+            FirefoxTestCase.tearDown(self)
+
+    def test_unknown_issuer(self):
+        self.marionette.set_context('content')
+
+        # Go to a site that has a cert with an unknown issuer
+        self.assertRaises(MarionetteException, self.marionette.navigate, self.test_data)
+
+        # Wait for about:error page
+        time.sleep(1)
```
whimboo: This comment should be `Wait for the DOM to receive events`. Maybe @chmanchester could also have a look at this given that we might want to have this in Marionette proper?

firefox_ui_tests/remote/security/test_unknown_issuer.py
```diff
@@ -0,0 +1,47 @@
...
+
+        # Check the issuer in cert_domain_link
+        issuer = (self.marionette.find_element(By.ID, 'cert_domain_link')).get_attribute(
+            'textContent')
+        self.assertEquals(issuer, 'ssl-selfsigned-unknownissuer.mozqa.com')
```
whimboo: Please separate the retrieval of elements and working with those. Here I would use the `get_attribute` call inside `assertEquals`.

firefox_ui_tests/remote/security/test_unknown_issuer.py
```diff
@@ -0,0 +1,47 @@
...
+
+        # Verify the "Get Me Out Of Here!" and "Add Exception" buttons appear
+        self.assertIsNotNone(self.marionette.find_element(By.ID, 'getMeOutOfHereButton'))
+        self.assertIsNotNone(self.marionette.find_element(By.ID, 'exceptionDialogButton'))
+
+        # Verify the error code is correct
+        text = (self.marionette.find_element(By.ID, 'technicalContentText')).get_attribute(
+            'textContent')
+        self.assertTrue('sec_error_unknown_issuer' in text)
```
whimboo: Same as mentioned above. Also there is `assertIn`, which you can use here.

firefox_ui_tests/remote/security/test_unknown_issuer.py
```diff
@@ -0,0 +1,47 @@
...
+        self.test_data = 'https://ssl-unknownissuer.mozqa.com'
+
+    def tearDown(self):
+        try:
+            self.windows.close_all([self.browser])
+        finally:
+            FirefoxTestCase.tearDown(self)
+
+    def test_unknown_issuer(self):
+        self.marionette.set_context('content')
```
whimboo: Personally I would do that in setUp if the whole test works in content scope. Reason is that it is part of the setup logic.

firefox_ui_tests/remote/security/test_unknown_issuer.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+class TestUnknownIssuer(FirefoxTestCase):
+
+    def setUp(self):
+        FirefoxTestCase.setUp(self)
+
+        self.url = 'https://ssl-unknownissuer.mozqa.com'
+
+        self.marionette.set_context('content')
```
whimboo: I'm a bit worried here. Do we ever reset the context to `chrome`? Right now we should fail in tearDown if something is broken here. I feel you should better use the `with` expression in the test method.

firefox_ui_tests/remote/security/test_unknown_issuer.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def test_unknown_issuer(self):
+        # Go to a site that has a cert with an unknown issuer
+        self.assertRaises(MarionetteException, self.marionette.navigate, self.url)
+
+        # Wait for the DOM to receive events
+        time.sleep(1)
+
+        # Check the issuer in cert_domain_link
+        issuer = self.marionette.find_element(By.ID, 'cert_domain_link')
```
whimboo: It's the domain and not the issuer.

firefox_ui_tests/remote/security/test_unknown_issuer.py
```diff
@@ -0,0 +1,39 @@
...
+            time.sleep(1)
+
+            # Check the link in cert_domain_link
+            link = self.marionette.find_element(By.ID, 'cert_domain_link')
+            self.assertEquals(link.get_attribute('textContent'),
+                              'ssl-selfsigned-unknownissuer.mozqa.com')
+
+            # Verify the "Get Me Out Of Here!" and "Add Exception" buttons appear
+            self.assertIsNotNone(self.marionette.find_element(By.ID, 'getMeOutOfHereButton'))
+            self.assertIsNotNone(self.marionette.find_element(By.ID, 'exceptionDialogButton'))
```
whimboo: I think it may be better to check with `is_displayed` but that's more an enhancement, which we even didn't do in the Mozmill test. You can file a followup bug as a good first bug.

whimboo:
Please rebase this PR to master and squash all the commits. Once done I can get it merged.

galgeek:
Hi, @whimboo!

This PR is ready to merge.

Followup bug filed here: https://bugzilla.mozilla.org/show_bug.cgi?id=1136237

PR #100, opened 2015-02-21
Bug 1132687 - Convert Mozmill testSSLDisabledErrorPage

galgeek:
https://bugzilla.mozilla.org/show_bug.cgi?id=1132687

I think that this first commit may be pretty close, but it's certainly WIP, as the expected error page appears to fail to load on my system, while the mozmill version of the test runs fine. I'm curious to see what Travis discovers.

### DIFF
--- a/firefox_puppeteer/__init__.py
+++ b/firefox_puppeteer/__init__.py
@@ -54,6 +54,13 @@ class Puppeteer(object):
         See the :class:`~api.places.Places` reference.
         """
 
+    @use_class_as_property('api.utils.Utils')
+    def utils(self):
+        """Provides an api for interacting with utility actions.
+
+        See the :class:`~api.utils.Utils` reference.
+        """
+
     @property
     def platform(self):
         """Returns the lowercased platform name.
--- a/firefox_puppeteer/api/utils.py
+++ b/firefox_puppeteer/api/utils.py
@@ -1,0 +1,76 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from marionette.errors import TimeoutException
+
+from ..base import BaseLib
+
+
+class Utils(BaseLib):
+    """Low-level access to utility actions."""
+
+    def sanitize(self, data_type):
+        """Sanitize user data, including cache, cookies, offlineApps, history, formdata,
+        downloads, passwords, sessions, siteSettings.
+
+        Usage:
+        sanitize():  Clears all user data.
+        sanitize({ "sessions": True }): Clears only session user data.
+
+        more: https://dxr.mozilla.org/mozilla-central/source/browser/base/content/sanitize.js
+
+        :param data_type: optional, Information specifying data to be sanitized
+        """
+
+        with self.marionette.using_context('chrome'):
+            result = self.marionette.execute_async_script("""
+              Cu.import("resource://gre/modules/Services.jsm");
+
+              var data_type = arguments[0];
+
+              var data_type = (typeof data_type === "undefined") ? {} : {
+                cache: data_type.cache || false,
+                cookies: data_type.cookies || false,
+                downloads: data_type.downloads || false,
+                formdata: data_type.formdata || false,
+                history: data_type.history || false,
+                offlineApps: data_type.offlineApps || false,
+                passwords: data_type.passwords || false,
+                sessions: data_type.sessions || false,
+                siteSettings: data_type.siteSettings || false
+              };
+
+              // Load the sanitize script
+              var tempScope = {};
+              Cc["@mozilla.org/moz/jssubscript-loader;1"]
+              .getService(Ci.mozIJSSubScriptLoader)
+              .loadSubScript("chrome://browser/content/sanitize.js", tempScope);
+
+              // Instantiate the Sanitizer
+              var s = new tempScope.Sanitizer();
+              s.prefDomain = "privacy.cpd.";
+              var itemPrefs = Services.prefs.getBranch(s.prefDomain);
+
+              // Apply options for what to sanitize
+              for (var pref in data_type) {
+                itemPrefs.setBoolPref(pref, data_type[pref]);
+              };
+
+              // Sanitize and wait for the promise to resolve
+              var finished = false;
+              s.sanitize().then(() => {
+                for (let pref in data_type) {
+                  itemPrefs.clearUserPref(pref);
+                };
+                marionetteScriptFinished(true);
+              }, aError => {
+                for (let pref in data_type) {
+                  itemPrefs.clearUserPref(pref);
+                };
+                marionetteScriptFinished(false);
+              });
+            """, script_args=[data_type])
+
+            if not result:
+                raise MarionetteException('Sanitizing of profile data failed.')
--- a/firefox_puppeteer/api/utils.py
+++ b/firefox_puppeteer/api/utils.py
@@ -2,7 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
-from marionette.errors import TimeoutException
+from marionette_driver.errors import MarionetteException
 
 from ..base import BaseLib
 
--- a/firefox_ui_tests/remote/security/test_ssl_disabled_error_page.py
+++ b/firefox_ui_tests/remote/security/test_ssl_disabled_error_page.py
@@ -4,8 +4,8 @@
 
 import time
 
-from marionette import By
-from marionette.errors import MarionetteException
+from marionette_driver import By
+from marionette_driver.errors import MarionetteException
 
 from firefox_ui_harness.testcase import FirefoxTestCase
 
--- a/firefox_ui_tests/remote/security/test_ssl_disabled_error_page.py
+++ b/firefox_ui_tests/remote/security/test_ssl_disabled_error_page.py
@@ -1,0 +1,49 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import time
+
+from marionette import By
+from marionette.errors import MarionetteException
+
+from firefox_ui_harness.testcase import FirefoxTestCase
+
+
+class TestSSLDisabledErrorPage(FirefoxTestCase):
+
+    def setUp(self):
+        FirefoxTestCase.setUp(self)
+
+        self.url = 'https://tlsv1-0.mozqa.com'
+
+        self.utils.sanitize({"sessions": True})
+
+        # Disable SSL 3.0, TLS 1.0 and TLS 1.1 for secure connections
+        # by forcing the use of TLS 1.2
+        # see: http://kb.mozillazine.org/Security.tls.version.*#Possible_values_and_their_effects
+        self.prefs.set_pref('security.tls.version.min', 3)
+        self.prefs.set_pref('security.tls.version.max', 3)
+
+    def test_ssl_disabled_error_page(self):
+        with self.marionette.using_context('content'):
+            # Open the test page
+            self.assertRaises(MarionetteException, self.marionette.navigate, self.url)
+
+            # Wait for the DOM to receive events
+            time.sleep(1)
+
+            # Verify "Secure Connection Failed" error page title
+            title = self.marionette.find_element(By.ID, 'errorTitleText')
+            nss_failure2title = self.browser.get_entity('nssFailure2.title')
+            self.assertEquals(title.get_attribute('textContent'), nss_failure2title)
+
+            # Verify "Try Again" button appears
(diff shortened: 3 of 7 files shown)


firefox_ui_tests/manifest.ini
```diff
@@ -1,2 +1,4 @@
 # functional tests
 [include:functional/manifest.ini]
+# remote tests
```
whimboo: nit: please remove both comments

firefox_ui_tests/remote/security/test_ssl_disabled_error_page.py
```diff
@@ -0,0 +1,55 @@
...
+
+class TestSSLDisabledErrorPage(FirefoxTestCase):
+
+    def setUp(self):
+        FirefoxTestCase.setUp(self)
+
+        self.url = 'https://tlsv1-0.mozqa.com'
+
+        tabbar = self.browser.tabbar
+        tabbar.close_all_tabs([tabbar.tabs[0]])
```
whimboo: No need to call that. This test does not rely on opening/closing tabs.

firefox_ui_tests/remote/security/test_ssl_disabled_error_page.py
```diff
@@ -0,0 +1,55 @@
...
+        # by forcing the use of TLS 1.2
+        # see: http://kb.mozillazine.org/Security.tls.version.*#Possible_values_and_their_effects
+        self.prefs.set_pref('security.tls.version.min', 3)
+        self.prefs.set_pref('security.tls.version.max', 3)
+
+        self.marionette.set_context('content')
+
+    def tearDown(self):
+        self.prefs.restore_all_prefs()
+        FirefoxTestCase.tearDown(self)
```
whimboo: try/finally please.

firefox_ui_tests/remote/security/test_ssl_disabled_error_page.py
```diff
@@ -0,0 +1,55 @@
...
+        self.assertRaises(MarionetteException, self.marionette.navigate, self.url)
+
+        # Wait for the DOM to receive events
+        time.sleep(1)
+
+        # Verify "Secure Connection Failed" error page title
+        title = self.marionette.find_element(By.ID, 'errorTitleText')
+
+        nss_failure2title = self.browser.get_entity('oldSecurityProtocol.title')
+        self.assertEquals(title.get_attribute('textContent'), nss_failure2title)
```
whimboo: This has been changed yesterday for Nightly. Please sync with the already updated Mozmill test on bug 1135616.

firefox_ui_tests/remote/security/test_ssl_disabled_error_page.py
```diff
@@ -0,0 +1,55 @@
...
+
+        # Verify "Try Again" button appears
+        self.assertIsNotNone(self.marionette.find_element(By.ID, 'errorTryAgain'))
+
+        # Verify the error message is correct
+        text = self.marionette.find_element(By.ID, 'errorShortDescText')
+        errorcode = self.marionette.find_element(By.ID, 'errorcode')
+        self.assertEquals(errorcode.get_attribute('textContent'), 'ssl_error_unsupported_version')
```
whimboo: The `errorcode` DOM node is no longer available. Please update this part too.

firefox_ui_tests/remote/security/test_ssl_disabled_error_page.py
```diff
@@ -0,0 +1,96 @@
...
+        FirefoxTestCase.setUp(self)
+
+        self.url = 'https://tlsv1-0.mozqa.com'
+
+        # mozmill: utils.sanitize({ sessions: true }); // necessary? test succeeds without this
+        # also: utils.sanitize used by 3 geolocation tests; better to add to library, maybe prefs?
+        # also: https://dxr.mozilla.org/mozilla-central/source/browser/base/content/sanitize.js#417
+        try:
+            self.marionette.execute_async_script("""
+                Cu.import("resource://gre/modules/Services.jsm");
```
whimboo: This code block clearly has to be moved to a library. I don't know which one so maybe we also put it into a utils.py module under api?

firefox_ui_tests/remote/security/test_ssl_disabled_error_page.py
```diff
@@ -0,0 +1,96 @@
...
+
+                var spec = { sessions: true };
```
whimboo: Keep in mind that this needs to be configurable from the outside, so it does not only cover sessions.

firefox_ui_tests/remote/security/test_ssl_disabled_error_page.py
```diff
@@ -0,0 +1,96 @@
...
+
+        # Disable SSL 3.0, TLS 1.0 and TLS 1.1 for secure connections
+        # by forcing the use of TLS 1.2
+        # see: http://kb.mozillazine.org/Security.tls.version.*#Possible_values_and_their_effects
+        self.prefs.set_pref('security.tls.version.min', 3)
+        self.prefs.set_pref('security.tls.version.max', 3)
+
+    def tearDown(self):
+        try:
+            self.prefs.restore_all_prefs()
```
whimboo: This is called by FirefoxTestCase.tearDown() itself. So no need for this line.

firefox_ui_tests/remote/security/test_ssl_disabled_error_page.py
```diff
@@ -0,0 +1,96 @@
...
+            time.sleep(1)
+
+            # Verify "Secure Connection Failed" error page title
+            title = self.marionette.find_element(By.ID, 'errorTitleText')
+
+            nss_failure2title = self.browser.get_entity('nssFailure2.title')
+            self.assertEquals(title.get_attribute('textContent'), nss_failure2title)
+
+            # Verify "Try Again" button appears
+            self.assertIsNotNone(self.marionette.find_element(By.ID, 'errorTryAgain'))
```
whimboo: Maybe better to check for `is_displayed` here. It should work given that it is content scope, right?

firefox_puppeteer/__init__.py
```diff
@@ -54,6 +54,13 @@ def places(self):
         See the :class:`~api.places.Places` reference.
         """
 
+    @use_class_as_property('api.utils.Utils')
+    def utils(self):
+        """
+        Provides an api for interacting with utility actions.
```
whimboo: The description has to be put into the first line.

firefox_puppeteer/api/utils.py
```diff
@@ -0,0 +1,76 @@
...
+        """Sanitize user data, including cache, cookies, offlineApps, history, formdata,
+        downloads, passwords, sessions, siteSettings.
+
+        Usage:
+        sanitize():  Clears all user data.
+        sanitize({ "sessions": True }): Clears only session user data.
+
+        more: https://dxr.mozilla.org/mozilla-central/source/browser/base/content/sanitize.js
+
+        :param spec: optional, Information specifying data to be sanitized
```
whimboo: `spec` is a really bad naming which we had in the mozmill tests. Lets try to find a better name for the parameter.

firefox_puppeteer/api/utils.py
```diff
@@ -0,0 +1,76 @@
...
+                cache: aSpec.cache || false,
+                cookies: aSpec.cookies || false,
+                downloads: aSpec.downloads || false,
+                formdata: aSpec.formdata || false,
+                history: aSpec.history || false,
+                offlineApps: aSpec.offlineApps || false,
+                passwords: aSpec.passwords || false,
+                sessions: aSpec.sessions || false,
+                siteSettings: aSpec.siteSettings || false
+              };
```
whimboo: Lets do all that on the Python side and limit the JS code as much as possible.

whimboo: Well stitch it! Lets do it later.

firefox_puppeteer/api/utils.py
```diff
@@ -0,0 +1,76 @@
...
+                  itemPrefs.clearUserPref(pref);
+                };
+                marionetteScriptFinished(true);
+              }, aError => {
+                for (let pref in spec) {
+                  itemPrefs.clearUserPref(pref);
+                };
+                marionetteScriptFinished(false);
+              });
+            """, script_args=[spec], script_timeout=10000)
```
whimboo: Please don't specify a timeout here, or is the 5s timeout not enough?

firefox_puppeteer/api/utils.py
```diff
@@ -0,0 +1,76 @@
...
+        except TimeoutException:
+            # in case of a timeout, continue, since this won't necessarily cause test failure
+            pass
```
whimboo: This would cause a test failure if other tests afterward would rely on a cleaned-up profile. If the max time could even exceed the 5s we might want to set it to 30 but not 10000s, which you did here. I think that Marionette takes it as seconds and not milliseconds.

whimboo:
This PR does not contain a unit test for the utils module. Please file a bug so we can get it done by a contributor. @chmanchester, could you help Barbara in landing this PR? Travis seems to be not working correctly at the moment, and I would like to see it passing. Make sure to squash all commits too. Thanks.

galgeek:
Here's the new bug for creating the unit test: 

https://bugzilla.mozilla.org/show_bug.cgi?id=1136902

chmanchester:
@whimboo you need to set r+ on this before I merge it.

firefox_puppeteer/api/utils.py
```diff
@@ -0,0 +1,76 @@
...
+        sanitize():  Clears all user data.
+        sanitize({ "sessions": True }): Clears only session user data.
+
+        more: https://dxr.mozilla.org/mozilla-central/source/browser/base/content/sanitize.js
+
+        :param data_type: optional, Information specifying data to be sanitized
+        """
+
+        try:
+            self.marionette.execute_async_script("""
```
chmanchester: This should throw when it returns false, right?

galgeek: I'm not sure what's best to handle errors here. I copied the execute_async_script framework from another library and haven't thought very hard about it... This test runs fine on my system for now, individually or with all the others, without this code. I wonder if the Marionette test environment requires it.

chmanchester: I just mean that in the current arrangement, if the promise is rejected the test will not fail. I'm not saying it's critical to the test, but you can fail the test in this case by just checking the return of execute_async_script and throwing an exception if it's false.

whimboo: > I just mean that in the current arrangement, if the promise is rejected
> the test will not fail. I'm not saying it's critical to the test, but
> you can fail the test in this case by just checking the return of
> execute_async_script and throwing an exception if it's false.

That's exactly what I was mentioning in my last review. We have to raise
an exception here! Otherwise tests which require a clean environment
will fail with unknown reasons. So please let an exception be raised here.

firefox_puppeteer/api/utils.py
```diff
@@ -0,0 +1,77 @@
...
+                    };
+                    marionetteScriptFinished(true);
+                  }, aError => {
+                    for (let pref in data_type) {
+                      itemPrefs.clearUserPref(pref);
+                    };
+                    marionetteScriptFinished(false);
+                  });
+                """, script_args=[data_type])
+            except TimeoutException:
```
whimboo: There is no need to catch this exception here. Lets it go through...

firefox_puppeteer/api/utils.py
```diff
@@ -0,0 +1,77 @@
...
+                  s.sanitize().then(() => {
+                    for (let pref in data_type) {
+                      itemPrefs.clearUserPref(pref);
+                    };
+                    marionetteScriptFinished(true);
+                  }, aError => {
+                    for (let pref in data_type) {
+                      itemPrefs.clearUserPref(pref);
+                    };
+                    marionetteScriptFinished(false);
```
whimboo: Check the return code of `execute_async_script` and if it is `False` raise an exception. Not sure if Marionette has already a good fit in it in its own errors.py module.

firefox_puppeteer/api/utils.py
```diff
@@ -0,0 +1,75 @@
...
+                };
+                marionetteScriptFinished(true);
+              }, aError => {
+                for (let pref in data_type) {
+                  itemPrefs.clearUserPref(pref);
+                };
+                marionetteScriptFinished(false);
+              });
+            """, script_args=[data_type])
+            if result == False:
```
whimboo: This is just `if not result`, and please add a blank line above.

galgeek:
Hi, @whimboo!

I've updated the PR per your most recent feedback—thank you!

whimboo:
The current bustage of Travis is due to [bug 1124667](https://bugzilla.mozilla.org/show_bug.cgi?id=1124667). I will test it locally then if all is working fine.

whimboo:
We had to land a breaking change on our repo which causes merge conflicts for this patch. Also I have checked the history of commits and the first commit in here is a commit for another test. You really want to strip that out. If you have problems with merging please push a squashed and rebased update.

whimboo:
PR has been merged as https://github.com/mozilla/firefox-ui-tests/commit/5df52262ceb5d6f53537af06999b56b9f0a54677

PR #103, opened 2015-02-23
Bug 1135874 - Add the ability to persist prefs through a restart to the ...

chmanchester:
...prefs library.r=whimboo

### DIFF
--- a/firefox_puppeteer/api/prefs.py
+++ b/firefox_puppeteer/api/prefs.py
@@ -206,7 +206,9 @@ class Preferences(BaseLib):
     def persist_prefs(self):
         """Persist current pref values to the default prefs file (typically prefs.js)
         """
-        self.marionette.execute_script("""
-          Cu.import("resource://gre/modules/Services.jsm");
-          Services.prefs.savePrefFile(null);
-        """)
+        with self.marionette.using_context('chrome'):
+            self.marionette.execute_script("""
+              Cu = Components.utils;
+              Cu.import("resource://gre/modules/Services.jsm");
+              Services.prefs.savePrefFile(null);
+            """)
--- a/firefox_puppeteer/api/prefs.py
+++ b/firefox_puppeteer/api/prefs.py
@@ -202,3 +202,11 @@ class Preferences(BaseLib):
             """, script_args=[pref_name, value])
 
         assert retval
+
+    def persist_prefs(self):
+        """Persist current pref values to the default prefs file (typically prefs.js)
+        """
+        self.marionette.execute_script("""
+          Cu.import("resource://gre/modules/Services.jsm");
+          Services.prefs.savePrefFile(null);
+        """)
--- a/firefox_puppeteer/tests/test_prefs.py
+++ b/firefox_puppeteer/tests/test_prefs.py
@@ -156,26 +156,60 @@ class testPreferences(FirefoxTestCase):
         self.assertEqual(self.prefs.get_pref(self.string_pref), 'test_string')
 
     def test_persist_prefs_across_restart(self):
+        harness_set_pref = 'app.update.auto'
+
         old_prefs = {
             self.bool_pref: self.prefs.get_pref(self.bool_pref),
             self.int_pref: self.prefs.get_pref(self.int_pref),
+            harness_set_pref: self.prefs.get_pref(harness_set_pref),
         }
 
         new_prefs = {
             self.bool_pref: not old_prefs[self.bool_pref],
             self.int_pref: old_prefs[self.int_pref] + 1,
+            harness_set_pref: not old_prefs[harness_set_pref],
         }
 
         for pref, val in new_prefs.iteritems():
             self.prefs.set_pref(pref, val)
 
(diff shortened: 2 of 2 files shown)


whimboo:
This test is not testing those bits as I have claimed about in the other bug. Mainly because you are not using a preference we set by default with the harness. Every preference you create inside a test will remain after a restart. Therefore you also do not need the extra method you have added to prefs.js, which indeed is a no-op, because Firefox itself will save changes to preferences to prefs.js automatically. If you take e.g. `app.update.auto` and set it to `True` in the test, you will see that it has been reset to `False` after the restart.

chmanchester:
I see. Those are re-set after a restart by marionette's automation, so that will need a different fix. The method I added is certainly not a no-op. The test here fails without it.

firefox_puppeteer/api/prefs.py
```diff
@@ -202,3 +202,13 @@ def set_pref(self, pref_name, value):
...
         assert retval
+
+    def persist_prefs(self):
+        """Persist current pref values to the default prefs file (typically prefs.js)
+        """
+        with self.marionette.using_context('chrome'):
+            self.marionette.execute_script("""
+              Cu = Components.utils;
+              Cu.import("resource://gre/modules/Services.jsm");
+              Services.prefs.savePrefFile(null);
```
whimboo: I still don't understand this. [savePrefFile](https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XPCOM/Reference/Interface/nsIPrefService#savePrefFile%28%29) will store the changes in prefs.js if the parameter as given is null. But that should be done anyhow at latest when Firefox closes. Why are changed preferences not been written out to the prefs.js file when we call `set_pref()`?

firefox_puppeteer/tests/test_prefs.py
```diff
@@ -154,3 +154,62 @@ def test_set_pref_new_values(self):
...
+            harness_set_pref: self.prefs.get_pref(harness_set_pref),
+        }
+
+        new_prefs = {
+            self.bool_pref: not old_prefs[self.bool_pref],
+            self.int_pref: old_prefs[self.int_pref] + 1,
+            harness_set_pref: not old_prefs[harness_set_pref],
+        }
+
+        for pref, val in new_prefs.iteritems():
```
whimboo: Mind using `name` and `value` here?

firefox_puppeteer/tests/test_prefs.py
```diff
@@ -154,3 +154,62 @@ def test_set_pref_new_values(self):
...
+            if pref != harness_set_pref:
+                self.assertEqual(self.prefs.get_pref(pref), val)
+            else:
+                self.assertEqual(self.prefs.get_pref(pref), old_prefs[pref])
+
+        self.prefs.restore_all_prefs()
+
+        # If a pref is passed as an argument to restart, it is always persisted.
+        self.marionette.restart(prefs=new_prefs)
```
whimboo: That's not allowed for me or at least I'm missing necessary changes you did for Marionette. I see the following failure: ´restart() got an unexpected keyword argument 'prefs'´

firefox_puppeteer/api/prefs.py
whimboo: For me this sounds like a bug in Marionette and I don't think we should workaround this. I will quickly test it with Mozmill too.

whimboo: It perfectly works in Mozmill that preferences persist across sessions. Thinking more about that I assume this is a left-over from the restart capability integration into Marionette?

whimboo: OMG!!! This is https://github.com/mozilla/firefox-ui-tests/blob/master/firefox_ui_harness/testcase.py#L35.

So we have two options:

1) Each test takes care of removing the prefs as set, which we most likely wanna do. I actually never thought about this problem.
2) We define an exclusion flag for tearDown.

firefox_puppeteer/tests/test_prefs.py
whimboo: So in case of restarts Marionette should not reset the modified prefs, but should do the action as you specify here with the prefs parameter automatically. This is clearly not Firefox UI test only, but would affect everyone. Please explain why a restart of the application should reset all the prefs.

firefox_puppeteer/api/prefs.py
whimboo: Hm, it doesn't look like that this is really the problem here. `tearDown` is really only called at the very end of the test. So I don't know why Marionette resets those preferences too.

chmanchester: My hypothesis is that marionette kills the browser in a way that doesn't hit the pref persisting code. I'll try to figure out what's going on.

whimboo: That's what actually is happening here. As reference here the bug which I filed: https://bugzilla.mozilla.org/show_bug.cgi?id=1137388

chmanchester:
We're not going to take this approach so closing. There is an alternative proposed in bug 1135874

PR #107, opened 2015-02-26
Bug 1133753 - Convert Mozmill about-window.js library to Python/Marionette

bobsilverberg:
This includes the current commits from PR #90 as `AboutWindow` depends on `SoftwareUpdate`. All of the changes for `AboutWindow` are included in a single commit.

### DIFF
--- a/firefox_puppeteer/docs/ui/about_window/window.rst
+++ b/firefox_puppeteer/docs/ui/about_window/window.rst
@@ -14,3 +14,53 @@ Deck
 .. autoclass:: firefox_puppeteer.ui.about_window.deck.Deck
    :members:
    :inherited-members:
+
+ApplyBillboardPanel
+-------------------
+
+.. autoclass:: firefox_puppeteer.ui.about_window.deck.ApplyBillboardPanel
+   :members:
+   :inherited-members:
+
+ApplyPanel
+----------
+
+.. autoclass:: firefox_puppeteer.ui.about_window.deck.ApplyPanel
+   :members:
+   :inherited-members:
+
+CheckForUpdatesPanel
+--------------------
+
+.. autoclass:: firefox_puppeteer.ui.about_window.deck.CheckForUpdatesPanel
+   :members:
+   :inherited-members:
+
+CheckingForUpdatesPanel
+-----------------------
+
+.. autoclass:: firefox_puppeteer.ui.about_window.deck.CheckingForUpdatesPanel
+   :members:
+   :inherited-members:
+
+DownloadAndInstallPanel
+-----------------------
+
+.. autoclass:: firefox_puppeteer.ui.about_window.deck.DownloadAndInstallPanel
+   :members:
+   :inherited-members:
+
+DownloadFailedPanel
+-------------------
+
+.. autoclass:: firefox_puppeteer.ui.about_window.deck.DownloadFailedPanel
+   :members:
+   :inherited-members:
+
+DownloadingPanel
+----------------
+
+.. autoclass:: firefox_puppeteer.ui.about_window.deck.DownloadingPanel
+   :members:
+   :inherited-members:
+
--- a/firefox_puppeteer/docs/ui/about_window/window.rst
+++ b/firefox_puppeteer/docs/ui/about_window/window.rst
@@ -1,0 +1,16 @@
+About Window
+============
+
+AboutWindow
+--------------
+
+.. autoclass:: firefox_puppeteer.ui.about_window.window.AboutWindow
+   :members:
+   :inherited-members:
+
+Deck
+----
+
+.. autoclass:: firefox_puppeteer.ui.about_window.deck.Deck
+   :members:
+   :inherited-members:
--- a/firefox_puppeteer/tests/test_about_window.py
+++ b/firefox_puppeteer/tests/test_about_window.py
@@ -2,7 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-from marionette import By
+from marionette_driver import By
 
 from firefox_ui_harness.testcase import FirefoxTestCase
 
--- a/firefox_puppeteer/tests/test_about_window.py
+++ b/firefox_puppeteer/tests/test_about_window.py
@@ -1,0 +1,63 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from marionette import By
+
+from firefox_ui_harness.testcase import FirefoxTestCase
+
+
+class TestAboutWindow(FirefoxTestCase):
+
+    def setUp(self):
+        FirefoxTestCase.setUp(self)
+
+        self.about_window = self.browser.open_about_window(trigger='menu')
+
+    def tearDown(self):
+        try:
+            self.windows.close_all([self.browser])
+        finally:
+            FirefoxTestCase.tearDown(self)
+
+    def test_basic(self):
+        self.assertEqual(self.about_window.window_type, 'Browser:About')
+
+    def test_check_for_updates_and_download(self):
+        # testing both of these as they need to occur in this order
+        self.about_window.check_for_updates()
+        self.assertEqual('downloadAndInstall', self.about_window.wizard_state)
+        self.about_window.download()
+        self.assertEqual('apply', self.about_window.wizard_state)
+
+    def test_close_window(self):
+        """Test closing the About window."""
+        self.about_window.close()
+        self.assertTrue(self.about_window.closed)
+
+    def test_open_window(self):
+        """Test various opening strategies."""
+        def opener(win):
+            menu = win.marionette.find_element(By.ID, 'aboutName')
+            menu.click()
+
+        open_strategies = ('menu',
+                           opener,
+                           )
+
+        for trigger in open_strategies:
+            if not self.about_window.closed:
+                self.about_window.close()
+            about_window = self.browser.open_about_window(trigger=trigger)
+            self.assertEquals(about_window, self.windows.current)
+            about_window.close()
+
+    def test_patch_info(self):
+        self.assertTrue(self.about_window.patch_info['download_duration'])
+        self.assertTrue(self.about_window.patch_info['channel'])
+
+    def test_updates_found(self):
+        self.assertTrue(self.about_window.updates_found)
+
+    def test_wizard_state(self):
+        self.assertTrue(self.about_window.wizard_state)
--- a/firefox_puppeteer/tests/test_about_window.py
+++ b/firefox_puppeteer/tests/test_about_window.py
@@ -12,7 +12,7 @@ class TestAboutWindow(FirefoxTestCase):
     def setUp(self):
         FirefoxTestCase.setUp(self)
 
-        self.about_window = self.browser.open_about_window(trigger='menu')
+        self.about_window = self.browser.open_about_window()
 
     def tearDown(self):
         try:
@@ -23,17 +23,52 @@ class TestAboutWindow(FirefoxTestCase):
     def test_basic(self):
         self.assertEqual(self.about_window.window_type, 'Browser:About')
 
-    def test_check_for_updates_and_download(self):
-        # testing both of these as they need to occur in this order
-        self.about_window.check_for_updates()
-        self.assertEqual('downloadAndInstall', self.about_window.wizard_state)
-        self.about_window.download()
-        self.assertEqual('apply', self.about_window.wizard_state)
+    def test_elements(self):
+        """Test correct retrieval of elements."""
+        self.assertNotEqual(self.about_window.dtds, [])
 
-    def test_close_window(self):
-        """Test closing the About window."""
-        self.about_window.close()
-        self.assertTrue(self.about_window.closed)
+        self.assertEqual(
+            self.about_window.deck.element.get_attribute('localName'),
+            'deck')
+
+        # apply panel
+        panel = self.about_window.deck.apply
+        self.assertEqual(panel.element.get_attribute('localName'), 'hbox')
+        self.assertEqual(panel.update_button.get_attribute('localName'), 'button')
+
+        # apply_billboard panel
+        panel = self.about_window.deck.apply_billboard
+        self.assertEqual(panel.element.get_attribute('localName'), 'hbox')
+        self.assertEqual(
+            panel.apply_billboard_button.get_attribute('localName'),
+            'button')
+
+        # check_for_updates panel
+        panel = self.about_window.deck.check_for_updates
+        self.assertEqual(panel.element.get_attribute('localName'), 'hbox')
+        self.assertEqual(
+            panel.check_for_updates_button.get_attribute('localName'),
+            'button')
+
+        # checking_for_updates panel
+        self.assertEqual(
+            self.about_window.deck.checking_for_updates.element.get_attribute('localName'),
+            'hbox')
+
+        # download_and_install panel
+        panel = self.about_window.deck.download_and_install
+        self.assertEqual(panel.element.get_attribute('localName'), 'hbox')
+        self.assertEqual(panel.download_button.get_attribute('localName'), 'button')
+
+        # download_failed panel
+        self.assertEqual(
+            self.about_window.deck.download_failed.element.get_attribute('localName'),
+            'hbox')
+
+        # downloading panel
+        self.assertEqual(
+            self.about_window.deck.downloading.element.get_attribute('localName'),
+            'hbox')
 
     def test_open_window(self):
         """Test various opening strategies."""
@@ -45,19 +80,12 @@ class TestAboutWindow(FirefoxTestCase):
                            opener,
                            )
 
+        self.about_window.close()
         for trigger in open_strategies:
-            if not self.about_window.closed:
-                self.about_window.close()
             about_window = self.browser.open_about_window(trigger=trigger)
             self.assertEquals(about_window, self.windows.current)
             about_window.close()
 
     def test_patch_info(self):
-        self.assertTrue(self.about_window.patch_info['download_duration'])
-        self.assertTrue(self.about_window.patch_info['channel'])
-
-    def test_updates_found(self):
-        self.assertTrue(self.about_window.updates_found)
-
-    def test_wizard_state(self):
-        self.assertTrue(self.about_window.wizard_state)
+        self.assertEqual(self.about_window.patch_info['download_duration'], -1)
+        self.assertNotEqual(self.about_window.patch_info['channel'], None)
--- a/firefox_puppeteer/tests/test_about_window.py
+++ b/firefox_puppeteer/tests/test_about_window.py
@@ -34,21 +34,17 @@ class TestAboutWindow(FirefoxTestCase):
         # apply panel
         panel = self.about_window.deck.apply
         self.assertEqual(panel.element.get_attribute('localName'), 'hbox')
-        self.assertEqual(panel.update_button.get_attribute('localName'), 'button')
+        self.assertEqual(panel.button.get_attribute('localName'), 'button')
 
         # apply_billboard panel
         panel = self.about_window.deck.apply_billboard
         self.assertEqual(panel.element.get_attribute('localName'), 'hbox')
-        self.assertEqual(
-            panel.apply_billboard_button.get_attribute('localName'),
-            'button')
+        self.assertEqual(panel.button.get_attribute('localName'), 'button')
 
         # check_for_updates panel
         panel = self.about_window.deck.check_for_updates
         self.assertEqual(panel.element.get_attribute('localName'), 'hbox')
-        self.assertEqual(
-            panel.check_for_updates_button.get_attribute('localName'),
-            'button')
+        self.assertEqual(panel.button.get_attribute('localName'), 'button')
 
         # checking_for_updates panel
         self.assertEqual(
@@ -58,17 +54,15 @@ class TestAboutWindow(FirefoxTestCase):
         # download_and_install panel
         panel = self.about_window.deck.download_and_install
         self.assertEqual(panel.element.get_attribute('localName'), 'hbox')
-        self.assertEqual(panel.download_button.get_attribute('localName'), 'button')
+        self.assertEqual(panel.button.get_attribute('localName'), 'button')
 
         # download_failed panel
-        self.assertEqual(
-            self.about_window.deck.download_failed.element.get_attribute('localName'),
-            'hbox')
+        self.assertEqual(self.about_window.deck.download_failed.element.get_attribute('localName'),
+                         'hbox')
 
         # downloading panel
-        self.assertEqual(
-            self.about_window.deck.downloading.element.get_attribute('localName'),
-            'hbox')
+        self.assertEqual(self.about_window.deck.downloading.element.get_attribute('localName'),
+                         'hbox')
 
     def test_open_window(self):
         """Test various opening strategies."""
@@ -88,4 +82,4 @@ class TestAboutWindow(FirefoxTestCase):
 
     def test_patch_info(self):
         self.assertEqual(self.about_window.patch_info['download_duration'], -1)
-        self.assertNotEqual(self.about_window.patch_info['channel'], None)
+        self.assertTrue(self.about_window.patch_info['channel'])
--- a/firefox_puppeteer/ui/about_window.py
+++ b/firefox_puppeteer/ui/about_window.py
@@ -1,0 +1,130 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from datetime import datetime
+
+from marionette import By, Wait
+
+from .windows import BaseWindow
+from ..api.software_update import SoftwareUpdate
+
+
+class AboutWindow(BaseWindow):
+    """Representation of the About window."""
+
+    TIMEOUT_UPDATE_APPLY = 300
+    TIMEOUT_UPDATE_CHECK = 30
+    TIMEOUT_UPDATE_DOWNLOAD = 360
+
+    window_type = 'Browser:About'
+
+    def __init__(self, marionette_getter, window_handle):
+        BaseWindow.__init__(self, marionette_getter, window_handle)
+
+        self.software_update = SoftwareUpdate(marionette_getter)
+        self.download_duration = -1
+
+    @property
+    def patch_info(self):
+        """ Returns information about the active update in the queue.
+
+        :returns: A dictionary with information about the active patch
+        """
+        patch = self.software_update.patch_info
+        patch['download_duration'] = self.download_duration
+        return patch
+
+    @property
+    def updates_found(self):
+        """ Checks if updates were found
+
+        :returns: True if updates were found
+        """
+
+        return self.wizard_state != 'noUpdatesFound'
+
+    @property
+    def wizard_state(self):
+        deck = self.marionette.find_element(By.ID, 'updateDeck')
+        return self.marionette.execute_script("""
+          return arguments[0].selectedPanel.id;
+        """, script_args=[deck])
+
+    def check_for_updates(self):
+        """Clicks on "Check for Updates" button, and waits for check to complete"""
+        Wait(self.marionette).until(
+            lambda _: self.wizard_state == 'checkForUpdates')
+        check_button = self.marionette.find_element(By.ID, 'checkForUpdatesButton')
+        check_button.click()
+
+        # Wait for the update checking to finish
+        Wait(self.marionette, self.TIMEOUT_UPDATE_CHECK).until(
+            lambda _: self.wizard_state != 'checkingForUpdates')
+
+    def download(self, wait_for_finish=True, timeout=TIMEOUT_UPDATE_DOWNLOAD):
+        """ Download the update
+
+        :param wait_for_finish: Should the function wait until the download has finished?
+        :param timeout: How long to wait for the download to finish
+        """
+        assert self.software_update.update_channel.default_channel ==\
+               self.software_update.update_channel.channel,\
+            'The update channel has been set correctly.'
+
+        if self.wizard_state == 'downloadAndInstall':
+            install_button = self.marionette.find_element(By.ID, 'downloadAndInstallButton')
+            install_button.click()
+
+        # Wait for the download to start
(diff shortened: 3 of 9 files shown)


firefox_puppeteer/ui/about_window.py
```diff
@@ -0,0 +1,130 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from datetime import datetime
+
+from marionette import By, Wait
+
+from .windows import BaseWindow
+from ..api.software_update import SoftwareUpdate
```
whimboo: I landed the patch for PageInfoWindow yesterday which adds its own window.py file under a subdirectory. You may want to do the same now.

firefox_puppeteer/ui/about_window.py
```diff
@@ -0,0 +1,130 @@
...
+
+        :returns: A dictionary with information about the active patch
+        """
+        patch = self.software_update.patch_info
+        patch['download_duration'] = self.download_duration
+        return patch
+
+    @property
+    def updates_found(self):
+        """ Checks if updates were found
```
whimboo: nit: missing final dot. Also there is no need to have an extra blank after the triple double-quotes.

firefox_puppeteer/ui/about_window.py
```diff
@@ -0,0 +1,130 @@
...
+
+        :returns: True if updates were found
+        """
+
+        return self.wizard_state != 'noUpdatesFound'
+
+    @property
+    def wizard_state(self):
+        deck = self.marionette.find_element(By.ID, 'updateDeck')
```
whimboo: The deck needs to be defined as a separate class. Please see the equivalent for page info window, which landed yesterday.

firefox_puppeteer/ui/about_window.py
```diff
@@ -0,0 +1,130 @@
...
+        return self.marionette.execute_script("""
+          return arguments[0].selectedPanel.id;
+        """, script_args=[deck])
+
+    def check_for_updates(self):
+        """Clicks on "Check for Updates" button, and waits for check to complete"""
+        Wait(self.marionette).until(
+            lambda _: self.wizard_state == 'checkForUpdates')
+        check_button = self.marionette.find_element(By.ID, 'checkForUpdatesButton')
```
whimboo: This will become a property on the to be created deck class.

firefox_puppeteer/ui/about_window.py
```diff
@@ -0,0 +1,130 @@
...
+        check_button.click()
+
+        # Wait for the update checking to finish
+        Wait(self.marionette, self.TIMEOUT_UPDATE_CHECK).until(
+            lambda _: self.wizard_state != 'checkingForUpdates')
+
+    def download(self, wait_for_finish=True, timeout=TIMEOUT_UPDATE_DOWNLOAD):
+        """ Download the update
+
+        :param wait_for_finish: Should the function wait until the download has finished?
```
whimboo: I would name it: `True, if the function has to wait for the download to be finished`. Also please mark it as optional, and add the default value. The prefs module has some good example for those.

firefox_puppeteer/ui/about_window.py
```diff
@@ -0,0 +1,130 @@
...
+               self.software_update.update_channel.channel,\
+            'The update channel has been set correctly.'
+
+        if self.wizard_state == 'downloadAndInstall':
+            install_button = self.marionette.find_element(By.ID, 'downloadAndInstallButton')
+            install_button.click()
+
+        # Wait for the download to start
+        Wait(self.marionette).until(
+            lambda _: self.wizard_state != 'downloadAndInstall')
```
whimboo: This should be inside the if block.

firefox_puppeteer/ui/about_window.py
```diff
@@ -0,0 +1,130 @@
...
+            #
+            # The current JS code is:
+            #
+            #     var wizard = updateWizard.handleUpdateWizardDialog();
+            #     wizard.waitForWizardPage(updateWizard.WIZARD_PAGES.updatesfoundbasic);
+            #     wizard.download();
+            #     wizard.close();
+            #     this._downloadDuration = wizard._downloadDuration;
+            #
+            #     return;
```
whimboo: We should throw a not-implemented exception to not cause a hang of the test.

firefox_puppeteer/ui/about_window.py
```diff
@@ -0,0 +1,130 @@
...
+
+        if wait_for_finish:
+            start_time = datetime.now()
+            self._wait_for_download_finished(timeout)
+            self.download_duration = datetime.now() - start_time
+
+    def _wait_for_download_finished(self, timeout):
```
whimboo: This should not be a private class. It also needs to be accessible from the update tests.

firefox_puppeteer/ui/about_window.py
```diff
@@ -0,0 +1,130 @@
...
+        """ Waits until the downloaded update has been applied
+
+        :param timeout: How long to wait for the update to apply
+        """
+
+        Wait(self.marionette, timeout).until(
+            lambda _: self.wizard_state == 'apply',
+            message='Final wizard page has been selected.')
+
+        Wait(self.marionette, timeout).until(
```
whimboo: Please leave and improve the comment we had for this code in the Mozmill version of the about window.

firefox_puppeteer/ui/windows.py
```diff
@@ -494,3 +497,26 @@ def callback(win):
                 raise ValueError('Unknown opening method: "%s"' % trigger)
 
         return BaseWindow.open_window(self, callback, BrowserWindow)
+
+    def open_about_window(self, trigger='menu'):
```
whimboo: nit: Please sort alphabetical

whimboo:
Bob, both platforms on Travis failed with your latest commit. I expect there is something wrong.

firefox_puppeteer/tests/test_about_window.py
```diff
@@ -0,0 +1,63 @@
...
+
+from firefox_ui_harness.testcase import FirefoxTestCase
+
+
+class TestAboutWindow(FirefoxTestCase):
+
+    def setUp(self):
+        FirefoxTestCase.setUp(self)
+
+        self.about_window = self.browser.open_about_window(trigger='menu')
```
whimboo: nit: Leave out the parameter so we fallback to the default.

firefox_puppeteer/tests/test_about_window.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def test_basic(self):
+        self.assertEqual(self.about_window.window_type, 'Browser:About')
+
+    def test_check_for_updates_and_download(self):
+        # testing both of these as they need to occur in this order
+        self.about_window.check_for_updates()
+        self.assertEqual('downloadAndInstall', self.about_window.wizard_state)
+        self.about_window.download()
+        self.assertEqual('apply', self.about_window.wizard_state)
```
whimboo: That test won't work and we cannot do it as a unit test. It would require an old version of Firefox. We even should not do an update check, given that in case of an update available it would automatically download it. Or would we have to click the download button? If the latter is the case, lets at least do an update check. Otherwise kill this test.

firefox_puppeteer/tests/test_about_window.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def test_close_window(self):
+        """Test closing the About window."""
+        self.about_window.close()
+        self.assertTrue(self.about_window.closed)
```
whimboo: Given that there is no other specification for closing the window we can get rid of this test. Also you test this already in `test_open_window`.

firefox_puppeteer/tests/test_about_window.py
```diff
@@ -0,0 +1,63 @@
...
+            menu = win.marionette.find_element(By.ID, 'aboutName')
+            menu.click()
+
+        open_strategies = ('menu',
+                           opener,
+                           )
+
+        for trigger in open_strategies:
+            if not self.about_window.closed:
+                self.about_window.close()
```
whimboo: This if condition should never be necessary in this loop. Given that you open a new about window in setUp() you want to call `close()` before the for loop.

firefox_puppeteer/tests/test_about_window.py
```diff
@@ -0,0 +1,63 @@
...
+            about_window = self.browser.open_about_window(trigger=trigger)
+            self.assertEquals(about_window, self.windows.current)
+            about_window.close()
+
+    def test_patch_info(self):
+        self.assertTrue(self.about_window.patch_info['download_duration'])
```
whimboo: The duration is `-1` by default. I assume you want to check for this value.

firefox_puppeteer/tests/test_about_window.py
```diff
@@ -0,0 +1,63 @@
...
+        self.assertTrue(self.about_window.patch_info['channel'])
```
whimboo: That's a string. So maybe check for not `None`?

firefox_puppeteer/tests/test_about_window.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def test_updates_found(self):
+        self.assertTrue(self.about_window.updates_found)
```
whimboo: Without an update check that will be `False`. See my comment above if we can do that even or not.

firefox_puppeteer/tests/test_about_window.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def test_wizard_state(self):
+        self.assertTrue(self.about_window.wizard_state)
```
whimboo: The wizard is a separate class, so make this its own test class in this module.

firefox_puppeteer/ui/about_window/deck.py
```diff
@@ -0,0 +1,33 @@
...
+    @property
+    def wizard_state(self):
+        """ Returns the current state of the update wizard."""
+        return self.marionette.execute_script("""
+          return arguments[0].selectedPanel.id;
+        """, script_args=[self.element])
+
+    def click_check_for_updates_button(self):
+        """Clicks on "Check for Updates" button."""
+        self._click_deck_button('checkForUpdatesButton')
```
whimboo: Each of the buttons we need want their own property which represents the underlying DOM element. So make all those methods to return the HTMLElement and being a property.

firefox_puppeteer/ui/about_window/deck.py
```diff
@@ -0,0 +1,33 @@
...
+    def _click_deck_button(self, element_id):
+        button = self.marionette.find_element(By.ID, element_id)
+        button.click()
+
+    @property
+    def wizard_state(self):
+        """ Returns the current state of the update wizard."""
+        return self.marionette.execute_script("""
+          return arguments[0].selectedPanel.id;
+        """, script_args=[self.element])
```
whimboo: We should also create panels instances for each of the known panels. You can copy and paste most of the code from the page info deck and its panels. Only make sure to not add a `select()` method, given that panels are selected by pressing the buttons and underlying JS code which determines the correct state.

firefox_puppeteer/ui/about_window/window.py
```diff
@@ -0,0 +1,140 @@
...
+from marionette_driver import By, Wait
+
+from ..windows import BaseWindow
+from ...api.software_update import SoftwareUpdate
+from deck import Deck
+
+
+class AboutWindow(BaseWindow):
+    """Representation of the About window."""
+
```
whimboo: Please add the two DTDs which contain necessary labels of elements and which are referenced here:
https://dxr.mozilla.org/mozilla-central/source/browser/base/content/aboutDialog.xul#11

Properties we don't have. But for both add a test as we do for the other window types in the basic test.

firefox_puppeteer/ui/about_window/window.py
```diff
@@ -0,0 +1,140 @@
...
+    TIMEOUT_UPDATE_APPLY = 300
+    TIMEOUT_UPDATE_CHECK = 30
+    TIMEOUT_UPDATE_DOWNLOAD = 360
+
+    window_type = 'Browser:About'
```
whimboo: Please move this to the top as very first class member.

firefox_puppeteer/ui/about_window/window.py
```diff
@@ -0,0 +1,140 @@
...
+from ...api.software_update import SoftwareUpdate
+from deck import Deck
+
+
+class AboutWindow(BaseWindow):
+    """Representation of the About window."""
+
+    TIMEOUT_UPDATE_APPLY = 300
+    TIMEOUT_UPDATE_CHECK = 30
+    TIMEOUT_UPDATE_DOWNLOAD = 360
```
whimboo: I don't know about exact rules for constants but in Python those should be lowercase afaik. At least PEP8 complains about.

firefox_puppeteer/ui/about_window/window.py
```diff
@@ -0,0 +1,140 @@
...
+
+    window_type = 'Browser:About'
+
+    def __init__(self, marionette_getter, window_handle):
+        BaseWindow.__init__(self, marionette_getter, window_handle)
```
whimboo: Use `*args` and `**kwargs` here.

firefox_puppeteer/ui/about_window/window.py
```diff
@@ -0,0 +1,140 @@
...
+
+        self.software_update = SoftwareUpdate(marionette_getter)
```
whimboo: Make it `lambda: self.marionette`

firefox_puppeteer/ui/about_window/window.py
bobsilverberg: That is strange. https://www.python.org/dev/peps/pep-0008/#constants

firefox_puppeteer/ui/about_window/window.py
```diff
@@ -0,0 +1,140 @@
...
+        self.download_duration = -1
```
whimboo: This should be private, right?

firefox_puppeteer/ui/about_window/window.py
```diff
@@ -0,0 +1,140 @@
...
+        patch['download_duration'] = self.download_duration
+        return patch
+
+    @property
+    def updates_found(self):
+        """ Checks if updates were found.
+
+        :returns: True if updates were found
+        """
+        return self.wizard_state != 'noUpdatesFound'
```
whimboo: This method will cause failures if called on different panels of the deck. If you want to return a definitive value it has to go through the software update library.

firefox_puppeteer/ui/about_window/window.py
```diff
@@ -0,0 +1,140 @@
...
+
+    @property
+    def wizard_state(self):
+        """ Returns the current state of the update wizard."""
+        return self.deck.wizard_state
```
whimboo: You can remove this method. It has to be accessed directly via the deck instance.

firefox_puppeteer/ui/about_window/window.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def check_for_updates(self):
+        """Clicks on "Check for Updates" button, and waits for check to complete"""
+        Wait(self.marionette).until(
+            lambda _: self.wizard_state == 'checkForUpdates')
```
whimboo: I don't see why we need this `Wait` call here. Given that we cannot set the panel directly the method has to assume that this panel is selected. If you want you could assert for it.

firefox_puppeteer/ui/about_window/window.py
```diff
@@ -0,0 +1,140 @@
...
+        self.deck.click_check_for_updates_button()
+
+        # Wait for the update checking to finish
+        Wait(self.marionette, self.TIMEOUT_UPDATE_CHECK).until(
+            lambda _: self.wizard_state != 'checkingForUpdates')
```
whimboo: Can you please add a message to the `until` call? You also want to add a check for `!= checkForUpdates`. If that is not included the test would fail if panels are changing too slowly after a click on the button. This is something which I added recently to the Mozmill Tests. If you don't see it, please update your local mozmill-tests repository. 

I would prefer to see this as a separate method similar to the Mozmill's about window.

firefox_puppeteer/ui/about_window/window.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def download(self, wait_for_finish=True, timeout=TIMEOUT_UPDATE_DOWNLOAD):
+        """ Download the update
+
+        :param wait_for_finish: Optional, True, if the function has to wait
+        for the download to be finished, default to `True`
```
whimboo: I would phrase it `Optional, If True the function ...`.

firefox_puppeteer/ui/about_window/window.py
```diff
@@ -0,0 +1,140 @@
...
+        :param timeout: How long to wait for the download to finish
```
whimboo: Please add the optional part here too.

firefox_puppeteer/ui/about_window/window.py
```diff
@@ -0,0 +1,140 @@
...
+        """
+        assert self.software_update.update_channel.default_channel ==\
+            self.software_update.update_channel.channel,\
+            'The update channel has been set correctly.'
```
whimboo: I think that we really have to use assertEqual() here if possible. With `assert` alone we do not see the different values anymore. This would make it very hard to debug.

firefox_puppeteer/ui/about_window/window.py
```diff
@@ -0,0 +1,140 @@
...
+        if self.wizard_state == 'downloadAndInstall':
+            self.deck.click_download_button()
+
+            # Wait for the download to start
+            Wait(self.marionette).until(
+                lambda _: self.wizard_state != 'downloadAndInstall')
+
+        # If there are incompatible addons we fallback on old software update dialog for updating
+        if self.wizard_state == 'applyBillboard':
+            self.deck.click_update_button()
```
whimboo: We should not click the button as long as the fallback window has not been implemented.

firefox_puppeteer/ui/about_window/window.py
```diff
@@ -0,0 +1,140 @@
...
+            #
+            #     return;
+
+        if wait_for_finish:
+            start_time = datetime.now()
+            self.wait_for_download_finished(timeout)
+            self.download_duration = datetime.now() - start_time
+
+    def wait_for_download_finished(self, timeout):
+        """ Waits until download is completed
```
whimboo: nit: final dot.

firefox_puppeteer/ui/about_window/window.py
```diff
@@ -0,0 +1,140 @@
...
+
+        :param timeout: How long to wait for the download to complete
+        """
+
+        Wait(self.marionette, timeout).until(
```
whimboo: nit: no empty line after docstrings

firefox_puppeteer/ui/about_window/window.py
```diff
@@ -0,0 +1,140 @@
...
+            lambda _: self.wizard_state != 'downloading',
+            message='Download has been completed.')
+
+        assert self.wizard_state != 'downloadFailed'
```
whimboo: Same here as above and for all the other asserts.

firefox_puppeteer/ui/about_window/window.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def wait_for_update_applied(self, timeout=TIMEOUT_UPDATE_APPLY):
+        """ Waits until the downloaded update has been applied
+
+        :param timeout: How long to wait for the update to apply
```
whimboo: nit: final dots and empty line below

firefox_puppeteer/ui/windows.py
```diff
@@ -100,6 +100,9 @@ def create_window_instance(self, handle, expected_class=None):
         elif window_type == 'Browser:page-info':
             from .pageinfo.window import PageInfoWindow
             window = PageInfoWindow(lambda: self.marionette, handle)
+        elif window_type == 'Browser:About':
+            from .about_window.window import AboutWindow
+            window = AboutWindow(lambda: self.marionette, handle)
```
whimboo: Please move up before BrowserWindow, so its alphabetical sorted

whimboo:
I made a lot of comments. Sorry for that but I had to throw out all of them. Let me know if you have specific questions so we can sort that out quickly. Thanks!

firefox_puppeteer/ui/about_window/deck.py
whimboo: Panels are represented as `deck` nodes and have an id. This link might help you: https://dxr.mozilla.org/mozilla-central/source/browser/base/content/aboutDialog.xul#52.

firefox_puppeteer/tests/test_about_window.py
bobsilverberg: I can see from Travis that this fails because there are no updates available, so I think we need to just nix this entire test.

firefox_puppeteer/tests/test_about_window.py
bobsilverberg: I agree. This test is not going to be valid. I will remove it.

firefox_puppeteer/ui/about_window/window.py
bobsilverberg: If we don't have properties should I include those in the test? Would I define an empty array in the `AboutWindow` object and then verify that it is empty in the test, or should we neither define nor test for it?

firefox_puppeteer/ui/about_window/window.py
bobsilverberg: @whimboo May I leave them as uppercase? This is how I've seen it done in other Python code, and PEP8 seems to support it.

firefox_puppeteer/ui/about_window/window.py
bobsilverberg: @whimboo I am not sure what you mean by this.

firefox_puppeteer/ui/about_window/window.py
whimboo: Strange. I may remember wrongly then? Yes, please leave what you have.

firefox_puppeteer/tests/test_about_window.py
whimboo: Sounds good. We will have the update tests soon to actually verify that.

firefox_puppeteer/ui/about_window/window.py
whimboo: If you don't have properties nothing has to be done for the AboutWindow. The BaseWindow already defines that empty array. Thinking more you could leave it out from the test given that we do not change it compared to the BaseWindow.

firefox_puppeteer/ui/about_window/window.py
whimboo: Imagine you open the about window and call this method. It will return true because the initial panel will be `!= 'noUpdatesFound'. The software update library is the place to query for an available update.

firefox_puppeteer/docs/ui/about_window/window.rst
```diff
@@ -0,0 +1,16 @@
...
+.. autoclass:: firefox_puppeteer.ui.about_window.window.AboutWindow
+   :members:
+   :inherited-members:
+
+Deck
+----
+
+.. autoclass:: firefox_puppeteer.ui.about_window.deck.Deck
+   :members:
+   :inherited-members:
```
whimboo: We should also document the Panel classes.

firefox_puppeteer/tests/test_about_window.py
```diff
@@ -0,0 +1,91 @@
...
+        panel = self.about_window.deck.apply
+        self.assertEqual(panel.element.get_attribute('localName'), 'hbox')
+        self.assertEqual(panel.update_button.get_attribute('localName'), 'button')
+
+        # apply_billboard panel
+        panel = self.about_window.deck.apply_billboard
+        self.assertEqual(panel.element.get_attribute('localName'), 'hbox')
+        self.assertEqual(
+            panel.apply_billboard_button.get_attribute('localName'),
+            'button')
```
whimboo: Please fix the indentation by moving line 43 up behind the opening bracket. If the type has to wrap align it with `p`. Same for the other asserts here.

firefox_puppeteer/tests/test_about_window.py
```diff
@@ -0,0 +1,91 @@
...
+
+        self.about_window.close()
+        for trigger in open_strategies:
+            about_window = self.browser.open_about_window(trigger=trigger)
+            self.assertEquals(about_window, self.windows.current)
+            about_window.close()
+
+    def test_patch_info(self):
+        self.assertEqual(self.about_window.patch_info['download_duration'], -1)
+        self.assertNotEqual(self.about_window.patch_info['channel'], None)
```
whimboo: Please use assertNotNone.

firefox_puppeteer/ui/about_window/deck.py
```diff
@@ -0,0 +1,190 @@
...
+
+        :returns: :class:`Panel` instance
+        """
+        mapping = {'apply': ApplyPanel,
+                   'applyBillboard': ApplyBillboardPanel,
+                   'checkForUpdates': CheckForUpdatesPanel,
+                   'checkingForUpdates': CheckingForUpdatesPanel,
+                   'downloadAndInstall': DownloadAndInstallPanel,
+                   'downloadFailed': DownloadFailedPanel,
+                   'downloading': ApplyBillboardPanel
```
whimboo: Why is that the same as the 2nd item?

firefox_puppeteer/ui/about_window/deck.py
```diff
@@ -0,0 +1,190 @@
...
+        return self.element.get_attribute('id') == other.element.get_attribute('id')
+
+
+class ApplyBillboardPanel(Panel):
+
+    @property
+    def apply_billboard_button(self):
+        """The DOM element which represents the Apply Billboard button.
+
+        :returns: Reference to the textbox element.
```
whimboo: It's a button and not a textbox. Same for other properties of other panels.

firefox_puppeteer/ui/about_window/window.py
```diff
@@ -0,0 +1,141 @@
...
+from deck import Deck
+
+
+class AboutWindow(BaseWindow):
+    """Representation of the About window."""
+    window_type = 'Browser:About'
+
+    dtds = [
+        'chrome://branding/locale/brand.dtd',
+        'chrome://browser/locale/aboutDialog.dtd'
```
whimboo: nit: please use a final comma which lowers the changed lines later.

firefox_puppeteer/ui/about_window/window.py
```diff
@@ -0,0 +1,141 @@
...
+    ]
+
+    TIMEOUT_UPDATE_APPLY = 300
+    TIMEOUT_UPDATE_CHECK = 30
+    TIMEOUT_UPDATE_DOWNLOAD = 360
+
+    def __init__(self, *args, **kwargs):
+        BaseWindow.__init__(self, *args, **kwargs)
+
+        self.software_update = SoftwareUpdate(lambda: self.marionette)
```
whimboo: This also needs to be private.

firefox_puppeteer/ui/about_window/window.py
```diff
@@ -0,0 +1,141 @@
...
+
+        :returns: A dictionary with information about the active patch
+        """
+        patch = self.software_update.patch_info
+        patch['download_duration'] = self._download_duration
+        return patch
+
+    def check_for_updates(self):
+        """Clicks on "Check for Updates" button, and waits for check to complete."""
+        self.deck.check_for_updates_button.click()
```
whimboo: This should be `self.deck.check_for_updates.button.click()`. Doesn't this fail?

firefox_puppeteer/ui/about_window/window.py
```diff
@@ -0,0 +1,141 @@
...
+        for the download to be finished, default to `True`
+        :param timeout: Optional, How long to wait for the download to finish,
+        default to 360 seconds
+        """
+        assert self.software_update.update_channel.default_channel == \
+               self.software_update.update_channel.channel, \
+            'The update channel has been set correctly. ' \
+            'default_channel: is {}, while channel is: {}'.format(
+                self.software_update.update_channel.default_channel,
+                self.software_update.update_channel.channel)
```
whimboo: Lots of code, maybe we can find something better later. This works for now.

firefox_puppeteer/ui/about_window/window.py
```diff
@@ -0,0 +1,141 @@
...
+
+        if self.deck.selected(self.deck.download_and_install):
+            self.deck.download_button.click()
```
whimboo: Same here. This should be `self.deck.download_and_install.button.click()`.

bobsilverberg:
Pushed a new commit with just the single PEP8 fix to see if Travis is okay. It's strange because I have a PEP8 checked set up with a pre-commit hook for the repo, and it does report things to me from time to time. I'm not sure why some errors get past it.

whimboo:
Strange. The Linux build totally failed. I just retriggered this build. Lets see.

firefox_puppeteer/tests/test_about_window.py
bobsilverberg: afaict, `assertNotNone` does not exist. We could change this to just `assertTrue`.

firefox_puppeteer/ui/about_window/window.py
```diff
@@ -0,0 +1,142 @@
...
+            self.deck.selected_panel != self.deck.checking_for_updates,
+            message='An update has been found.')
+
+    def wait_for_download_finished(self, timeout=TIMEOUT_UPDATE_DOWNLOAD):
+        """ Waits until download is completed.
+
+        :param timeout: Optional, How long to wait for the download to finish,
+        default to 360 seconds
+        """
+        Wait(self.marionette, timeout).until(
```
bobsilverberg: @whimboo This does not seem to be working. The code does not wait here, but just continues. I added the extra condition that the selected panel is not the `download_and_install` panel, in addition to not being the `downloading` panel, but it doesn't seem to be working.

whimboo: I think that I will have to fix it when working on the update tests. It's hard to proof otherwise.

firefox_puppeteer/tests/test_about_window.py
whimboo: It's actually [assertIsNotNone](https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertIsNotNone)

whimboo:
This PR is still failing also due to the not fixed `Cu` usage in prefs.py. I will work on the remaining fixes. Beside those two things all looks  great. I will open a new PR with my changes once its ready.

whimboo:
Closing in favor of PR #121.

PR #108, opened 2015-02-27
Bug 1137637 - Get rid of sleep(.5) in windows.py module. r=chmanchester

whimboo:


chmanchester:
When I run this with e10s the test hangs.

PR #109, opened 2015-02-27
Bug 1128968 - toolbar classes have to inherit from UIBaseClass and not BaseClass

abbn1090:
Bug 1128968 - toolbar classes have to inherit from UIBaseClass and not BaseClass


firefox_puppeteer/ui/toolbars.py
```diff
@@ -6,13 +6,21 @@
...
 
 
-class NavBar(BaseLib):
+class NavBar(UIBaseLib):
     """Provides access to the DOM elements contained in the
     navigation bar as well as the location bar."""
+    def __init__(self, marionette_getter, window):
+        UIBaseLib.__init__(self, marionette_getter, window)
+        # TODO: A "utility" module that sets up the client directly would be
+        # useful here.
```
whimboo: Please omit the comment. We know about it and a bug is on file for it.

firefox_puppeteer/ui/toolbars.py
```diff
@@ -38,13 +46,21 @@ def home_button(self):
...
         """Provides access to the DOM elements contained in the
         locationbar.
 
         See the :class:`LocationBar` reference.
         """
+        if not self._locationbar:
+
+
+        
+        	from .toolbars import LocationBar
```
whimboo: Lets do the import globally, and also please remove the empty lines above.

firefox_puppeteer/ui/toolbars.py
```diff
@@ -6,13 +6,21 @@
...
     """Provides access to the DOM elements contained in the
     navigation bar as well as the location bar."""
+    def __init__(self, marionette_getter, window):
+        UIBaseLib.__init__(self, marionette_getter, window)
+        # TODO: A "utility" module that sets up the client directly would be
+        # useful here.
+        self._locationbar = None
+
+        self.l10n = L10n(self.get_marionette)
+        self.keys = Keys(self.get_marionette)
```
whimboo: I don't see that we would need those two classes in the NavBar class. I think they can be removed. Or do I miss something?

firefox_puppeteer/ui/toolbars.py
```diff
@@ -55,26 +71,32 @@ def menu_button(self):
...
+    def __init__(self, marionette_getter, window):
+        UIBaseLib.__init__(self, marionette_getter, window)
         # TODO: A "utility" module that sets up the client directly would be
         # useful here.
+        self._autocomplete_results = None
+        self._identity_popup = None
+        
+		
         self.l10n = L10n(self.get_marionette)
         self.keys = Keys(self.get_marionette)
```
whimboo: Both class instances are not needed anymore. You can remove them in favor of `self.window.send_shortcut`.

firefox_puppeteer/ui/toolbars.py
```diff
@@ -144,13 +165,21 @@ def history_drop_marker(self):
...
         locationbar.
 
         See the :class:`IdentityPopup` reference.
         """
+        if not self._identity_popup:
+
+
+        
+        	from .toolbars import IdentityPopup
+        	self._identity_popup = IdentityPopup(lambda: self.marionette, self.window)
```
whimboo: It looks like that you are using tabs for indentation. Please disable that feature of your text editor and instead use 4 blanks. Also remove the empty lines above.

firefox_puppeteer/ui/toolbars.py
```diff
@@ -340,8 +369,16 @@ def selected_index(self):
...
+class IdentityPopup(UIBaseLib):
     """Wraps DOM elements and methods for interacting with the identity popup."""
+    def __init__(self, marionette_getter, window):
+        UIBaseLib.__init__(self, marionette_getter, window)
+        # TODO: A "utility" module that sets up the client directly would be
+        # useful here.
+     
+
+        self.l10n = L10n(self.get_marionette)
+        self.keys = Keys(self.get_marionette)
```
whimboo: We don't need those either here.

whimboo:
@abbn1090, I have added a couple of comments which should guide you into the right direction. Please let me know if some of those are unclear.

abbn1090:
thank you so much for the feedback . i made the changes you've mentioned .
can you please see if i missed anything.
thanks

whimboo:
@abbn1090 it would be good if you could add the review flag on the github link attachment on the bug in Bugzilla if a patch is ready for review. Also before doing it make sure that your changes are passing Travis CI. This is not the case here. Please see the `X Failed` line with details about the failure. You are still using too long lines. I would suggest you run the pep8 command locally to ensure it doesn't fail.

firefox_puppeteer/ui/toolbars.py
```diff
@@ -6,13 +6,19 @@
...
 
 
-class NavBar(BaseLib):
+class NavBar(UIBaseLib):
     """Provides access to the DOM elements contained in the
     navigation bar as well as the location bar."""
+    def __init__(self, marionette_getter, window):
+        UIBaseLib.__init__(self, marionette_getter, window)
+        
+        from .toolbars import LocationBar
```
whimboo: Why did you move this up here from the locationbar property? Please move it back.

firefox_puppeteer/ui/toolbars.py
```diff
@@ -55,26 +65,27 @@ def menu_button(self):
...
-    @use_class_as_property('ui.toolbars.AutocompleteResults')
+    @property
     def autocomplete_results(self):
         """Provides access to and methods for the location bar
         autocomplete results.
 
         See the :class:`AutocompleteResults` reference."""
+        if not self._autocomplete_results:
+        	from .toolbars import AutocompleteResults
+        	self._autocomplete_results = AutocompleteResults(lambda: self.marionette, self.window)
```
whimboo: You have still tabulators enabled for indentation. Please disable that in your editor and use 4 chars instead.

firefox_puppeteer/ui/toolbars.py
```diff
@@ -191,10 +206,10 @@ def reload_url(self, trigger='button', force=False):
         if trigger == 'button':
             self.reload_button.click()
         elif trigger == 'shortcut':
-            cmd_key = self.l10n.get_entity(LocationBar.dtds, 'reloadCmd.commandkey')
-            self.urlbar.send_keys(cmd_key)
+            cmd_key = self.window.get_entity('reloadCmd.commandkey')
+            self.window.send_shortcut(cmd_key)
         elif trigger == 'shortcut2':
-            self.urlbar.send_keys(self.keys.F5)
+            self.urlbar.send_keys(Keys.F5)
```
whimboo: Lets use `self.window.send_shortcut` here too.

firefox_puppeteer/ui/toolbars.py
```diff
@@ -229,15 +244,11 @@ def value(self):
...
     """Wraps DOM elements and methods for interacting with autocomplete results."""
 
-    def __init__(self, *args, **kwargs):
-        BaseLib.__init__(self, *args, **kwargs)
-        # TODO: A "utility" module that sets up the client directly would be
-        # useful here.
-        self.l10n = L10n(self.get_marionette)
-        self.keys = Keys(self.get_marionette)
+    def __init__(self, marionette_getter, window):
+        UIBaseLib.__init__(self, marionette_getter, window)
```
whimboo: If there is no custom initialization code you can simply remove this method. Python is smart enough to call `__init_` from the super class automatically.

firefox_puppeteer/ui/toolbars.py
```diff
@@ -340,8 +351,10 @@ def selected_index(self):
         return self.results.get_attribute('selectedIndex')
 
 
-class IdentityPopup(BaseLib):
+class IdentityPopup(UIBaseLib):
     """Wraps DOM elements and methods for interacting with the identity popup."""
+    def __init__(self, marionette_getter, window):
+        UIBaseLib.__init__(self, marionette_getter, window)
```
whimboo: Same here. It can be removed due to no custom code in this method.

firefox_puppeteer/ui/toolbars.py
```diff
@@ -38,13 +42,18 @@ def home_button(self):
...
+    @property
     def locationbar(self):
         """Provides access to the DOM elements contained in the
         locationbar.
 
         See the :class:`LocationBar` reference.
         """
+        if not self._locationbar:
+            from .toolbars import LocationBar
+            self._locationbar = LocationBar(lambda: self.marionette, self.window, self.element)
```
whimboo: The last parameter for the element has to be the DOM node for the location bar itself. Currently you can find that inside the LocationBar class as [`urlbar`](https://github.com/mozilla/firefox-ui-tests/blob/master/firefox_puppeteer/ui/toolbars.py#L215) property. Call the `find_element()` method before this line and pass in the element. Similar to what I do here: https://github.com/mozilla/firefox-ui-tests/blob/master/firefox_puppeteer/ui/windows.py#L473

whimboo: When you do this the urlbar property can return `self.element` given that we cache that.

firefox_puppeteer/ui/toolbars.py
```diff
@@ -55,26 +64,28 @@ def menu_button(self):
...
+    @property
     def autocomplete_results(self):
         """Provides access to and methods for the location bar
         autocomplete results.
 
         See the :class:`AutocompleteResults` reference."""
+        if not self._autocomplete_results:
+            from .toolbars import AutocompleteResults
+            self._autocomplete_results = AutocompleteResults(lambda: self.marionette, self
+                                                             .window, self.element)
```
whimboo: Same as above for the location bar. Find the root element of the autocomplete result here and pass it in.

firefox_puppeteer/ui/toolbars.py
```diff
@@ -144,13 +154,19 @@ def history_drop_marker(self):
...
     def identity_popup(self):
         """Provides utility members for accessing and manipulating the
         locationbar.
 
         See the :class:`IdentityPopup` reference.
         """
+        if not self._identity_popup:
+            from .toolbars import IdentityPopup
+            self._identity_popup = IdentityPopup(lambda: self.marionette, self
+                                                 .window, self.element)
```
whimboo: The same applies here.

firefox_puppeteer/ui/windows.py
```diff
@@ -449,13 +450,20 @@ def is_private(self):
...
         """Provides access to the navigation bar. This is the toolbar containing
         the back, forward and home buttons. It also contains the location bar.
 
         See the :class:`~ui.toolbars.NavBar` reference.
         """
+        self.switch_to()
+
+        if not self._navbar:
+            from .toolbars import NavBar
+            self._navbar = NavBar(lambda: self.marionette, self, self.window_element)
```
whimboo: The same here for the element to pass in. In this case it has to be the DOM node with the ID `nav-bar`. It's not used in the NavBar class yet, but will be available then via the `element` property.

whimboo:
This has been fixed meanwhile via another PR.

PR #112, opened 2015-03-02
Bug 1132710 - Convert Mozmill testSubmitUnencryptedInfoWarning

galgeek:
https://bugzilla.mozilla.org/show_bug.cgi?id=1132710

### DIFF
--- a/firefox_ui_tests/remote/security/test_submit_unencrypted_info_warning.py
+++ b/firefox_ui_tests/remote/security/test_submit_unencrypted_info_warning.py
@@ -1,0 +1,64 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import time
+
+from marionette_driver import By
+
+from marionette_driver.errors import NoAlertPresentException
+from marionette_driver.marionette import Alert
+
+from firefox_ui_harness.testcase import FirefoxTestCase
+
+
+class TestSubmitUnencryptedInfoWarning(FirefoxTestCase):
+
+    def setUp(self):
+        FirefoxTestCase.setUp(self)
+
+        self.url = 'https://ssl-dv.mozqa.com/data/firefox/security/unencryptedsearch.html'
+        self.test_string = 'mozilla'
+
+        self.prefs.set_pref('security.warn_submit_insecure', True)
+
+    def test_submit_unencrypted_info_warning(self):
+        with self.marionette.using_context('content'):
+            # Navigate to the test page
+            self.marionette.navigate(self.url)
+
+            # Get the page's search box and submit button
+            searchbox = self.marionette.find_element(By.ID, 'q')
+            button = self.marionette.find_element(By.ID, 'submit')
+
+            # Prepare to handle warning
+            with self.assertRaises(NoAlertPresentException):
+                warning = self.marionette.switch_to_alert()
+                warning.text
+
+            # Get the warning message text and replace its two instances of "##" with "\n\n"
+            message = self.browser.get_property('formPostSecureToInsecureWarning.message')
+            message = message.replace('##', '\n\n')
+
+            # Use the page's search box to submit information
+            searchbox.send_keys(self.test_string)
+            button.click()
+
+            # Handle warning and check its message text
+            self.wait_for_condition(lambda _: self.alert_present())
+            warning = self.marionette.switch_to_alert()
+            self.assertEqual(warning.text, message)
+            warning.accept()
+
+            # Wait while the search results page (re)loads,
+            # the check that search_term contains the expected text
+            time.sleep(1)
+            search_term = self.marionette.find_element(By.ID, 'search-term')
+            self.assertEqual(search_term.get_attribute('textContent'), self.test_string)
+
+    def alert_present(self):
+        try:
+            Alert(self.marionette).text
+            return True
+        except NoAlertPresentException:
+            return False
--- a/firefox_ui_tests/remote/security/test_submit_unencrypted_info_warning.py
+++ b/firefox_ui_tests/remote/security/test_submit_unencrypted_info_warning.py
@@ -4,7 +4,7 @@
 
 import time
 
-from marionette_driver import By
+from marionette_driver import By, expected
 
 from marionette_driver.errors import NoAlertPresentException
 from marionette_driver.marionette import Alert
@@ -35,27 +35,28 @@ class TestSubmitUnencryptedInfoWarning(FirefoxTestCase):
             button.click()
 
             # Define a function, per modal dialog unit tests for marionette, to handle warning.
-            def alert_present(mi):
+            def alert_present(mn):
                 try:
-                    Alert(mi).text
+                    Alert(mn).text
                     return True
                 except NoAlertPresentException:
                     return False
+
+            # Wait for the warning, check its message text, and "accept" it.
+            self.wait_for_condition(lambda _: alert_present(self.marionette))
+            warning = Alert(self.marionette)
 
             # Get the warning message text and replace its two instances of "##" with "\n\n".
             message = self.browser.get_property('formPostSecureToInsecureWarning.message')
             message = message.replace('##', '\n\n')
 
-            # Wait for the warning, check its message text, and "accept" it.
-            self.wait_for_condition(lambda _: alert_present(self.marionette))
-            warning = Alert(self.marionette)
+            # Verify the text matches, then accept the warning
             self.assertEqual(warning.text, message)
             warning.accept()
             self.wait_for_condition(lambda _: not alert_present(self.marionette))
 
-            # Wait while the search results page (re)loads
-            # (note that neither expected element_present nor element_displayed help here),
+            # Wait while the page updates,
             # then check that search_term contains the expected text.
-            time.sleep(0.5)
+            self.wait_for_condition(expected.element_stale(searchbox))
             search_term = self.marionette.find_element(By.ID, 'search-term')
             self.assertEqual(search_term.get_attribute('textContent'), self.test_string)
--- a/firefox_ui_tests/remote/security/test_submit_unencrypted_info_warning.py
+++ b/firefox_ui_tests/remote/security/test_submit_unencrypted_info_warning.py
@@ -21,7 +21,6 @@ class TestSubmitUnencryptedInfoWarning(FirefoxTestCase):
         self.test_string = 'mozilla'
 
         self.prefs.set_pref('security.warn_submit_insecure', True)
-        self.prefs.set_pref('prompts.tab_modal.enabled', True)
 
     def test_submit_unencrypted_info_warning(self):
         with self.marionette.using_context('content'):
@@ -48,11 +47,10 @@ class TestSubmitUnencryptedInfoWarning(FirefoxTestCase):
                 except NoAlertPresentException:
                     return False
 
-            # Handle warning and check its message text.
+            # Wait for the warning, check its message text, and "accept" it.
             self.wait_for_condition(lambda _: alert_present(self))
-            warning = self.marionette.switch_to_alert()
-            self.assertEqual(warning.text, message)
-            warning.accept()
+            self.assertEqual(Alert(self.marionette).text, message)
+            Alert(self.marionette).accept()
             self.wait_for_condition(lambda _: not alert_present(self))
 
             # Wait while the search results page (re)loads,
--- a/firefox_ui_tests/remote/security/test_submit_unencrypted_info_warning.py
+++ b/firefox_ui_tests/remote/security/test_submit_unencrypted_info_warning.py
@@ -21,44 +21,42 @@ class TestSubmitUnencryptedInfoWarning(FirefoxTestCase):
         self.test_string = 'mozilla'
 
         self.prefs.set_pref('security.warn_submit_insecure', True)
+        self.prefs.set_pref('prompts.tab_modal.enabled', True)
 
     def test_submit_unencrypted_info_warning(self):
         with self.marionette.using_context('content'):
-            # Navigate to the test page
+            # Navigate to the test page.
             self.marionette.navigate(self.url)
 
-            # Get the page's search box and submit button
+            # Get the page's search box and submit button.
             searchbox = self.marionette.find_element(By.ID, 'q')
             button = self.marionette.find_element(By.ID, 'submit')
 
-            # Prepare to handle warning
-            with self.assertRaises(NoAlertPresentException):
-                warning = self.marionette.switch_to_alert()
-                warning.text
-
-            # Get the warning message text and replace its two instances of "##" with "\n\n"
+            # Get the warning message text and replace its two instances of "##" with "\n\n".
             message = self.browser.get_property('formPostSecureToInsecureWarning.message')
             message = message.replace('##', '\n\n')
 
-            # Use the page's search box to submit information
+            # Use the page's search box to submit information.
             searchbox.send_keys(self.test_string)
             button.click()
 
-            # Handle warning and check its message text
-            self.wait_for_condition(lambda _: self.alert_present())
+            # Define a function, from marionette client unit tests, to help handle warning.
+            def alert_present(self):
+                try:
+                    Alert(self.marionette).text
+                    return True
+                except NoAlertPresentException:
+                    return False
+
+            # Handle warning and check its message text.
+            self.wait_for_condition(lambda _: alert_present(self))
             warning = self.marionette.switch_to_alert()
             self.assertEqual(warning.text, message)
             warning.accept()
+            self.wait_for_condition(lambda _: not alert_present(self))
 
             # Wait while the search results page (re)loads,
-            # the check that search_term contains the expected text
+            # then check that search_term contains the expected text.
             time.sleep(1)
             search_term = self.marionette.find_element(By.ID, 'search-term')
             self.assertEqual(search_term.get_attribute('textContent'), self.test_string)
-
-    def alert_present(self):
-        try:
-            Alert(self.marionette).text
-            return True
-        except NoAlertPresentException:
-            return False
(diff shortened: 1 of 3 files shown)


firefox_ui_tests/remote/security/test_submit_unencrypted_info_warning.py
```diff
@@ -0,0 +1,64 @@
...
+            self.marionette.navigate(self.url)
+
+            # Get the page's search box and submit button
+            searchbox = self.marionette.find_element(By.ID, 'q')
+            button = self.marionette.find_element(By.ID, 'submit')
+
+            # Prepare to handle warning
+            with self.assertRaises(NoAlertPresentException):
+                warning = self.marionette.switch_to_alert()
+                warning.text
```
chmanchester: This could probably be self.assertFalse(alert_present())

whimboo: Please test the negative case and how the test would handle the case if an alert is shown. It should not be stuck at this point but cleanup itself correctly.

galgeek:
This is the formulation the marionette client unit tests use to check alert messages for tab modal dialogs:
https://dxr.mozilla.org/mozilla-central/source/testing/marionette/client/marionette/tests/unit/test_modal_dialogs.py#76

Chris and I talked on IRC about whether the with... block at line 35 of the test in the original commit is necessary in the context of firefox-ui-tests, and we agreed that it's not.

I've noticed that the marionette unit tests set pref "prompts.tab_modal.enabled" to True, and I've added that here. (There's another class in the marionette unit tests file for "global modals" that sets this pref to False, but there are not yet any unit tests defined in that class.)

chmanchester:
I wouldn't worry about that pref, I don't think it will effect this modal. The unit tests for global modals in that test case are run courtesy of the superclass. Sneaky, I know ;)

galgeek:
Thanks again for the feedback!

I’ve removed the likely unnecessary pref and rewritten the test to use Alert() rather than a variable.

firefox_ui_tests/remote/security/test_submit_unencrypted_info_warning.py
```diff
@@ -0,0 +1,60 @@
...
+        FirefoxTestCase.setUp(self)
+
+        self.url = 'https://ssl-dv.mozqa.com/data/firefox/security/unencryptedsearch.html'
+        self.test_string = 'mozilla'
+
+        self.prefs.set_pref('security.warn_submit_insecure', True)
+
+    def test_submit_unencrypted_info_warning(self):
+        with self.marionette.using_context('content'):
+            # Navigate to the test page.
```
whimboo: nit: unnecessary comment.

firefox_ui_tests/remote/security/test_submit_unencrypted_info_warning.py
```diff
@@ -0,0 +1,60 @@
...
+            self.marionette.navigate(self.url)
+
+            # Get the page's search box and submit button.
+            searchbox = self.marionette.find_element(By.ID, 'q')
+            button = self.marionette.find_element(By.ID, 'submit')
+
+            # Get the warning message text and replace its two instances of "##" with "\n\n".
+            message = self.browser.get_property('formPostSecureToInsecureWarning.message')
+            message = message.replace('##', '\n\n')
```
whimboo: nit: please move those lines closer to the code they are used.

firefox_ui_tests/remote/security/test_submit_unencrypted_info_warning.py
```diff
@@ -0,0 +1,60 @@
...
+            searchbox.send_keys(self.test_string)
+            button.click()
+
+            # Define a function, from marionette client unit tests, to help handle warning.
+            def alert_present(self):
+                try:
+                    Alert(self.marionette).text
+                    return True
+                except NoAlertPresentException:
+                    return False
```
whimboo: I wonder if we should get a helper added to the [expected.py](https://dxr.mozilla.org/mozilla-central/source/testing/marionette/driver/marionette_driver/expected.py) module of marionette_driver here. @chmanchester what do you think? If you agree I will go ahead and file a new bug for it.

firefox_ui_tests/remote/security/test_submit_unencrypted_info_warning.py
```diff
@@ -0,0 +1,60 @@
...
+            message = message.replace('##', '\n\n')
+
+            # Use the page's search box to submit information.
+            searchbox.send_keys(self.test_string)
+            button.click()
+
+            # Define a function, from marionette client unit tests, to help handle warning.
+            def alert_present(self):
+                try:
+                    Alert(self.marionette).text
```
whimboo: Is there a reason we have to access `text` here? Is `Alert()` only not enough?

firefox_ui_tests/remote/security/test_submit_unencrypted_info_warning.py
```diff
@@ -0,0 +1,60 @@
...
+            # Get the warning message text and replace its two instances of "##" with "\n\n".
+            message = self.browser.get_property('formPostSecureToInsecureWarning.message')
+            message = message.replace('##', '\n\n')
+
+            # Use the page's search box to submit information.
+            searchbox.send_keys(self.test_string)
+            button.click()
+
+            # Define a function, from marionette client unit tests, to help handle warning.
+            def alert_present(self):
```
whimboo: Please pass in a marionette instance. `self` is not necessary here.

firefox_ui_tests/remote/security/test_submit_unencrypted_info_warning.py
```diff
@@ -0,0 +1,60 @@
...
+                try:
+                    Alert(self.marionette).text
+                    return True
+                except NoAlertPresentException:
+                    return False
+
+            # Wait for the warning, check its message text, and "accept" it.
+            self.wait_for_condition(lambda _: alert_present(self))
+            self.assertEqual(Alert(self.marionette).text, message)
+            Alert(self.marionette).accept()
```
whimboo: Please cache the created instance of `Alert`.

firefox_ui_tests/remote/security/test_submit_unencrypted_info_warning.py
```diff
@@ -0,0 +1,60 @@
...
+            self.wait_for_condition(lambda _: not alert_present(self))
```
whimboo: I don't see the need for this check. `accept` should work as tested in Marionette. Or why would we need that?

firefox_ui_tests/remote/security/test_submit_unencrypted_info_warning.py
```diff
@@ -0,0 +1,60 @@
...
+
+            # Wait while the search results page (re)loads,
+            # then check that search_term contains the expected text.
+            time.sleep(1)
+            search_term = self.marionette.find_element(By.ID, 'search-term')
```
whimboo: No need for sleep at this point. Use the `[element_present](https://dxr.mozilla.org/mozilla-central/source/testing/marionette/driver/marionette_driver/expected.py#18)` helper of the expected module of marionette. But as mentioned already yesterday please file a bug for adding an equivalent of waitForPageLoad() to our library.

firefox_ui_tests/remote/security/test_submit_unencrypted_info_warning.py
galgeek: Alert() alone is not enough.

My impression is that Alert() is always hanging around ready to be used but not always "present", its "presence" defined by having a truthy text property. @chmanchester, is that correct?

firefox_ui_tests/remote/security/test_submit_unencrypted_info_warning.py
galgeek: Sorry! I misunderstood your comment on bugzilla.

firefox_ui_tests/remote/security/test_submit_unencrypted_info_warning.py
chmanchester: Alert(self.marionette) just creates an object on the python side. You have to call a method like text to actually dispatch a command that attempts to interact with the alert. The dialog is evinced by calling the method without getting the NoAlertPresent exception.

firefox_ui_tests/remote/security/test_submit_unencrypted_info_warning.py
galgeek: Chris suggested using it here.

This line is present in the tearDown for marionette's modal dialog unit tests.
https://dxr.mozilla.org/mozilla-central/source/testing/marionette/client/marionette/tests/unit/test_modal_dialogs.py#26

Alert.accept() doesn't appear to do more than send the message. 
https://dxr.mozilla.org/mozilla-central/source/testing/marionette/driver/marionette_driver/marionette.py#516

firefox_ui_tests/remote/security/test_submit_unencrypted_info_warning.py
galgeek: This test page updates strictly via javascript, and all of its elements are present from the initial page load, so element_present succeeds here before the search term text can be checked. 

I've tried expect.element_displayed here, too, with both search-term and search-results, but it times out. This is apparently know to be buggy: https://bugzilla.mozilla.org/show_bug.cgi?id=865232

time.sleep(0.5) works as well for me as the original.

I can file that bug, but if this test succeeds with a quick sleep, and we're using that in so many other places, I'm not sure how this test makes much of a case for the importance of a waitforPageLoad() replacement.

galgeek:
Please check the conversation in the line notes for lots more information!

firefox_ui_tests/remote/security/test_submit_unencrypted_info_warning.py
```diff
@@ -0,0 +1,61 @@
...
+            def alert_present(mi):
+                try:
+                    Alert(mi).text
+                    return True
+                except NoAlertPresentException:
+                    return False
+
+            # Get the warning message text and replace its two instances of "##" with "\n\n".
+            message = self.browser.get_property('formPostSecureToInsecureWarning.message')
+            message = message.replace('##', '\n\n')
```
whimboo: nit: You moved it a bit down but not right before it is used. The `Alert` instantiation can come first.

firefox_ui_tests/remote/security/test_submit_unencrypted_info_warning.py
whimboo: Still waiting for a reply from @chmanchester here.

firefox_ui_tests/remote/security/test_submit_unencrypted_info_warning.py
whimboo: So why does the instantiation not fail for an alert if there is no alert present? Something similar to retrieving an element. Personally I find that a bit awkward to have to call a member to find out if the alert exists.

firefox_ui_tests/remote/security/test_submit_unencrypted_info_warning.py
```diff
@@ -0,0 +1,61 @@
...
+            # Get the page's search box and submit button.
+            searchbox = self.marionette.find_element(By.ID, 'q')
+            button = self.marionette.find_element(By.ID, 'submit')
+
+            # Use the page's search box to submit information.
+            searchbox.send_keys(self.test_string)
+            button.click()
+
+            # Define a function, per modal dialog unit tests for marionette, to handle warning.
+            def alert_present(mi):
```
whimboo: nit: please use `marionette` or `mn` as name. The latter is what Chris is usually using afair.

firefox_ui_tests/remote/security/test_submit_unencrypted_info_warning.py
```diff
@@ -0,0 +1,61 @@
...
+                    return False
+
+            # Get the warning message text and replace its two instances of "##" with "\n\n".
+            message = self.browser.get_property('formPostSecureToInsecureWarning.message')
+            message = message.replace('##', '\n\n')
+
+            # Wait for the warning, check its message text, and "accept" it.
+            self.wait_for_condition(lambda _: alert_present(self.marionette))
+            warning = Alert(self.marionette)
+            self.assertEqual(warning.text, message)
```
whimboo: What's happening with the test when this assertion fails? I assume the alert is not getting closed and this test and all the others will hang?

firefox_ui_tests/remote/security/test_submit_unencrypted_info_warning.py
```diff
@@ -0,0 +1,61 @@
...
+            warning.accept()
+            self.wait_for_condition(lambda _: not alert_present(self.marionette))
+
+            # Wait while the search results page (re)loads
+            # (note that neither expected element_present nor element_displayed help here),
+            # then check that search_term contains the expected text.
+            time.sleep(0.5)
```
whimboo: Can't we check for the searchbox being a stale element here? I really would like to get rid of the sleep.

galgeek:
Thanks for your feedback, Henrik! I've updated the PR.

Additionally, I note that you wondered in line notes above "if we should get a helper added to the expected.py module of marionette_driver here. @chmanchester what do you think? If you agree I will go ahead and file a new bug for it."

My understanding is that Marionette clears up alerts between test files, so that no tearDown is necessary here.

firefox_ui_tests/remote/security/test_submit_unencrypted_info_warning.py
chmanchester: Sorry, this sort of thing is really likely to get lost when github outdates a diff. I have no problem with it, but I'm certainly not the only person you need to talk to for extending marionette APIs, and I don't see how this is blocking the PR. I would recommend you just file the bug for it and flag needinfo to solicit opinions, or post a patch for review if you want this added. For what it's worth the Alert API as it is is inspired by selenium, you can see the discussion around this is in the bug that landed it.

A better approach with the existing marionette APIs might be to use a Wait that explicitly ignores the NoAlertPresentException. Something like (untested code):

```python
alert = self.marionette.switch_to_alert()
wait = Wait(self.marionette, ignored_exceptions=NoAlertPresentException)
wait.unti(lambda _: alert.text == <expected>)
```

Which would only succeed once the alert was present.

chmanchester:
Marionette doesn't do anything to clean up Alerts between tests, but looking at this test there's not much else you could to to make sure it gets cleared for the next test. I commented inline about extending the expected API.

galgeek:
Marionette's unit tests for modal dialogs are here:
https://dxr.mozilla.org/mozilla-central/source/testing/marionette/client/marionette/tests/unit/test_modal_dialogs.py

Here's the bug discussing the current marionette modal dialog implementation:
https://bugzilla.mozilla.org/show_bug.cgi?id=906712

firefox_ui_tests/remote/security/test_submit_unencrypted_info_warning.py
```diff
@@ -0,0 +1,53 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import time
```
whimboo: This module is not used anymore.

firefox_ui_tests/remote/security/test_submit_unencrypted_info_warning.py
```diff
@@ -0,0 +1,53 @@
...
+            button.click()
+
+            # Get the expected warning text and replace its two instances of "##" with "\n\n".
+            message = self.browser.get_property('formPostSecureToInsecureWarning.message')
+            message = message.replace('##', '\n\n')
+
+            # Wait for the warning, verify the expected text matches warning, accept the warning
+            wait = Wait(self.marionette, ignored_exceptions=NoAlertPresentException)
+            warning = Alert(self.marionette)
+            wait.until(lambda _: warning.text == message)
```
whimboo: When you do a negative test here and replace `==` with `!=`, you will see that the alert will stay open. So what you want here is maybe a try/finally block, so the alert will be accepted however the check finishes.

firefox_ui_tests/remote/security/test_submit_unencrypted_info_warning.py
```diff
@@ -0,0 +1,53 @@
...
+            # Use the page's search box to submit information.
+            searchbox.send_keys(self.test_string)
+            button.click()
+
+            # Get the expected warning text and replace its two instances of "##" with "\n\n".
+            message = self.browser.get_property('formPostSecureToInsecureWarning.message')
+            message = message.replace('##', '\n\n')
+
+            # Wait for the warning, verify the expected text matches warning, accept the warning
+            wait = Wait(self.marionette, ignored_exceptions=NoAlertPresentException)
```
whimboo: There is no need to pre-declare it. Just use it like `Wait(...).until(...)`, whereby the code in `until()` is moved to the next line with 4 char of indentation.

firefox_ui_tests/remote/security/test_submit_unencrypted_info_warning.py
```diff
@@ -0,0 +1,53 @@
...
+            warning = Alert(self.marionette)
+            wait.until(lambda _: warning.text == message)
+            warning.accept()
+
+            # Wait while the page updates,
+            self.wait_for_condition(expected.element_stale(searchbox))
+            self.assertRaises(NoAlertPresentException, lambda mn: Alert(mn).text, self.marionette)
```
whimboo: If you check for an assertion of a property please use `with self.assertRaises(NoAlertPresentException):`

galgeek: I don't think we need this with the try: finally: block in place above.

galgeek:
Thanks again for your feedback!

I've omitted the assertRaises here; with the try: finally block, we should be sure to close the alert.

whimboo:
This looks fine to me now, and also works as expected. Thanks!

Merged as https://github.com/mozilla/firefox-ui-tests/commit/36ed081c78c95412e9f9f408dd7eed0c9a6ba7d8

PR #113, opened 2015-03-05
Add coveralls.io support

whimboo:
This PR is used for testing the integration of coveralls.io. It's not something to review yet.


whimboo:
It looks like the coverage command only accepts Python files but no scripts. So we may have to create a "setup.py test" equivalent.

chmanchester:
Cleaning up old PRs. Please re-open when work re-commences on this.

PR #114, opened 2015-03-05
Bug 1137396 - Add support for mozinstall to install/uninstall Firefox. r=chmanchester

whimboo:


setup.py
```diff
@@ -8,6 +8,8 @@
 
 deps = [
     'marionette-client == 0.9',
+    'mozinstall == 1.11',
+    'mozlog == 2.10',
```
chmanchester: mozlog comes in with marionette. If the version is too low we can probably bump it in the client's requirements.txt

firefox_ui_harness/runtests.py
```diff
@@ -4,18 +4,37 @@
...
-class ReleaseTestRunner(BaseMarionetteTestRunner):
+class FirefoxUIOptions(BaseMarionetteOptions):
+
+    def parse_args(self, *args, **kwargs):
+        options, test_files = BaseMarionetteOptions.parse_args(self,
+                                                               *args, **kwargs)
+
+        if not any([(k.startswith('log_') and v is not None and '-' in v)
+                    for (k, v) in vars(options).items()]):
+            options.log_mach = '-'
```
chmanchester: This if can be skipped if you specify {"mach": sys.stdout} instead of "tbpl" below.

firefox_ui_harness/runtests.py
```diff
@@ -29,9 +48,46 @@ def __init__(self, *args, **kwargs):
...
+    finally:
+        # Ensure to uninstall the binary if it has been installed before
+        if installed:
+            options.logger.info('Uninstalling build at "%s"...' % install_folder)
+            mozinstall.uninstall(install_folder)
+
+    return runner
+
+
+def cli(runner_class=FirefoxUITestRunner, parser_class=FirefoxUIOptions):
```
chmanchester: This function is just copied from marionette's runtests.py, as long as it is no need to pull it in.

firefox_ui_harness/runtests.py
```diff
@@ -4,18 +4,37 @@
...
+import firefox_puppeteer
 import firefox_ui_tests
 
-from .arguments import ReleaseTestParser
 from .default_prefs import default_prefs
 from .testcase import FirefoxTestCase
 
 
-class ReleaseTestRunner(BaseMarionetteTestRunner):
+class FirefoxUIOptions(BaseMarionetteOptions):
```
chmanchester: Why pull this into the same file? It's fine to do, just curious.

firefox_ui_harness/runtests.py
```diff
@@ -29,9 +48,46 @@ def __init__(self, *args, **kwargs):
         self.test_handlers = [FirefoxTestCase]
 
 
-def run():
-    cli(runner_class=ReleaseTestRunner, parser_class=ReleaseTestParser)
+def startTestRunner(runner_class, options, tests):
+    try:
+        # If the specified binary is an installer it needs to be installed
+        installed = False
+        if mozinstall.is_installer(options.binary):
```
chmanchester: Should we consider a different option altogether (--installer)? If I have an option called "binary" I'd expect it to really be a binary.

setup.py
whimboo: I was told to include other packages as dependencies if we make use of those ourselves. Even when there is a dependency set for that other package by a package we also depend on. I'm really confused to see those two different feedback. So what is correct?

firefox_ui_harness/runtests.py
whimboo: I don't actually know what's best here. What do you think we should use?

chmanchester: The mach formatter. It has color.

setup.py
chmanchester: Now I'm not sure. If we end up with multiple versions of mozlog that sounds bad, but maybe that's dealt with automatically.

firefox_ui_harness/runtests.py
whimboo: Well, how else we should get our own startTestRunner() method called?

firefox_ui_harness/runtests.py
whimboo: I mimic what we have in marionette base.py and what gaiatest is doing. I don't see a reason why to offload those tiny bits into various modules.

firefox_ui_harness/runtests.py
chmanchester: I see. I guess this is fine to do, we can set our own logging options and do our own command line stuff later here, too.

firefox_ui_harness/runtests.py
whimboo: It was working fine for us all the years long. But if you think that is better, tell me and I will do it. If we do it, then they are exclusive.

setup.py
whimboo: We won't. Running setup.py would raise conflicts. At least I could also add a `mozlog >= 2.7` if you want.

chmanchester: Great. Leave this if it's not causing problems.

firefox_ui_harness/runtests.py
chmanchester: I guess for this to make sense we'll have to add a condition to make sure they aren't both specified, like you say. So I'll leave it up to you whether this is worth doing right now.

firefox_ui_harness/runtests.py
whimboo: This will be fixed in the next commit as talked on IRC.

setup.py
whimboo: One other benefit would be that our package will not be broken if a new release of mozlog has a regression in some part. We can clearly control which version of dependencies we want to have.

firefox_ui_harness/runtests.py
whimboo: I talked to @jonallengriffin on IRC and he agrees in using an --installer option. I will go ahead and get this changed.

firefox_ui_harness/runtests.py
```diff
@@ -3,19 +3,50 @@
...
+    def parse_args(self, *args, **kwargs):
+        options, test_files = BaseMarionetteOptions.parse_args(self, *args, **kwargs)
+
+        # It is not allowed to specify both options
+        if options.installer:
+            if options.binary:
+                self.error('options --binary and --installer are mutually exclusive')
+
+            # Spoofing so verify_usage() is not showing a failure
+            options.binary = 'to_be_set'
```
chmanchester: This is sort of too bad, but I can't really think of a better way.

firefox_ui_harness/runtests.py
```diff
@@ -3,19 +3,50 @@
...
+                        action='store',
+                        help='installer of a Gecko application to use for running the tests')
+
+    def parse_args(self, *args, **kwargs):
+        options, test_files = BaseMarionetteOptions.parse_args(self, *args, **kwargs)
+
+        # It is not allowed to specify both options
+        if options.installer:
+            if options.binary:
+                self.error('options --binary and --installer are mutually exclusive')
```
chmanchester: Raise an exception or exit too.

.travis.yml
```diff
@@ -32,12 +31,11 @@ before_script:
...
 
-    # Download Firefox Nightly, which is compatible with the Marionette client version used
-    - "cd && mozdownload -t daily"
-    - mozinstall *-firefox-*
+    # Download latest Firefox Nightly, which is compatible with the Marionette client version used
+    - mozdownload -t daily
 
 script:
-    - firefox-ui-tests --binary $BINARY
+    - firefox-ui-tests --binary *-firefox-*
```
chmanchester: Travis is failing, this probably needs an update.

whimboo:
PR got merged as https://github.com/mozilla/firefox-ui-tests/commit/e09f45926c82081fd23112c1459077230afa0975.

PR #117, opened 2015-03-06
Bug 1140294 - focused_chrome_window_handle fails if there is no active window. r=chmanchester

whimboo:


firefox_puppeteer/ui/windows.py
```diff
@@ -42,11 +42,16 @@ def focused_chrome_window_handle(self):
...
+              Components.utils.import("resource://gre/modules/Services.jsm");
+
+              let win = Services.focus.activeWindow;
+              if (win) {
+                return win.QueryInterface(Components.interfaces.nsIInterfaceRequestor)
+                          .getInterface(Components.interfaces.nsIDOMWindowUtils)
+                          .outerWindowID.toString();
+              }
+
+              return null;
```
chmanchester: Good not to fail completely here, but in the two places this member is used returning None would still result in a test failure, so I guess we need to update those callers to use Wait as well if we're sure this is the race we think it is.

whimboo: The test failure you are talking about here, I can only see at https://github.com/mozilla/firefox-ui-tests/blob/master/firefox_puppeteer/tests/test_windows.py#L150. For any other usage of this method I cannot see that it could fail. Which other instance are you talking about?

chmanchester: I'm referring to the two uses in that test.

firefox_puppeteer/tests/test_windows.py
```diff
@@ -128,9 +128,16 @@ def closer(window):
         win2 = BaseWindow(lambda: self.marionette, win2.handle)
 
         self.assertEquals(len(self.marionette.chrome_window_handles), 2)
-
         win2.close(callback=closer)
-        win1.switch_to()
+
+        win1.focus()
+
+        # We cannot open a chrome window in the background yet, so this will fail
```
chmanchester: I don't understand this comment. The opener script opens this in the foreground, how does that mean we "cannot"?

chmanchester: I don't understand the comment. The opener script opens in the foreground, how does that mean we "cannot" open a window in the background?

whimboo: There are options you can specify to `window.open()` which let the new chrome window open in the background.

firefox_puppeteer/ui/windows.py
```diff
@@ -293,14 +300,15 @@ def get_property(self, property_id):
...
+    def open_window(self, callback=None, expected_window_class=None, is_focused=True):
         """Opens a new top-level chrome window.
 
         :param callback: Optional, function to trigger the window to open. It is
          triggered with the current :class:`BaseWindow` as parameter.
          Defaults to `window.open()`.
-
         :param expected_class: Optional, check for the correct window class.
+        :param is_focused: Optional, waits until the new window will be focused.
+         Defaults to `True`.
```
chmanchester: I think this parameter is a little confusing as documented. It doesn't just turn on the wait when is_focused=True, it waits for it not to be focused when is_focused=False. Would it be clearer if we just had "wait_for_focus" as the parameter? The case we're not expecting a new window to be focused seems more rare.

firefox_puppeteer/tests/test_windows.py
chmanchester: But you didn't specify them? I don't understand where the actual limitation it.

firefox_puppeteer/ui/windows.py
whimboo: Oh, I changed the handling but missed to update the comment. This is just the expected focus state of the newly opened window. It can be `True` or `False`. So we would have to wait for both possible states. As you say opening windows in the background is rare, that's why I choose `True`. Would that make sense? I wouldn't explicitly expose the wait in the description given that we do it anyway.

firefox_puppeteer/tests/test_windows.py
whimboo: I wanted to with the initial implementation but it was too hard and I wasn't able to get it working. Given that we don't have this feature yet, we should not mention it here. I will update the comment.

firefox_puppeteer/ui/windows.py
whimboo: How does it sound now? Is that ok for you?

chmanchester: I guess that's fine. This means we can't open a window without waiting for its focused state, but that's probably ok.

whimboo: I don't think you want to do anything with a window in case it hasn't the focus yet, which is the case in most of the tests. So this simply eases test creation by not having to call focus() or have a wait() call each time a new window is opened.

whimboo:
I pushed an update which removes this extra unit test, given that it has no value to us at the moment. Also I renamed the parameter to `expect_focus` as @chmanchester mentioned on IRC.

Travis has currently problems with OS X builds. Given that the Linux build was successful, and also all former builds for the API change were passing on OS X, I'm going to land this PR now.

PR #119, opened 2015-03-06
Bug 1125732 - Add support for AppVeyor for testing on Windows. r=chmanchester

whimboo:


whimboo:
The current version of this patch works and shows that our tests fail a lot on Windows. I will work locally to fix the failures.

whimboo:
I disabled the webhook to ensure that we don't run with a broken appveyor config for other PRs.

firefox_ui_harness/runtests.py
```diff
@@ -68,9 +68,6 @@ def startTestRunner(runner_class, options, tests):
         if options.installer:
             installer = os.path.realpath(options.installer)
 
-            if not mozinstall.is_installer(installer):
```
chmanchester: If this method doesn't do what we expect on windows can we guard this block based on the detected platform rather than skip this validation on all platforms?

whimboo: Well, I missed that this method is already called inside of `mozinstall.install()`. Sorry for not mentioning it.

PR #120, opened 2015-03-09
Bug 1132651 - Convert Mozmill testGreenLarry

galgeek:
https://bugzilla.mozilla.org/show_bug.cgi?id=1132651

### DIFF
--- a/firefox_ui_tests/remote/security/test_ev_certificate.py
+++ b/firefox_ui_tests/remote/security/test_ev_certificate.py
@@ -1,0 +1,117 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from marionette_driver.errors import NoSuchElementException
+
+from firefox_ui_harness.decorators import skip_if_e10s, skip_under_xvfb
+from firefox_ui_harness.testcase import FirefoxTestCase
+
+
+class TestEVCertificate(FirefoxTestCase):
+
+    def setUp(self):
+        FirefoxTestCase.setUp(self)
+
+        self.url = 'https://ssl-ev.mozqa.com/'
+
+        self.identity_popup = self.browser.navbar.locationbar.identity_popup
+
+    def tearDown(self):
+        try:
+            self.identity_popup.close(force=True)
+            self.windows.close_all([self.browser])
+        except NoSuchElementException:
+            # TODO: A NoSuchElementException may be thrown here when the test is skipped
+            # as under xvfb.
+            pass
+        finally:
+            FirefoxTestCase.tearDown(self)
+
+    @skip_if_e10s
+    @skip_under_xvfb
+    def test_ev_certificate(self):
+        with self.marionette.using_context('content'):
+            self.marionette.navigate(self.url)
+
+        # Get the information from the certificate
+        cert = self.browser.tabbar.selected_tab.certificate
+        address = self.security.get_address_from_certificate(cert)
+
+        # Check the identity popup label displays
+        self.assertEqual(self.identity_popup.organization_label.get_attribute('value'),
+                         cert['organization'])
+        self.assertEqual(self.identity_popup.country_label.get_attribute('value'),
+                         '(' + address['country'] + ')')
+
+        # Check the favicon
+        # TODO: find a better way to check, e.g., mozmill's isDisplayed
+        favicon = self.browser.navbar.locationbar.favicon
+        self.wait_for_condition(lambda _: favicon.get_attribute('hidden') == 'false')
+
+        # Check the identity popup box
+        self.assertEqual(self.identity_popup.box.get_attribute('className'),
+                         'verifiedIdentity')
+
+        self.identity_popup.box.click()
+        self.wait_for_condition(lambda _: self.identity_popup.is_open)
+
+        # Check the idenity popup doorhanger
+        self.assertEqual(self.identity_popup.popup.get_attribute('className'),
+                         'verifiedIdentity')
+
+        # Check that the lock icon is visible
+        self.assertNotEqual(self.identity_popup.encryption_icon.
+                            value_of_css_property('list-style-image'), 'none')
+
+        # Check the certificate's domain name
+        # Bug 443116 - Larry strips the 'www.' from the host name using the eTLDservice.
+        # This is expected behavior.
+        self.assertEqual(self.identity_popup.host.get_attribute('textContent'),
+                         self.security.get_domain_from_common_name(cert['commonName']))
+
+        # Check the organization name
+        self.assertEqual(self.identity_popup.owner.get_attribute('textContent'),
+                         cert['organization'])
+
+        # Check the owner location string
+        # More information:
+        # hg.mozilla.org/mozilla-central/file/eab4a81e4457/browser/base/content/browser.js#l7012
+        location = self.browser.get_property('identity.identified.state_and_country')
+        location = location.replace('%S', address['state'], 1).replace('%S', address['country'])
+        location = address['city'] + '\n' + location
+        self.assertEqual(self.identity_popup.owner_location.get_attribute('textContent'),
+                         location)
+
+        # Check the verifier
+        l10n_verifier = self.browser.get_property('identity.identified.verifier')
+        l10n_verifier = l10n_verifier.replace('%S', cert['issuerOrganization'])
+        self.assertEqual(self.identity_popup.verifier.get_attribute('textContent'),
+                         l10n_verifier)
+
+        # Check the encryption label
+        self.assertEqual(self.identity_popup.encryption_label.get_attribute('textContent'),
+                         self.browser.get_property('identity.encrypted2'))
+
+        # Open the Page Info window by clicking the More Information button
+        page_info = self.browser.open_page_info_window(
+            lambda _: self.identity_popup.more_info_button.click())
+
+        try:
+            # Verify that the current panel is the security panel
+            self.assertEqual(page_info.deck.selected_panel, page_info.deck.security)
+
+            # Verify the domain listed on the security panel
+            self.assertIn(cert['commonName'],
+                          page_info.deck.security.domain.get_attribute('value'))
+
+            # Verify the owner listed on the security panel
+            self.assertEqual(page_info.deck.security.owner.get_attribute('value'),
+                             cert['organization'])
+
+            # Verify the verifier listed on the security panel
+            self.assertEqual(page_info.deck.security.verifier.get_attribute('value'),
+                             cert['issuerOrganization'])
+        finally:
+            page_info.close()
+            self.browser.focus()
--- a/firefox_puppeteer/api/security.py
+++ b/firefox_puppeteer/api/security.py
@@ -20,7 +20,7 @@ class Security(BaseLib):
         with the following entries: `city`, `country`, `postal_code`, `state`, `street`.
 
         :param certificate: A JSON object representing the certificate, which can usually be
-         retrieved via the current tab: `self.browser.tabbar.selectedTab.certificate`.
+         retrieved via the current tab: `self.browser.tabbar.selected_tab.certificate`.
 
         :returns: Address details as dictionary
         """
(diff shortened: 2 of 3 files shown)


firefox_ui_tests/remote/security/test_green_larry.py
```diff
@@ -0,0 +1,107 @@
...
+            self.windows.close_all([self.browser])
+        finally:
+            FirefoxTestCase.tearDown(self)
+
+    def test_green_larry(self):
+        with self.marionette.using_context('content'):
+            self.marionette.navigate(self.url)
+
+        # Get the information from the certificate
+        cert = self.browser.tabbar.tabs[0].certificate
```
whimboo: Please use `.selectedTab` here.

firefox_ui_tests/remote/security/test_green_larry.py
```diff
@@ -0,0 +1,107 @@
...
+        address = self.security.get_address_from_certificate(cert)
+
+        # Check the identity popup label displays
+        org_label = self.identity_popup.organization_label
+        country_label = self.identity_popup.country_label
+        self.assertEqual(org_label.get_attribute('value'), cert['organization'])
```
whimboo: You don't need the extra variables when you split the arguments into two lines, first line the current value and second line the expected value.

firefox_ui_tests/remote/security/test_green_larry.py
```diff
@@ -0,0 +1,107 @@
...
+        self.assertEqual(country_label.get_attribute('value'), '(' + address['country'] + ')')
+
+        # Check the favicon
+        favicon = self.browser.navbar.locationbar.favicon
+        self.wait_for_condition(lambda _: favicon.get_attribute('hidden') == 'false')
```
whimboo: I assume `is_displayed` doesn't work here neither? We may want to flag it with a TODO.

firefox_ui_tests/remote/security/test_green_larry.py
```diff
@@ -0,0 +1,107 @@
...
+
+        # Check the identity popup box
+        identity_box = self.identity_popup.box
+        self.assertEqual(identity_box.get_attribute('className'), 'verifiedIdentity')
```
whimboo: Same here as above for the split into two lines.

firefox_ui_tests/remote/security/test_green_larry.py
```diff
@@ -0,0 +1,107 @@
...
+
+        identity_box.click()
+
+        # Check the idenity popup doorhanger
+        doorhanger = self.identity_popup.popup
+        self.assertEqual(doorhanger.get_attribute('className'), 'verifiedIdentity')
+
+        # TODO: uncomment code when bug 1134872 lands
```
whimboo: Thanks for that fix! Looks like we can uncomment that now.

firefox_ui_tests/remote/security/test_green_larry.py
```diff
@@ -0,0 +1,107 @@
...
+
+        # Check the favicon
+        favicon = self.browser.navbar.locationbar.favicon
+        self.wait_for_condition(lambda _: favicon.get_attribute('hidden') == 'false')
+
+        # Check the identity popup box
+        identity_box = self.identity_popup.box
+        self.assertEqual(identity_box.get_attribute('className'), 'verifiedIdentity')
+
+        identity_box.click()
```
whimboo: You have to wait for the popup to be open.

firefox_ui_tests/remote/security/test_green_larry.py
```diff
@@ -0,0 +1,107 @@
...
+        # css_info_lock_icon = lock_icon.value_of_css_property('list-style-image')
+        # self.assertNotEqual(css_info_lock_icon, 'none')
+
+        # Check the host name
+        # Bug 443116 - Larry strips the 'www.' from the host name using the eTLDservice.
+        # This is expedted behavior.
+        host = self.identity_popup.host
+        common_name = self.marionette.execute_script("""
+          return Services.eTLD.getBaseDomainFromHost(arguments[0]);
+        """, script_args=[cert['commonName']])
```
whimboo: Please add a method to the Security class, which returns the domain. Keep in mind that [it is recommended](https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XPCOM/Reference/Interface/nsIEffectiveTLDService#getBaseDomainFromHost%28%29) to use `getBaseDomain()`.

firefox_ui_tests/remote/security/test_green_larry.py
```diff
@@ -0,0 +1,107 @@
...
+
+        # Check the organization name
+        owner = self.identity_popup.owner
+        self.assertEqual(owner.get_attribute('textContent'), cert['organization'])
+
+        # Check the owner location string 
+        owner_location = self.identity_popup.owner_location
+        location = self.browser.get_property('identity.identified.state_and_country')
+        location = location.replace('%S', address['state'], 1).replace('%S', address['country'])
+        location = address['city'] + '\n' + location
```
whimboo: Can you please add a @see link to https://hg.mozilla.org/mozilla-central/file/eab4a81e4457/browser/base/content/browser.js#l7012 ? It would be good to get this explained. Also in case of changes in that area.

firefox_ui_tests/remote/security/test_green_larry.py
```diff
@@ -0,0 +1,107 @@
...
+        self.assertEqual(owner_location.get_attribute('textContent'), location)
+
+        # Check the verifier
+        verifier = self.identity_popup.verifier
+        l10n_verifier = self.browser.get_property('identity.identified.verifier')
+        l10n_verifier = l10n_verifier.replace('%S', cert['issuerOrganization'])
+        self.assertEqual(verifier.get_attribute('textContent'), l10n_verifier)
```
whimboo: Same here and for the following. Get rid of the extra variables as much as possible.

firefox_ui_tests/remote/security/test_green_larry.py
```diff
@@ -0,0 +1,107 @@
...
+        # Open the Page Info window by clicking the More Information button
+        page_info = self.browser.open_page_info_window(
+            lambda _: self.identity_popup.more_info_button.click())
+
+        # Verify that the current panel is the security panel
+        self.assertEqual(page_info.deck.selected_panel, page_info.deck.security)
+
+        # Verify the domain listed on the security panel
+        domain = page_info.deck.security.domain
+        self.assertIn(cert['commonName'], domain.get_attribute('value'))
```
whimboo: The cert contains the expected value. Use it as second argument.

firefox_ui_tests/remote/security/test_green_larry.py
```diff
@@ -0,0 +1,107 @@
...
+    def setUp(self):
+        FirefoxTestCase.setUp(self)
+
+        self.url = 'https://ssl-ev.mozqa.com/'
+
+        self.identity_popup = self.browser.navbar.locationbar.identity_popup
+
+    def tearDown(self):
+        try:
+            self.windows.close_all([self.browser])
```
whimboo: You also have to ensure the identity popup is closed. Otherwise it will cause focus issues for following tests.

firefox_ui_tests/remote/security/test_green_larry.py
```diff
@@ -0,0 +1,109 @@
...
+        favicon = self.browser.navbar.locationbar.favicon
+        self.wait_for_condition(lambda _: favicon.get_attribute('hidden') == 'false')
+
+        # Check the identity popup box
+        self.assertEqual(self.identity_popup.box.get_attribute('className'),
+                         'verifiedIdentity')
+
+        self.identity_popup.box.click()
+        self.wait_for_condition(expected.element_present(
+                                lambda _: self.identity_popup.popup))
```
whimboo: This doesn't look right to me. We are waiting for the popup to be open, and not that it exists. This wait will return immediately.

firefox_ui_tests/remote/security/test_green_larry.py
```diff
@@ -0,0 +1,109 @@
...
+        # This is expedted behavior.
+        self.assertEqual(self.identity_popup.host.get_attribute('textContent'),
+                         self.security.get_domain_from_common_name(cert['commonName']))
+
+        # Check the organization name
+        self.assertEqual(self.identity_popup.owner.get_attribute('textContent'),
+                         cert['organization'])
+
+        # Check the owner location string
+        # More informaiton:
```
whimboo: nit: spelling mistake

firefox_ui_tests/remote/security/test_green_larry.py
```diff
@@ -0,0 +1,109 @@
...
+    def setUp(self):
+        FirefoxTestCase.setUp(self)
+
+        self.url = 'https://ssl-ev.mozqa.com/'
+
+        self.identity_popup = self.browser.navbar.locationbar.identity_popup
+
+    def tearDown(self):
+        try:
+            # TODO: PR 105/115: insure identity_popup closes
```
whimboo: For a final review lets wait for its landing. It should be done by today. If we don't hear back from both, mind taking this other PR and finish it up?

galgeek: I'll look for that update tomorrow morning. 

I should be able to finish it up if others don't get to it.

firefox_ui_tests/remote/security/test_green_larry.py
```diff
@@ -0,0 +1,115 @@
...
+
+        self.identity_popup = self.browser.navbar.locationbar.identity_popup
+
+    def tearDown(self):
+        try:
+            self.identity_popup.close(force=True)
+            self.windows.close_all([self.browser])
+        except NoSuchElementException:
+            # TODO: A NoSuchElementException is thrown here when the identity popup
+            # closes earlier, apparently when the page info window opens successfully
```
whimboo: You can remove the reference to the page info window. It's not related here given that it will also fail when the test method is skipped.

whimboo: But if its also happening when the page info window is closed, this is more related to the current window being active. Keep in mind that closing page info will not make the browser active automatically. I assume calling `self.browser.focus()` fix this problem?

whimboo: If that is the case you should always use the full access line of elements like `self.browser.navbar...` because it will auto-focus the window when `.navbar` is returned. Once bug 1128968 is fixed, we could also call `self.window.focus()` in subclasses without having to go through the whole chain again.

firefox_ui_tests/remote/security/test_green_larry.py
```diff
@@ -0,0 +1,117 @@
...
+        self.assertEqual(self.identity_popup.encryption_label.get_attribute('textContent'),
+                         self.browser.get_property('identity.encrypted2'))
+
+        # Open the Page Info window by clicking the More Information button
+        page_info = self.browser.open_page_info_window(
+            lambda _: self.identity_popup.more_info_button.click())
+
+        try:
+            # Verify that the current panel is the security panel
+            self.assertEqual(page_info.deck.selected_panel, page_info.deck.security)
```
whimboo: This fails with e10s enabled because the security panel is not selected automatically. Try to reproduce that manually. I strongly feel this is a bug in Firefox.

firefox_ui_tests/remote/security/test_green_larry.py
```diff
@@ -0,0 +1,120 @@
...
+        self.assertEqual(self.identity_popup.encryption_label.get_attribute('textContent'),
+                         self.browser.get_property('identity.encrypted2'))
+
+        # Open the Page Info window by clicking the More Information button
+        page_info = self.browser.open_page_info_window(
+            lambda _: self.identity_popup.more_info_button.click())
+
+        try:
+            # TODO: Bug 1144493 workaround to make sure Page Info's security tab is selected
+            page_info.deck.select(page_info.deck.security)
```
whimboo: No workaround please, as what I told on IRC yesterday. So please skip the test for e10s.

firefox_ui_tests/remote/security/test_green_larry.py
```diff
@@ -0,0 +1,120 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from marionette_driver.errors import NoSuchElementException
+
+from firefox_ui_harness.decorators import skip_under_xvfb
+from firefox_ui_harness.testcase import FirefoxTestCase
+
+
+class TestGreenLarry(FirefoxTestCase):
```
whimboo: Please rename the test to something more understandable. I would suggest `test_ev_certificate.py`

firefox_ui_tests/remote/security/test_ev_certificate.py
```diff
@@ -0,0 +1,118 @@
...
+            self.assertEqual(page_info.deck.security.owner.get_attribute('value'),
+                             cert['organization'])
+
+            # Verify the verifier listed on the security panel
+            self.assertEqual(page_info.deck.security.verifier.get_attribute('value'),
+                             cert['issuerOrganization'])
+
+            page_info.close()
+        finally:
+            self.browser.focus()
```
whimboo: This will not close the page info window.  You surely want to move it into finally too, just before this line.

PR #122, opened 2015-03-09
Bug 1129843 - Convert update tests to Marionette

whimboo:
Currently a WIP which will get frequent updates over the next days. Nothing to review yet.

### DIFF
--- a/firefox_ui_harness/__init__.py
+++ b/firefox_ui_harness/__init__.py
@@ -2,4 +2,12 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-from runtests import cli
+from runtests import (cli,
+                      cli_update,
+                      )
+
+from testcases import (UpdateTestCase,
+                       )
+
+# Bug 1145638 To be moved to testcases
+from testcase import FirefoxTestCase
--- a/firefox_ui_harness/options/update.py
+++ b/firefox_ui_harness/options/update.py
@@ -1,0 +1,63 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from base import FirefoxUIOptions
+
+
+class UpdateOptions(FirefoxUIOptions):
+    def __init__(self, **kwargs):
+        FirefoxUIOptions.__init__(self, **kwargs)
+
+        # Inheriting object must call this __init__ to set up option handling
+        group = self.add_option_group('Update Tests')
+        group.add_option('--update-allow-mar-channel',
+                         dest='update_mar_channels',
+                         default=[],
+                         action='append',
+                         metavar='MAR_CHANNEL',
+                         help='Additional MAR channel to be allowed for updates, '
+                              'e.g. "firefox-mozilla-beta" for updating a release '
+                              'build to the latest beta build.')
+        group.add_option('--update-channel',
+                         dest='update_channel',
+                         metavar='CHANNEL',
+                         help='Channel to use for the update check.')
+        group.add_option('--update-direct-only',
+                         dest='update_direct_only',
+                         default=False,
+                         action='store_true',
+                         help='Only perform a direct update')
+        group.add_option('--update-fallback-only',
+                         dest='update_fallback_only',
+                         default=False,
+                         action='store_true',
+                         help='Only perform a fallback update')
+        group.add_option('--update-override-url',
+                         dest='update_override_url',
+                         metavar='URL',
+                         help='Force specified URL to use for update checks.')
+        group.add_option('--update-target-version',
+                         dest='update_target_version',
+                         metavar='VERSION',
+                         help='Version of the updated build.')
+        group.add_option('--update-target-buildid',
+                         dest='update_target_buildid',
+                         metavar='BUILD_ID',
+                         help='Build ID of the updated build.')
+
+    def parse_args(self, *args, **kwargs):
+        options, tests = FirefoxUIOptions.parse_args(self, *args, **kwargs)
+
+        return (options, tests)
+
+    def verify_usage(self, options, tests):
+        if options.update_direct_only and options.update_fallback_only:
+            self.error('Options --update-direct-only and --update-fallback-only '
+                       'are mutually exclusive.')
+
+        if options.update_target_buildid and not options.update_target_version:
+            self.error('Option --update-target-buildid also expects '
+                       '--update-target-version to be set.')
+
+        FirefoxUIOptions.verify_usage(self, options, tests)
--- a/firefox_ui_harness/runners/base.py
+++ b/firefox_ui_harness/runners/base.py
@@ -2,7 +2,16 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-default_prefs = {
+import os
+
+import mozinfo
+from marionette import BaseMarionetteTestRunner
+
+import firefox_ui_tests
+from ..testcase import FirefoxTestCase
+
+
+DEFAULT_PREFS = {
     'app.update.auto': False,
     'app.update.enabled': False,
     'browser.dom.window.dump.enabled': True,
@@ -17,7 +26,6 @@ default_prefs = {
     'browser.shell.checkDefaultBrowser': False,
     'browser.startup.page': 0,
     'browser.tabs.animate': False,
-    'browser.tabs.remote.autostart': False,
     'browser.tabs.warnOnClose': False,
     'browser.tabs.warnOnOpen': False,
     'browser.uitour.enabled': False,
@@ -41,9 +49,43 @@ default_prefs = {
     'focusmanager.testmode': True,
     'geo.provider.testing': True,
     'javascript.options.showInConsole': True,
+    'marionette.logging': False,
     'security.notification_enable_delay': 0,
     'signon.rememberSignons': False,
     'startup.homepage_welcome_url': 'about:blank',
     'toolkit.startup.max_resumed_crashes': -1,
     'toolkit.telemetry.enabled': False,
 }
+
+
+class FirefoxUITestRunner(BaseMarionetteTestRunner):
+    def __init__(self, **kwargs):
+        BaseMarionetteTestRunner.__init__(self, **kwargs)
+
+        if not self.server_root:
+            self.server_root = firefox_ui_tests.resources
+
+        # Bug 1146847: This needs a refactoring given that our default
+        # preferences are not coming from the command line
+        self.prefs.update(DEFAULT_PREFS)
+
+        if not kwargs.get('e10s'):
+            self.prefs.update({'browser.tabs.remote.autostart': False})
+
+        self.test_handlers = [FirefoxTestCase]
+
+    def get_application_folder(self, binary):
+        """Returns the directory of the application."""
+        if mozinfo.isMac:
+            end_index = binary.find('.app') + 4
+            return binary[:end_index]
+        else:
+            return os.path.dirname(binary)
+
+    def run_tests(self, tests):
+        # Ensure Marionette is always reset before starting tests
+        # This might need support in Marionette base
+        self.marionette = None
+        self.tests = []
+
+        BaseMarionetteTestRunner.run_tests(self, tests)
--- a/firefox_ui_harness/runners/update.py
+++ b/firefox_ui_harness/runners/update.py
@@ -1,0 +1,97 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import shutil
+import sys
+import tempfile
+
+import mozfile
+import mozinstall
+
+from base import FirefoxUITestRunner
+from ..testcases import UpdateTestCase
+
+import firefox_ui_tests
+
+
+DEFAULT_PREFS = {
+    'app.update.log': True,
+}
+
+
+class UpdateTestRunner(FirefoxUITestRunner):
+
+    def __init__(self, **kwargs):
+        FirefoxUITestRunner.__init__(self, **kwargs)
+
+        self.original_bin = self.bin
+
+        self.prefs.update(DEFAULT_PREFS)
+
+        # In case of overriding the update URL, set the appropriate preference
+        override_url = kwargs.pop('update_override_url', None)
+        if override_url:
+            self.prefs.update({'app.update.url.override': override_url})
+
+        self.run_direct_update = not kwargs.pop('update_fallback_only', False)
+        self.run_fallback_update = not kwargs.pop('update_direct_only', False)
+
+        self.test_handlers = [UpdateTestCase]
+
+    def duplicate_application(self, application_folder):
+        """Creates a copy of the specified binary."""
+        target_folder = tempfile.mkdtemp('.binary-update-tests')
+
+        self.logger.info('Creating a copy of the application at "%s".' % target_folder)
+        mozfile.remove(target_folder)
+        shutil.copytree(application_folder, target_folder)
+
+        return target_folder
+
+    def run_tests(self, tests):
+        # Used to store the last occurred exception because we execute
+        # run_tests() multiple times
+        self.exc_info = None
+
+        failed = 0
+        source_folder = self.get_application_folder(self.original_bin)
+
+        def _run_tests(manifest):
+            target_folder = None
+
+            try:
+                target_folder = self.duplicate_application(source_folder)
+                self.bin = mozinstall.get_binary(target_folder, 'Firefox')
+
+                FirefoxUITestRunner.run_tests(self, [manifest])
+
+            except Exception:
+                self.exc_info = sys.exc_info()
+                self.logger.error('Failure during execution of the update test.',
+                                  exc_info=self.exc_info)
+
+            finally:
+                self.logger.info('Removing copy of the application at "%s"' % target_folder)
+                try:
+                    mozfile.remove(target_folder)
+                except IOError as e:
+                    self.logger.error('Cannot remove copy of application: "%s"' % str(e))
+
+        # Run direct update tests if wanted
+        if self.run_direct_update:
+            _run_tests(manifest=firefox_ui_tests.manifest_update_direct)
+            failed += self.failed
+
+        # Run fallback update tests if wanted
+        if self.run_fallback_update:
+            _run_tests(manifest=firefox_ui_tests.manifest_update_fallback)
+            failed += self.failed
+
+        # Combine failed tests for all run_test() executions
+        self.failed = failed
+
+        # If exceptions happened, re-throw the last one
+        if self.exc_info:
+            ex_type, exception, tb = self.exc_info
+            raise ex_type, exception, tb
--- a/firefox_ui_harness/runtests.py
+++ b/firefox_ui_harness/runtests.py
@@ -2,62 +2,21 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-import copy
 import os
 import sys
 import tempfile
 
 import mozinstall
 
-from marionette import (BaseMarionetteOptions,
-                        BaseMarionetteTestRunner,)
 from mozlog import structured
 
-import firefox_puppeteer
-import firefox_ui_tests
+from firefox_ui_harness.options import (FirefoxUIOptions,
+                                        UpdateOptions,
+                                        )
 
-from .default_prefs import default_prefs
-from .testcase import FirefoxTestCase
-
-
-class FirefoxUIOptions(BaseMarionetteOptions):
-
-    def __init__(self, **kwargs):
-        BaseMarionetteOptions.__init__(self, **kwargs)
-
-        self.add_option('--installer',
-                        dest='installer',
-                        action='store',
-                        help='installer of a Gecko application to use for running the tests')
-
-    def parse_args(self, *args, **kwargs):
-        options, test_files = BaseMarionetteOptions.parse_args(self, *args, **kwargs)
-
-        # It is not allowed to specify both options for binary and installer
-        if options.installer:
-            if options.binary:
-                self.error('options --binary and --installer are mutually exclusive')
-
-            # Spoofing so verify_usage() is not showing a failure
-            options.binary = 'to_be_set'
-
-        if not test_files:
-            test_files = [firefox_puppeteer.manifest, firefox_ui_tests.manifest]
-        return (options, test_files)
-
-
-class FirefoxUITestRunner(BaseMarionetteTestRunner):
-    def __init__(self, *args, **kwargs):
-        if not kwargs.get('server_root'):
-            kwargs['server_root'] = firefox_ui_tests.resources
-
-        prefs = kwargs.get('prefs', {})
-        runner_prefs = copy.deepcopy(default_prefs)
-        runner_prefs.update(prefs)
-        kwargs['prefs'] = runner_prefs
-
-        BaseMarionetteTestRunner.__init__(self, *args, **kwargs)
-        self.test_handlers = [FirefoxTestCase]
+from firefox_ui_harness.runners import (FirefoxUITestRunner,
+                                        UpdateTestRunner,
+                                        )
 
 
 def startTestRunner(runner_class, options, tests):
@@ -69,8 +28,8 @@ def startTestRunner(runner_class, options, tests):
             installer = os.path.realpath(options.installer)
 
             dest_folder = tempfile.mkdtemp()
-            options.logger.info('Installing build "%s" to "%s"...' % (installer,
-                                                                      dest_folder))
+            options.logger.info('Installing application "%s" to "%s"' % (installer,
+                                                                         dest_folder))
             install_folder = mozinstall.install(installer, dest_folder)
             options.binary = mozinstall.get_binary(install_folder, 'firefox')
 
@@ -80,7 +39,7 @@ def startTestRunner(runner_class, options, tests):
     finally:
         # Ensure to uninstall the binary if it has been installed before
         if install_folder and os.path.exists(install_folder):
-            options.logger.info('Uninstalling build at "%s"...' % install_folder)
+            options.logger.info('Uninstalling application at "%s"' % install_folder)
             mozinstall.uninstall(install_folder)
 
     return runner
@@ -90,6 +49,7 @@ def cli(runner_class=FirefoxUITestRunner, parser_class=FirefoxUIOptions):
     parser = parser_class(usage='%prog [options] test_file_or_dir <test_file_or_dir> ...')
     structured.commandline.add_logging_group(parser)
     options, tests = parser.parse_args()
+
     parser.verify_usage(options, tests)
 
     logger = structured.commandline.setup_logging(
@@ -97,9 +57,19 @@ def cli(runner_class=FirefoxUITestRunner, parser_class=FirefoxUIOptions):
 
     options.logger = logger
 
-    runner = startTestRunner(runner_class, options, tests)
-    if runner.failed > 0:
-        sys.exit(10)
+    try:
+        runner = startTestRunner(runner_class, options, tests)
+        if runner.failed > 0:
+            sys.exit(10)
+
+    except Exception:
+        logger.error('Failure during execution of the update test.',
+                     exc_info=True)
+        sys.exit(1)
+
+
+def cli_update():
+    cli(runner_class=UpdateTestRunner, parser_class=UpdateOptions)
 
 
 if __name__ == '__main__':
--- a/firefox_ui_harness/testcase.py
+++ b/firefox_ui_harness/testcase.py
@@ -5,15 +5,30 @@
 from marionette import MarionetteTestCase
 
 from firefox_puppeteer import Puppeteer
+from firefox_puppeteer.ui.windows import BrowserWindow
 
 
 class FirefoxTestCase(MarionetteTestCase, Puppeteer):
-    """
-    Test case that inherits from a Puppeteer object so Firefox specific
-    libraries are exposed to test scope.
+    """Base testcase class for Firefox Desktop tests.
+
+    It enhances the Marionette testcase by inserting the Puppeteer mixin class,
+    so Firefox specific API modules are exposed to test scope.
     """
     def __init__(self, *args, **kwargs):
         MarionetteTestCase.__init__(self, *args, **kwargs)
+
+    def restart(self, flags=None):
+        """Restart Firefox and re-initialize data.
+
+        :param flags: Specific restart flags for Firefox
+        """
+        self.marionette.restart(in_app=True)
+
+        # Marionette doesn't keep the former context, so restore to chrome
+        self.marionette.set_context('chrome')
+
+        # Ensure that we always have a valid browser instance available
+        self.browser = self.windows.switch_to(lambda win: type(win) is BrowserWindow)
 
     def setUp(self, *args, **kwargs):
         MarionetteTestCase.setUp(self, *args, **kwargs)
@@ -36,6 +51,7 @@ class FirefoxTestCase(MarionetteTestCase, Puppeteer):
             # browser window which has at least one open tab
             # TODO: We might have to make this more error prone in case the
             # original window has been closed.
+            self.browser.focus()
             self.browser.tabbar.tabs[0].switch_to()
 
             self.prefs.restore_all_prefs()
--- a/firefox_ui_harness/testcases/__init__.py
+++ b/firefox_ui_harness/testcases/__init__.py
@@ -1,0 +1,5 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from update import UpdateTestCase
--- a/firefox_ui_harness/testcases/update.py
+++ b/firefox_ui_harness/testcases/update.py
@@ -1,0 +1,234 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from ..testcase import FirefoxTestCase
+
+from firefox_puppeteer.api.software_update import SoftwareUpdate
+from firefox_puppeteer.ui.update_wizard import UpdateWizardDialog
+
+
+class UpdateTestCase(FirefoxTestCase):
+
+    def __init__(self, *args, **kwargs):
+        FirefoxTestCase.__init__(self, *args, **kwargs)
+
+        self.target_buildid = kwargs.pop('update_target_buildid')
+        self.target_version = kwargs.pop('update_target_version')
+
+        self.update_channel = kwargs.pop('update_channel')
+        self.default_update_channel = None
+
+        self.update_mar_channels = set(kwargs.pop('update_mar_channels'))
+        self.default_mar_channels = None
+
+        self.updates = []
+
+    def setUp(self, is_fallback=False):
+        FirefoxTestCase.setUp(self)
+        self.software_update = SoftwareUpdate(lambda: self.marionette)
+
+        # Bug 604364 - Preparation to test multiple update steps
+        self.current_update_index = 0
+
+        self.staging_directory = self.software_update.staging_directory
+
+        # If requested modify the default update channel. It will be active
+        # after the next restart of the application
+        # Bug 1142805 - Modify file via Python directly
+        if self.update_channel:
+            # Backup the original content and the path of the channel-prefs.js file
+            self.default_update_channel = {
+                'content': self.software_update.update_channel.file_contents,
+                'path': self.software_update.update_channel.file_path,
+            }
+            self.software_update.update_channel.default_channel = self.update_channel
+
+        # If requested modify the list of allowed MAR channels
+        # Bug 1142805 - Modify file via Python directly
+        if self.update_mar_channels:
+            # Backup the original content and the path of the update-settings.ini file
+            self.default_mar_channels = {
+                'content': self.software_update.mar_channels.config_file_contents,
+                'path': self.software_update.mar_channels.config_file_path,
+            }
+            self.software_update.mar_channels.add_channels(self.update_mar_channels)
+
+        # Bug 1142805 - Until we don't modify the channel-prefs.js and update-settings.ini
+        # files before Firefox gets started, a restart of Firefox is necessary to
+        # accept the new update channel.
+        self.restart()
+
+        # Dictionary which holds the information for each update
+        self.updates = [{
+            'build_pre': self.software_update.build_info,
+            'build_post': None,
+            'fallback': is_fallback,
+            'patch': {},
+            'success': False,
+        }]
+
+        self.assertEqual(self.software_update.update_channel.default_channel,
+                         self.software_update.update_channel.channel)
+
+        self.assertTrue(self.update_mar_channels.issubset(
+                        self.software_update.mar_channels.channels))
+
+        # Check if the user has permissions to run the update
+        self.assertTrue(self.software_update.allowed)
+
+    def assert_update_applied(self, update_status):
+        """Checks if an update has been applied correctly.
+
+        :param update_status: All the data collected during the update process
+        """
+        # Get the information from the last update
+        info = update_status[self.current_update_index]
+
+        # The upgraded version should be identical with the version given by
+        # the update and we shouldn't have run a downgrade
+        check = self.marionette.execute_script("""
+          Components.utils.import("resource://gre/modules/Services.jsm");
+
+          return  Services.vc.compare(arguments[0], arguments[1]);
+        """, script_args=[info['build_post']['version'], info['build_pre']['version']])
+
+        self.assertGreaterEqual(check, 0, 'The version of the upgraded build is higher or equal')
+
+        # If a target version has been specified, check if it matches the updated build
+        if self.target_version:
+            self.assertEqual(info['build_post']['version'], self.target_version)
+
+        # The post buildid should be identical with the buildid contained in the patch
+        self.assertEqual(info['build_post']['buildid'], info['patch']['buildid'])
+
+        # If a target buildid has been specified, check if it matches the updated build
+        if self.target_buildid:
+            self.assertEqual(info['build_post']['buildid'], self.target_buildid)
+
+        # An upgrade should not change the builds locale
+        self.assertEqual(info['build_post']['locale'], info['build_pre']['locale'])
+
+        # Check that no application-wide add-ons have been disabled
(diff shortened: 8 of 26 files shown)


firefox_ui_harness/__init__.py
```diff
@@ -2,4 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-from runtests import cli
+from mixins import *
```
chmanchester: I'm not a fan of wildcard imports (or mixins), but I'll keep reading.

firefox_ui_harness/mixins/base.py
```diff
@@ -39,9 +49,54 @@
...
+        BaseMarionetteTestRunner.__init__(self, **kwargs)
+
+        if not self.server_root:
+            self.server_root = firefox_ui_tests.resources
+
+        # Set default preferences for Firefox UI tests
+        # Bug 1141129 - Workaround necessary given that self.prefs defaults to None
+        prefs = self.prefs or {}
+        prefs.update(DEFAULT_PREFS)
+        self.prefs = prefs
```
chmanchester: If you look at BaseMarionetteTestRunner you'll see that effectively the only pref set in the runner is e10s, and only when a command line option specifies it. So long as DEFAULT_PREFS turns off e10s, this means my command line options isn't honored now.

firefox_ui_harness/runtests.py
```diff
@@ -87,9 +44,10 @@ def startTestRunner(runner_class, options, tests):
 
 
 def cli(runner_class=FirefoxUITestRunner, parser_class=FirefoxUIOptions):
-    parser = parser_class(usage='%prog [options] test_file_or_dir <test_file_or_dir> ...')
+    parser = parser_class(usage='%prog [options] manifest <test_file_or_dir> ...')
```
chmanchester: This looks wrong (it was wrong before).

firefox_ui_harness/runtests.py
```diff
@@ -102,5 +60,9 @@ def cli(runner_class=FirefoxUITestRunner, parser_class=FirefoxUIOptions):
         sys.exit(10)
 
 
+def cli_update():
```
chmanchester: I would ditch this wrapper function.

firefox_ui_harness/runtests.py
```diff
@@ -102,5 +60,9 @@ def cli(runner_class=FirefoxUITestRunner, parser_class=FirefoxUIOptions):
...
+    cli(runner_class=UpdateTestRunner, parser_class=UpdateOptions)
+
+
 if __name__ == '__main__':
     sys.exit(cli())
```
chmanchester: either cli(UpdateTestRunner, UpdateOptions) or cli_update() here, I expect.

firefox_ui_harness/mixins/updates.py
```diff
@@ -0,0 +1,80 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from base import FirefoxUIOptions, FirefoxUITestRunner
+from ..firefox_ui_test import FirefoxUITestCase
+
+import firefox_ui_tests
+
+
+class UpdateOptions(FirefoxUIOptions):
```
chmanchester: This is a linear inheritance hierarchy. Calling these mixins doesn't make sense to me.

firefox_ui_tests/remote/manifest.ini
```diff
@@ -1 +1,4 @@
+[DEFAULT]
+remote = true
```
chmanchester: These types are just one per manifest right now, as long as they are let's not add them.

firefox_ui_harness/mixins/base.py
```diff
@@ -39,9 +49,54 @@
...
+        # will abort due to no binary specified. Once the bug is fixed we can move
+        # the whole block to self.base_verify_usage().
+        if options.installer:
+            if options.binary:
+                self.error('Options --binary and --installer are mutually exclusive.')
+
+            options.binary = 'FAKED_VALUE'
+
+        if not tests:
+            tests = [firefox_puppeteer.manifest, firefox_ui_tests.manifest_all]
```
chmanchester: I need to look at this, but do we always want to run the update tests by default?

firefox_ui_tests/update/test_update.py
```diff
@@ -0,0 +1,73 @@
...
+            # Open the about window and check for updates
+            about_window = self.browser.open_about_window()
+            update_available = about_window.check_for_updates()
+            self.assertTrue(update_available)
+
+            # Download update and wait until it has been applied
+            about_window.download()
+            about_window.wait_for_update_applied()
+
+            # Bug 1141679 - Socket is not cleaned up if in_app=True
```
chmanchester: This is fixed.

firefox_ui_tests/update/test_update.py
```diff
@@ -0,0 +1,73 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from firefox_ui_harness.testcase import FirefoxTestCase
```
chmanchester: I didn't review this file.

firefox_ui_harness/__init__.py
whimboo: Mixins no longer exist. I was using them at the beginning because it was easier to get it working by mostly copying from the gaia tests. But with different entry points it was a nightmare. I will finally update the folder when all is ready.

firefox_ui_harness/mixins/base.py
whimboo: Our intention is to force preference values for our tests. Whenever something in Marionette changes it shouldn't totally bust our tests. In this case I agree, but if we can turn on e10s soon, I can remove this specific preference from our default list.

firefox_ui_harness/runtests.py
whimboo: Can you please be more explicit? Not sure what you are referring here.

firefox_ui_harness/runtests.py
whimboo: You cannot do this. We need specific methods for every entry point, which gets added. You cannot specify the runner class and parser class on the setup.py level.

firefox_ui_harness/runtests.py
whimboo: See above. We do not only have an entry point for updates, but also for other kind of tests. This is just running the default set of tests, which will be unit, functional, and remote tests.

firefox_ui_harness/mixins/updates.py
whimboo: Where are you seeing a mixin? There are none anymore. But what I have to do is to rename `update_verify_usage` so we override `verify_usage` of the base class. This will be fixed in the next commit.

firefox_ui_tests/remote/manifest.ini
whimboo: That's a left-over when I tried to specify the tests to run via the `--type` option. Given that wasn't working and I removed this, we can also remove those entries for now. But what I would like to see are types like `security` which we could stick to functional and remote security tests, so that you can fine-control which tests to run.

firefox_ui_harness/mixins/base.py
whimboo: No we don't want to. Thanks for that hint. We should indeed not add them to the top-level manifest file. Once we have specific entry points for functional and remote tests, I would even remove that top-level manifest completely.

firefox_ui_harness/mixins/base.py
chmanchester: But I'm saying the way we use self.prefs in the runner it's something set by a user trying to run the tests, so we want to retain these. Marionette's "defaults" are in geckoinstance.py, and we overwrite them already with the old behavior.

firefox_ui_harness/runtests.py
chmanchester: The cli is actually more like: "firefox-ui-tests [options] <test_file_or_manifest>", right?

chmanchester: The cli is actually something like "firefox-ui-tests [options] manifest_or_test_file", right?

firefox_ui_harness/runtests.py
chmanchester: Oh, ok.

firefox_ui_harness/runtests.py
chmanchester: Ok, nevermind.

firefox_ui_harness/runtests.py
whimboo: The arguments can be a manifest, a test file, or a directory. So we have three options. Maybe I should simple leave this change out of this PR and file a separate bug for it.

chmanchester: As long as we're here might as well get it right.

firefox_ui_harness/testcase.py
```diff
@@ -5,23 +5,37 @@
...
+        self.marionette.set_context('chrome')
+
+        self.browser = self.windows.switch_to(lambda win: type(win) is BrowserWindow)
+
+    def restart(self, flags=None):
+        """Restart Firefox and re-initialize data.
+
+        :param flags: Specific restart flags for Firefox
+        """
+        self.marionette.restart()
```
chmanchester: Don't we need the in application initiated restart? Wasn't this a hard and urgent requirement for introducing these tests?

firefox_ui_harness/testcase.py
```diff
@@ -5,23 +5,37 @@
...
+    It enhances the Marionette testcase by inserting the Puppeteer mixin class,
+    so Firefox specific API modules are exposed to test scope.
     """
     def __init__(self, *args, **kwargs):
         MarionetteTestCase.__init__(self, *args, **kwargs)
 
+    def _init_data(self):
+        self.marionette.set_context('chrome')
+
+        self.browser = self.windows.switch_to(lambda win: type(win) is BrowserWindow)
```
chmanchester: This helper seems poorly named and I'm not really know what its purpose is. Why can't we trust a test to leave us switched in to the main browser window?

firefox_ui_harness/testcase.py
```diff
@@ -5,23 +5,37 @@
...
+
+    def restart(self, flags=None):
+        """Restart Firefox and re-initialize data.
+
+        :param flags: Specific restart flags for Firefox
+        """
+        self.marionette.restart()
+        self._init_data()
```
chmanchester: Ok, here I can see we might need to do some additional work to get switched into the correct window, but maybe not at the beginning of every test.

firefox_ui_harness/testcases/__init__.py
```diff
@@ -0,0 +1,5 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from update import *
```
chmanchester: I'm assuming the wildcard imports will be taken out in the final version.

firefox_ui_harness/runners/base.py
```diff
@@ -39,9 +48,42 @@
...
+        BaseMarionetteTestRunner.__init__(self, **kwargs)
+
+        if not self.server_root:
+            self.server_root = firefox_ui_tests.resources
+
+        # Set default preferences for Firefox UI tests
+        # Bug 1141129 - Workaround necessary given that self.prefs defaults to None
+        prefs = self.prefs or {}
+        prefs.update(DEFAULT_PREFS)
+        self.prefs = prefs
```
chmanchester: This has the same bug I mentioned before about overriding things set from the commandline, right? You just need to update the default prefs with the harness prefs instead of the other way around.

chmanchester: The old code for this handled this case fine updating doesn't seem related to the update tests at all.

firefox_ui_harness/testcase.py
whimboo: I had to remove it at some point because it was not working. I wanted to investigate it but looks like I forgot about it. I tried to re-add `in_app=True` but the tests are still failing because Firefox doesn't restart! I will get this investigated and a bug filed. Thanks for noticing that!

whimboo: Looks like the locally installed packages for driver and client were totally messed-up. When installing the client from m-c it downloads the driver from pypi but not picking it up from m-c too. I fixed that and it works now.

firefox_ui_harness/testcase.py
whimboo: We need a properly setup `self.browser` instance for our tests to work. Given that FirefoxTestCase is already doing it at the beginning of the test, we could move this code into `restart()`, so it only gets called after a restart of the application.

whimboo: Ups, sorry. That is the FirefoxTestCase class. So we should always have that line here find the right window. In this case the browser window. It will not necessarily be the focused one right now. If we don't do that here, we can fail if Firefox restarts with a set of different windows open, and the `current_chrome_window_handle` is not a browser window.

firefox_ui_harness/testcase.py
whimboo: Feel free to propose a better name for it.

firefox_ui_harness/testcases/__init__.py
whimboo: Yes, that's part of the cleanup story I mentioned yesterday. With all the code moving around I wanted to have it as easy as possible.

firefox_ui_harness/runners/base.py
whimboo: I do not think that we should let the harness overwrite our own defined preferences. Harness preferences should always have lower priority compared to preference we need for our tests. Ideally we should not modify preferences set by cli options, or at least remove such an option if it is not wanted. Let me know what you think.

chmanchester: When I pull your branch and specify "--e10s" on the commandline, e10s is not enabled. Ignoring user intervention in this fashion is a regression, I don't know what's controversial about this point.

firefox_ui_harness/testcase.py
chmanchester: Ok, so can we move this helper inline here and not add the additional check to setUp? If a test is switched into the wrong window at setUp it means the previous test has leaked a window and we're unlikely to succeed by virtue of switching back to a browser window.

firefox_puppeteer/__init__.py
```diff
@@ -86,6 +86,14 @@ def security(self):
...
         """
 
+    @use_class_as_property('api.software_update.SoftwareUpdate')
+    def software_update(self):
+        """
+        Provides an api for accessing application update related properties and helpers.
+
+        See the :class:`~api.software_update.SoftwareUpdate` reference.
+        """
+
```
chmanchester: We don't want this exposed to regular test cases, do we?

firefox_ui_harness/runners/base.py
whimboo: I think it would be better to remove the `browser.tabs.remote.autostart` pref from the array above, so we do not force it. But I stay with my point that an underlying harness should not rule the prefs to use.

firefox_ui_harness/testcase.py
whimboo: > Ok, so can we move this helper inline here and not add the additional
> check to setUp? If a test is switched into the wrong window at setUp it
> means the previous test has leaked a window and we're unlikely to
> succeed by virtue of switching back to a browser window.

The most important bits here are for restart tests, where other windows
can be open after the restart. So we cannot move that method into setUp.
If you really think we don't need this feature for setUp, we should have
it only in the restart() helper method.

firefox_puppeteer/__init__.py
whimboo: It's seldom used, so I can remove it and import when required. Sure.

firefox_ui_harness/runners/base.py
chmanchester: I pointed this out before, but in this case the "underlying harness" only ever corresponds to direct intervention by the user. As I have also said before this gets turned on and off in nightly builds, and we don't know what we might need to do to the prefs marionette provides in the future, so we need to make sure it's off as long as we need it off.

firefox_ui_harness/runners/update.py
```diff
@@ -0,0 +1,76 @@
...
+            target_folder = None
+
+            try:
+                target_folder = self.duplicate_application(source_folder)
+                self.bin = mozinstall.get_binary(target_folder, 'Firefox')
+
+                self.type = update_type
+                FirefoxUITestRunner.run_tests(self, tests)
+
+            except Exception:
```
chmanchester: Which part of this do we think might throw an exception? Just the mozinstall part? The marionette automation has a check on the runner object where if its "failed" field is greater than 0, the exit code of the application is non-zero. We should do something similar here if we're worried about exceptions to make sure harness failures always translate to the calling environment.

firefox_ui_harness/runners/update.py
```diff
@@ -0,0 +1,76 @@
...
+                # TODO: Store exception and re-raise after all tests have been run
+                self.logger.exception('Failure during execution of the update test.')
```
chmanchester: This is a mozlogger, I think this needs to be logger.error or logger.critical.

firefox_ui_harness/runners/update.py
chmanchester: Other harnesses I've worked with assume the central run_tests method doesn't itself throw. When individual test cases throw exceptions these are unexpected errors and then test failures, I suspect even here these are the cases we care about most.

firefox_ui_harness/runners/update.py
```diff
@@ -0,0 +1,76 @@
...
+        source_folder = self.get_application_folder(self.original_bin)
+
+        def _run_tests(update_type):
+            target_folder = None
+
+            try:
+                target_folder = self.duplicate_application(source_folder)
+                self.bin = mozinstall.get_binary(target_folder, 'Firefox')
+
+                self.type = update_type
```
chmanchester: This isn't used anywhere.

firefox_ui_tests/update/manifest.ini
```diff
@@ -0,0 +1,5 @@
+[test_direct_update.py]
+direct = true
```
chmanchester: You aren't consuming this key from the parsed manifest anywhere, are you?

firefox_ui_harness/testcases/update.py
```diff
@@ -0,0 +1,133 @@
...
+class UpdateTestCase(FirefoxTestCase):
+
+    def __init__(self, *args, **kwargs):
+        FirefoxTestCase.__init__(self, *args, **kwargs)
+
+        self.target_buildid = kwargs.pop('update_target_buildid')
+        self.target_version = kwargs.pop('update_target_version')
+
+        self.update_channel = kwargs.pop('update_channel')
+        self.update_mar_channels = set(kwargs.pop('update_mar_channels'))
```
chmanchester: Every time you "pop" here, it's a KeyError when that argument wasn't provided from the command line, so let's mark them as obligatory in verify_usage.

firefox_ui_harness/options/update.py
```diff
@@ -0,0 +1,67 @@
...
+        options, tests = FirefoxUIOptions.parse_args(self, *args, **kwargs)
+
+        tests = [firefox_ui_tests.manifest_update]
+
+        return (options, tests)
+
+    def verify_usage(self, options, tests):
+        if options.update_direct_only and options.update_fallback_only:
+            self.error('Options --update-direct-only and --update-fallback-only '
+                       'are mutually exclusive.')
```
chmanchester: According to how these are used they aren't really mutually exclusive.

firefox_ui_harness/runners/update.py
```diff
@@ -0,0 +1,76 @@
...
+                self.logger.info('Removing copy of the application at "%s"' % target_folder)
+                mozfile.remove(target_folder)
+
+        # Run direct update tests if wanted
+        if self.run_direct_update:
+            _run_tests(update_type='direct')
+
+        # Run fallback update tests if wanted
+        if self.run_fallback_update:
+            _run_tests(update_type='fallback')
```
chmanchester: Whatever's controlling direct/fallback/both doesn't seem resolved to me. The current set up seems to be that all the tests specified by a manifest will be run either once or twice depending on whether one or both of these options are specified, and each time they're run with a fresh binary. Even if this is what's intended the interface should be cleaned up a bit so it's clear.

firefox_ui_harness/testcases/update.py
```diff
@@ -0,0 +1,133 @@
...
+
+        self.updates = []
+
+    def setUp(self, is_fallback=False):
+        FirefoxTestCase.setUp(self)
+
+        self.current_update_index = 0
```
chmanchester: I don't see this incremented anywhere, this seems like a holdover from the mozmill interface.

firefox_ui_harness/testcases/update.py
```diff
@@ -0,0 +1,133 @@
...
+        self.restart()
+
+        # Dictionary which holds the information for each update
+        self.updates = [{
+            'build_pre': self.software_update.build_info,
+            'build_post': self.software_update.build_info,
+            'fallback': is_fallback,
+            'patch': {},
+            'success': False,
+        }]
```
chmanchester: This list never seems to extend beyond length one, maybe it can just be dict.

firefox_ui_harness/testcases/update.py
```diff
@@ -0,0 +1,133 @@
...
+
+        # The upgraded version should be identical with the version given by
+        # the update and we shouldn't have run a downgrade
+        check = self.marionette.execute_script("""
+          Components.utils.import("resource://gre/modules/Services.jsm");
+
+          return  Services.vc.compare(arguments[0], arguments[1]);
+        """, script_args=[info['build_post']['version'], info['build_pre']['version']])
+
+        self.assertGreaterEqual(check, 0, 'The version of the upgraded build is higher or equal')
```
chmanchester: I probably just haven't wrapped my head around this yet, but why is this check not strictly greater? Do some updates not result in version increases?

firefox_ui_harness/testcases/update.py
```diff
@@ -0,0 +1,133 @@
...
+        self.assertEqual(info['build_post']['locale'], info['build_pre']['locale'])
+
+        # Check that no application-wide add-ons have been disabled
+        self.assertEqual(info['build_post']['disabled_addons'],
+                         info['build_pre']['disabled_addons'])
+
+    def restore_config_files(self):
+        # Reset channel-prefs.js file if modified
+        try:
+            if hasattr(self, 'default_update_channel'):
```
chmanchester: I would prefer to default these properties to None rather than use hasattr.

firefox_ui_harness/testcases/update.py
```diff
@@ -0,0 +1,133 @@
...
+
+        # Reset update-settings.ini file if modified
+        try:
+            if hasattr(self, 'default_mar_channels'):
+                path = self.default_mar_channels['path']
+                self.logger.info('Restoring mar channel defaults for: {}'.format(path))
+                with open(path, 'w') as f:
+                    f.write(self.default_mar_channels['content'])
+        except IOError:
+            self.logger.exception('Failed to reset the default mar channels.')
```
chmanchester: Same comment about how this isn't defined on a mozlog logger.

firefox_ui_harness/testcases/update.py
```diff
@@ -0,0 +1,133 @@
...
+    def restore_config_files(self):
+        # Reset channel-prefs.js file if modified
+        try:
+            if hasattr(self, 'default_update_channel'):
+                path = self.default_update_channel['path']
+                self.logger.info('Restoring channel defaults for: {}'.format(path))
+                with open(path, 'w') as f:
+                    f.write(self.default_update_channel['content'])
+        except IOError:
+            self.logger.exception('Failed to reset the default update channel.')
```
chmanchester: If you aren't going to let the exception propagate, call "self.fail()" to ensure the test fails if that's what's intended here.

firefox_ui_tests/update/test_direct_update.py
```diff
@@ -0,0 +1,58 @@
...
+            update_available = about_window.check_for_updates()
+            self.assertTrue(update_available)
+
+            # Download update and wait until it has been applied
+            about_window.download()
+            about_window.wait_for_update_applied()
+
+        finally:
+            self.updates[self.current_update_index]['patch'] = about_window.patch_info
+
```
chmanchester: This method from here up is shared with the other class.

firefox_ui_tests/update/test_fallback_update.py
```diff
@@ -0,0 +1,100 @@
...
+                                 dialog.wizard.error_patching)
+
+                # Start downloading the fallback update
+                dialog.download()
+                dialog.close()
+
+            finally:
+                self.updates[self.current_update_index]['patch'] = dialog.patch_info
+
+        # Restart Firefox to apply the update
```
chmanchester: From here down is shared code with the other class.

firefox_ui_harness/testcase.py
whimboo: > Ok, so can we move this helper inline here and not add the additional
> check to setUp? If a test is switched into the wrong window at setUp it
> means the previous test has leaked a window and we're unlikely to
> succeed by virtue of switching back to a browser window.

What's the problem for at least guarding our harness to continue with
tests if at least one browser window is open, but which is not the
current window. If we have code like you proposed, we would mark all the
following tests as broken.

In tearDown we throw in case of not all additional windows have been
closed. I think that is a wonderful indicator if a test needs a fix. But
shall all remaining ones really suffer from that breakage if we have a
chance to pick up a browser window?

firefox_ui_harness/runners/update.py
whimboo: An exception could definitely happen in copying the application to the target location. Also I checked `run_tests` in Marionette and it could also raise exceptions on various places. Mainly when operating with the file system. There is no general raise/except block for that method. So I feel uncomfortable right now to trust this method.

Keep in mind that if direct updates are failing, we do not want to skip fallback updates but execute them however the direct updates ended-up.

firefox_ui_harness/runners/update.py
whimboo: Huh? Why don't we support `exception`? This is such a helpful method, which handles all the traceback stuff for you.

firefox_ui_harness/runners/update.py
whimboo: `self.type` comes from MarionetteTestCase and specifies the type of tests to run. You can choose them by adding specific variables to the manifest files.

firefox_ui_tests/update/manifest.ini
whimboo: See above. All this is code from the Marionette harness itself.

firefox_ui_harness/testcases/update.py
whimboo: This is not true. UpdateOptions define a [default value](https://github.com/mozilla/firefox-ui-tests/pull/122/files#diff-d58666e1f5c3e602384c1489b217af51R16) which in this case is an empty array.

firefox_ui_harness/options/update.py
whimboo: I don't see what you mean here. If you specify --update-direct-only it doesn't make sense to also allow --update-fallback-only. No test would be run at all. So if none of those options are specified we run both, otherwise only the one which is wanted.

firefox_ui_harness/runners/update.py
whimboo: Maybe you are confused because the CLI options are named `--update-directly-only` instead of `--update-skip-fallback`? If not lets have a Vidyo chat to get that clarified.

firefox_ui_harness/testcases/update.py
whimboo: This exists on purpose. I told multiple times that we want to support updates in multiple steps with Marionette. We haven't finished this off in Mozmill, so this is a straight port-over, which I or someone else will pick up once this PR has been merged. So lets leave this in to not have to rewrite all the logic twice.

firefox_ui_harness/testcases/update.py
whimboo: Please see above.

firefox_ui_harness/testcases/update.py
whimboo: We are testing nightly builds for Nightly and Aurora. So versions will not change under those circumstances. If an explicit check is necessary you will have to specify the target build id  via the CLI option. This is how our tests work since years.

firefox_ui_harness/testcases/update.py
whimboo: I can change this for sure.

firefox_ui_harness/testcases/update.py
whimboo: We don't need those exceptions yet. It's just clean-up code, which should not mark the test as fail if it's not succeeding. 

In case for the allowed mar channels we need to fail the test once we have implemented code for signature checks as requested here: https://github.com/mozilla/mozmill-automation/issues/163. I will port this issue to Bugzilla now.

firefox_ui_tests/update/test_direct_update.py
whimboo: I can combine that for sure. Just wanted to see what could be necessary for multiple update steps. But once we work on those we can split the shared code where necessary.

firefox_ui_harness/runners/base.py
whimboo: Then the `--e10s` option for marionette is misbehaving given that it gives you totally inconsistent states. If that option is specified Firefox should run with multiple processes, otherwise not. I think what is really necessary here is a fix in marionette client: https://dxr.mozilla.org/mozilla-central/source/testing/marionette/client/marionette/runner/base.py#473

If setting the pref to False in the client is not wanted, the option should be renamed to `--force-e10s`. As of now it is totally misleading about the underlying behavior. Exactly that's why my confusion.

I will check how to make an exception for this preference but I stay with the point that our preferences have to overrule preferences as set by Marionette.

whimboo: My next commit will contain a fix for that.

firefox_ui_harness/testcases/update.py
chmanchester: Sure, it's not "every time" you pop here, but you don't set a default for all of them.

firefox_ui_harness/testcases/update.py
chmanchester: Ok can you add a comment to that effect?

firefox_ui_harness/runners/update.py
chmanchester: We talked about this and it should work as expected, I was mistaken about the significance of self.type for the runner.

firefox_ui_harness/testcases/update.py
```diff
@@ -0,0 +1,133 @@
...
+
+        # Bug 1142805 - Until we don't modify the channel-prefs.js and update-settings.ini
+        # files before Firefox gets started, a restart of Firefox is necessary to
+        # accept the new update channel.
+        self.restart()
+
+        # Dictionary which holds the information for each update
+        self.updates = [{
+            'build_pre': self.software_update.build_info,
+            'build_post': self.software_update.build_info,
```
chmanchester: Let's not add this key until after the update is performed. I had to look at this pretty closely to conclude it does the right thing based on build_post getting updated after the update. The assertions are mostly using assertEqual, so if we had mistakenly been aliasing build_pre in build_post they would almost all pass despite the error.

firefox_ui_harness/testcases/update.py
```diff
@@ -0,0 +1,133 @@
...
+        # If a target version has been specified, check if it matches the updated build
+        if self.target_version:
+            self.assertEqual(info['build_post']['version'], self.target_version)
+
+        # The post buildid should be identical with the buildid contained in the patch
+        self.assertEqual(info['build_post']['buildid'], info['patch']['buildid'])
+
+        # If a target buildid has been specified, check if it matches the updated build
+        if self.target_buildid:
+            self.assertEqual(info['build_post']['buildid'], self.target_buildid)
```
chmanchester: Can we assert somewhere in here that the pre and post build ids differ?

firefox_ui_harness/options/update.py
chmanchester: When I run these locally it looks like the direct updates are run only by default. How do I run both?

firefox_ui_tests/update/manifest.ini
chmanchester: I see now this is working basically as expected, but what does this give us beyond just having separate manifests?

firefox_ui_harness/runners/base.py
whimboo: As discussed with Chris yesterday I will keep the code as we have now and we check later what's necessary to do.

firefox_ui_harness/testcases/update.py
whimboo: As discussed yesterday all methods on the logger have an additional parameter which can be used to instruct the logger to print traceback information: http://mozbase.readthedocs.org/en/latest/mozlog_structured.html#mozlog.structured.structuredlog.StructuredLogger

This works pretty fine, so I'm going to change the code appropriately. Thanks!

firefox_ui_harness/testcases/update.py
whimboo: >> +        self.current_update_index = 0
> 
> Ok can you add a comment to that effect?

Interestingly I do not see where this comment has been added. It's
simply not visible. I will add a comment given that this seems to be
useful here.

firefox_ui_harness/testcases/update.py
whimboo: A default for command line options is only necessary if it is != None. Given that only `update_mar_channels` requires an array as default it has been set this way. There is no default necessary for `update_channel`.

If you are still having something else in mind please tell me exactly, because I'm still not getting the problem here. :S

firefox_ui_harness/testcases/update.py
whimboo: We have to do this to workaround the fact that a possible crash or whatever else doesn't let us finish the test. If we move this assignment to the end of the test, we won't have a post build assigned. This would add a lot more extra cases to the dashboard (treeherder view).

Even that we are safer now for Python tests, I would propose that we keep this for now and if necessary find an alternative solution as a follow-up bug. The code as is has been proven to work over the last years.

firefox_ui_harness/testcases/update.py
whimboo: There is no need for that. See line 91 where we already assert that the post build id has to be equal to the build id of the downloaded patch. That is the reference we need.

firefox_ui_harness/options/update.py
whimboo: When you do not specify those optional options both type of updates are run. You really only need this in case you want to run only the one or the other.

firefox_ui_tests/update/manifest.ini
whimboo: We can do this. The benefit would be that the other test for the other type is not marked as skipped. So I will test it, and if it works as expected I will move the tests to individual sub folders and create appropriate manifest files.

firefox_ui_harness/runners/base.py
whimboo: So after more tests I cannot keep the code as is! It's totally broken and I already filed bug 1146847 to get the priority order fixed. If we keep it none of the preferences we need for update tests are kept but thrown away. Frankly I don't have the time to check where in Marionette this happens. So I will place a workaround for the e10s mode in here for now. I hope that will be fine.

firefox_ui_harness/runners/update.py
```diff
@@ -0,0 +1,76 @@
...
+        FirefoxUITestRunner.__init__(self, **kwargs)
+
+        self.original_bin = self.bin
+
+        self.prefs.update(DEFAULT_PREFS)
+
+        # In case of overriding the update URL, set the appropriate preference
+        override_url = kwargs.pop('update_override_url', None)
+        if override_url:
+            self.prefs.update({'app.update.url.override': override_url})
```
whimboo: Some of the recent changes for preference handling broke the override url handling of Firefox. Not sure why, but even with this preference set, the `app.update.url` pref is not getting updated.

whimboo: Hm, it seems to work at least for the update check and when I run with `--gecko-log -` I see:

*** AUS:SVC Checker:checkForUpdates - sending request to: http://localhost?force=1

We will have to get this value into our update status array only.

firefox_ui_harness/runners/base.py
```diff
@@ -2,7 +2,16 @@
...
+import mozinfo
+from marionette import BaseMarionetteTestRunner
+
+import firefox_ui_tests
+from ..testcase import FirefoxTestCase
+
+
+DEFAULT_PREFS = {
     'app.update.auto': False,
     'app.update.enabled': False,
```
whimboo: We have a problem here, given that this pref is not obeyed and always set to `True` for all tests! This would mean a restart test could upgrade the binary under test, which is totally forbidden. All this can be added to our discussion about which defaults preferences have higher priority. Here we definitely fail with the current approach. I will file a bug so we can get this fixed ASAP.

whimboo: At the end I reverted the pref handling code to really use `self.prefs` to set our default preferences. This at least fixes the problem @mjzffr is also seeing for her long run Marionette tests based on our framework. She filed [bug 1146990](https://bugzilla.mozilla.org/show_bug.cgi?id=1146990) for it.

firefox_ui_harness/testcases/update.py
chmanchester: I think I was having problems because I was using the wrong script (firefox-ui-tests vs. firefox-ui-update). Don't worry about it.

firefox_ui_harness/runners/base.py
chmanchester: I'll comment in the bug, but I can't reproduce this issue.

firefox_ui_harness/runners/base.py
chmanchester: You need to provide steps to reproduce this issue.

firefox_ui_harness/runners/base.py
```diff
@@ -41,9 +49,43 @@
...
+
+        if not self.server_root:
+            self.server_root = firefox_ui_tests.resources
+
+        # Bug 1146847: This needs a refactoring given that our default
+        # preferences are not coming from the command line
+        self.prefs.update(DEFAULT_PREFS)
+
+        if not kwargs.get('e10s'):
+            self.prefs.update({'browser.tabs.remote': False})
```
chmanchester: This is the wrong pref name, and it changes the behavior from before slightly. If I revert this to how this was handled before the same set of prefs get associated with our instance.

firefox_ui_harness/runners/update.py
```diff
@@ -0,0 +1,97 @@
...
+            except Exception:
+                self.exc_info = sys.exc_info()
+                self.logger.error('Failure during execution of the update test.',
+                                  exc_info=self.exc_info)
+
+            finally:
+                self.logger.info('Removing copy of the application at "%s"' % target_folder)
+                try:
+                    mozfile.remove(target_folder)
+                except IOError, e:
```
chmanchester: nit: `except IOError as e:`

firefox_ui_harness/testcase.py
```diff
@@ -5,28 +5,41 @@
...
-        self._start_handle_count = len(self.marionette.window_handles)
         self.marionette.set_context('chrome')
 
-        self.browser = self.windows.current
-        self.browser.focus()
-        with self.marionette.using_context(self.marionette.CONTEXT_CONTENT):
-            # Ensure that we have a default page opened
-            self.marionette.navigate(self.prefs.get_pref('browser.newtab.url'))
+        # Ensure that we always have a valid browser instance available
+        self.browser = self.windows.switch_to(lambda win: type(win) is BrowserWindow)
```
chmanchester: I asked in the last review that setUp not be changed. With `self.browser.focus()` in tearDown this doesn't seem necessary.

firefox_ui_harness/testcases/update.py
chmanchester: Then can you have its value be `None` here? I'm skeptical of making decisions here based on a dashboard we haven't made yet.

firefox_ui_harness/runners/base.py
whimboo: Just modify a test on the master branch and add a print statement for `app.update.enabled`. You will see that it is set to `True`. You won't be able to replicate it on my update_test branch.

chmanchester: I just tried again from master. I get:
```
(Pdb) self.prefs.get_pref('app.update.enabled')
False
(Pdb) self.marionette.instance.prefs['app.update.enabled']
False
(Pdb) 
```

firefox_ui_harness/runners/base.py
whimboo: Ups, sorry. That was a bad copy&paste from somewhere else. Thanks for noticing that. I should have indeed used `browser.tabs.remote.autostart` here. Without `--e10s` we don't run multip-process tabs, and with it specified we do. So I hope you are ok with it as pointed out on the bug. Follow-up commit will be up soon.

firefox_ui_harness/runners/base.py
whimboo: Strange. I also see this now. I don't know what was going on earlier today. At least this was the problem why my local update tests were failing yesterday.

firefox_ui_harness/testcase.py
whimboo: Sorry, that was a rebase I had to do and it clearly went wrong. It clashed with one of my latest landed PRs. I will fix that so we have the behavior from master.

firefox_ui_harness/testcases/update.py
whimboo: I will do it. So we will tweak it when I returned.

firefox_ui_harness/runners/base.py
```diff
@@ -41,9 +49,43 @@
...
+
+        if not self.server_root:
+            self.server_root = firefox_ui_tests.resources
+
+        # Bug 1146847: This needs a refactoring given that our default
+        # preferences are not coming from the command line
+        self.prefs.update(DEFAULT_PREFS)
+
+        if not kwargs.get('e10s'):
+            self.prefs.update({'browser.tabs.remote.autostart': False})
```
chmanchester: This still has an undesirable effect if we're running against a nightly that doesn't have e10s on by default, because I'll be able to pass --e10s but not get e10s enabled and otherwise this change has no effect on the set of prefs enabled for tests. I don't think this comment should reference that bug because I think that bug's going to need to be resolved invalid.

firefox_ui_harness/testcases/update.py
```diff
@@ -0,0 +1,234 @@
...
+
+        # Reset update-settings.ini file if modified
+        try:
+            if self.default_mar_channels:
+                path = self.default_mar_channels['path']
+                self.logger.info('Restoring mar channel defaults for: {}'.format(path))
+                with open(path, 'w') as f:
+                    f.write(self.default_mar_channels['content'])
+        except IOError:
+            self.logger.exception('Failed to reset the default mar channels.',
```
chmanchester: This needs to be `logger.error` like the others.

whimboo: Damn. This somehow got back. I clearly fixed that earlier. Thanks for spotting it.

PR #115, opened 2015-03-05
Bug 1139594: Add close method to toolbar.py

Nebelhom:
This PR addresses [bug 1132624](https://bugzilla.mozilla.org/show_bug.cgi?id=1132624) and is based on PR #105 

Thanks @Ethcelon :)

PR #105, opened 2015-02-26
Added close method to IdentityPopup r=whimboo

Ethcelon:
This is required for tests using identity popup
#99

PR #124, opened 2015-03-11
Bug 1139594 - Add close method for identity_popup to toolbars.py

galgeek:
https://bugzilla.mozilla.org/show_bug.cgi?id=1139594

This PR is based on #105 and #115. Thanks, @Ethcelon and @Nebelhom!

The code from those earlier PRs fails with a NoSuchElementException when I call identity_popup.close() from test_green_larry. It looks like opening page info removes the identity popup from the DOM.

### DIFF
--- a/firefox_puppeteer/ui/toolbars.py
+++ b/firefox_puppeteer/ui/toolbars.py
@@ -2,7 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
-from marionette_driver import Wait, By
+from marionette_driver import By, expected, Wait
 
 from ..api.keys import Keys
 from ..api.l10n import L10n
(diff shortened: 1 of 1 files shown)


firefox_puppeteer/ui/toolbars.py
```diff
@@ -446,3 +446,21 @@ def verifier(self):
...
         """
         return self.marionette.find_element(By.ID, 'identity-popup-content-verifier')
+
+    def close(self, force=False):
+        """Closes the identity popup by hitting the escape key.
+
+        :param force: Optional, if True, the popup will be force closed,
+         default to `False`.
+        """
+        if expected.element_not_present(By.ID, 'identity-popup') or not self.is_open:
```
whimboo: The panel does always exist. It's not getting removed from the DOM. So the expected check should really not be necessary. Do you have some example code I can execute to see that?

whimboo:
I have taken the opportunity with Johannes to continue on his PR. So the issue will be fixed with PR #115.

PR #130, opened 2015-03-18
Bug 1132688 - Convert testMixedScriptContentBlocking to a firefox-ui-tes...

chmanchester:
...t. r=whimboo

### DIFF
--- a/firefox_ui_tests/remote/security/test_mixed_script_content_blocking.py
+++ b/firefox_ui_tests/remote/security/test_mixed_script_content_blocking.py
@@ -1,0 +1,116 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from marionette_driver import By, Wait
+
+from firefox_ui_harness.decorators import skip_under_xvfb
+from firefox_ui_harness.testcase import FirefoxTestCase
+
+
+class TestMixedScriptContentBlocking(FirefoxTestCase):
+
+    def setUp(self):
+        FirefoxTestCase.setUp(self)
+        self.url = 'https://mozqa.com/data/firefox/security/mixed_content_blocked/index.html'
+        self.test_elements = [
+            ('result1', 'Insecure script one'),
+            ('result2', 'Insecure script from iFrame'),
+            ('result3', 'Insecure plugin'),
+            ('result4', 'Insecure stylesheet'),
+        ]
+        self.locationbar = self.browser.navbar.locationbar
+
+    def tearDown(self):
+        try:
+            self.marionette.execute_script("arguments[0].hidePopup();",
+                                           script_args=[self.locationbar.notification_popup])
+        finally:
+            FirefoxTestCase.tearDown(self)
+
+    def _expect_protection_status(self, enabled):
+        if enabled:
+            color, icon_filename, state = (
+                "rgb(0, 136, 0)",
+                "identity-icons-https.png",
+                "blocked"
+            )
+        else:
+            color, icon_filename, state = (
+                "rgb(255, 0, 0)",
+                "identity-icons-https-mixed-active.png",
+                "unblocked"
+            )
+
+        icon_id = "bad-content-%s-notification-icon" % state
+        icon = self.marionette.find_element(By.ID, icon_id)
+        Wait(self.marionette).until(
+            lambda _: icon.get_attribute('showing') == 'true',
+            message="The icon should be showing"
+        )
+
+        favicon = self.locationbar.favicon
+
+        Wait(self.marionette).until(
+            lambda _: icon_filename in favicon.value_of_css_property('list-style-image'),
+            message="The correct icon is displayed"
+        )
+
+        with self.marionette.using_context('content'):
+            for identifier, description in self.test_elements:
+                el = self.marionette.find_element(By.ID, identifier)
+                Wait(self.marionette).until(
+                    lambda mn: el.value_of_css_property('color') == color,
+                    message=("%s has been %s" % (description, state))
+                )
+
+    def expect_protection_enabled(self):
+        self._expect_protection_status(True)
+
+    def expect_protection_disabled(self):
+        self._expect_protection_status(False)
+
+    @skip_under_xvfb
+    def test_mixed_content_page(self):
+        with self.marionette.using_context('content'):
+            self.marionette.navigate(self.url)
+
+        self.expect_protection_enabled()
+
+        # TODO: This should consume the ui class for popups when implemented in
+        # bug 1144873.
+        popup = self.locationbar.notification_popup
+        self.assertEqual(popup.get_attribute('state'), 'closed')
+
+        popupbox = self.marionette.find_element(By.ID,
+                                                'bad-content-blocked-notification-icon')
+        popupbox.click()
+
+        Wait(self.marionette).until(
+            lambda _: popup.get_attribute('state') == 'open',
+            message="The notification popup should be open"
+        )
+
+        notification_element = self.marionette.find_element(By.ID, 'bad-content-notification')
+
+        button_label = self.browser.get_entity('mixedContentBlocked2.options')
+        options_button = notification_element.find_element(By.ANON_ATTRIBUTE,
+                                                           {'label': button_label})
+        options_button.click()
+
+        item_label = self.browser.get_entity('mixedContentBlocked2.unblock.label')
+        menu_item = notification_element.find_element(By.ANON_ATTRIBUTE,
+                                                      {'label': item_label})
+        menu_item.click()
+
+        Wait(self.marionette).until(
+            lambda _: popup.get_attribute('state') == 'closed',
+            message="The notification popup should be closed"
+        )
+
+        self.expect_protection_disabled()
+
+        with self.marionette.using_context('content'):
+            self.marionette.navigate(self.url)
+
+        self.expect_protection_disabled()
--- a/firefox_ui_tests/remote/security/manifest.ini
+++ b/firefox_ui_tests/remote/security/manifest.ini
@@ -1,4 +1,5 @@
 [test_ev_certificate.py]
+[test_mixed_script_content_blocking.py]
 [test_safe_browsing_notification.py]
 [test_safe_browsing_warning_pages.py]
 [test_security_notification.py]


firefox_ui_tests/remote/security/test_mixed_script_content_blocking.py
```diff
@@ -0,0 +1,100 @@
...
+
+from firefox_ui_harness.decorators import skip_under_xvfb
+from firefox_ui_harness.testcase import FirefoxTestCase
+
+
+class TestMixedScriptContentBlocking(FirefoxTestCase):
+
+    def setUp(self):
+        FirefoxTestCase.setUp(self)
+        self.test_url = "https://mozqa.com/data/firefox/security/mixed_content_blocked/index.html"
```
whimboo: In all of our tests we use `self.url` or `self.urls`. To be consistent please adjust it here. Also please use single quotes.

firefox_ui_tests/remote/security/test_mixed_script_content_blocking.py
```diff
@@ -0,0 +1,100 @@
...
+        ]
+        self.locationbar = self.browser.navbar.locationbar
+
+    def tearDown(self):
+        self.marionette.set_context("chrome")
+        self.marionette.execute_script("arguments[0].hidePopup();",
+                                       script_args=[self.locationbar.notification_popup])
+        FirefoxTestCase.tearDown(self)
+
+    def _expect_protection_status(self, enabled):
```
whimboo: Personally I would propose to move helper methods to the end of the class, so the test itself comes first. That's what you usually want to have a look at.

firefox_ui_tests/remote/security/test_mixed_script_content_blocking.py
```diff
@@ -0,0 +1,100 @@
...
+                Wait(self.marionette).until(
+                    lambda mn: el.value_of_css_property('color') == color,
+                    message=("%s has been %s" % (description, state))
+                )
+
+    def expect_protection_enabled(self):
+        self._expect_protection_status(True)
+
+    def expect_protection_disabled(self):
+        self._expect_protection_status(False)
```
whimboo: I don't see a need for those helper methods. It's just that you set a default value. I would make the parameter for _expect_protection_status a kwarg with the name `enabled`. So it obvious in the test to call `self.expect_protection_status(enabled=True)`.

firefox_ui_tests/remote/security/test_mixed_script_content_blocking.py
```diff
@@ -0,0 +1,100 @@
...
+
+    @skip_under_xvfb
+    def test_mixed_content_page(self):
+        with self.marionette.using_context('content'):
+            self.marionette.navigate(self.test_url)
+
+        self.expect_protection_enabled()
+
+        panel = self.locationbar.notification_popup
+        self.assertEqual(panel.get_attribute('state'), 'closed')
```
whimboo: Please add a TODO line for [bug 1139544](https://bugzilla.mozilla.org/show_bug.cgi?id=1139544).

firefox_ui_tests/remote/security/test_mixed_script_content_blocking.py
```diff
@@ -0,0 +1,100 @@
...
+
+        popupbox = self.marionette.find_element(By.ID,
+                                                'bad-content-blocked-notification-icon')
```
whimboo: This element is inside the notification bar, so please use `popup.find_element()` here.

firefox_ui_tests/remote/security/test_mixed_script_content_blocking.py
```diff
@@ -0,0 +1,100 @@
...
+        self._expect_protection_status(False)
+
+    @skip_under_xvfb
+    def test_mixed_content_page(self):
+        with self.marionette.using_context('content'):
+            self.marionette.navigate(self.test_url)
+
+        self.expect_protection_enabled()
+
+        panel = self.locationbar.notification_popup
```
whimboo: s/panel/popup.

firefox_ui_tests/remote/security/test_mixed_script_content_blocking.py
```diff
@@ -0,0 +1,100 @@
...
+        self.assertEqual(panel.get_attribute('state'), 'closed')
+
+        popupbox = self.marionette.find_element(By.ID,
+                                                'bad-content-blocked-notification-icon')
+        popupbox.click()
+
+        self.wait_for_condition(lambda _: panel.get_attribute('state') == 'open')
+
+        parent_element = self.marionette.find_element(By.ID, 'bad-content-notification')
```
whimboo: Lets call this variable `notification`.

firefox_ui_tests/remote/security/test_mixed_script_content_blocking.py
```diff
@@ -0,0 +1,100 @@
...
+        item_label = self.browser.get_entity('mixedContentBlocked2.unblock.label')
+        menu_item = parent_element.find_element(By.ANON_ATTRIBUTE,
+                                                {'label': item_label})
+        menu_item.click()
+
+        self.wait_for_condition(lambda _: panel.get_attribute('state') == 'closed')
+
+        self.expect_protection_disabled()
+
+        self.locationbar.reload_url()
```
whimboo: This method does not wait for the page being loaded. I would expect race conditions. Given that this test doesn't test the reload feature of the locationbar we can use `navigate()`.

firefox_ui_tests/remote/security/test_mixed_script_content_blocking.py
```diff
@@ -0,0 +1,100 @@
...
+        self.test_elements = [
+            ('result1', 'Insecure script one'),
+            ('result2', 'Insecure script from iFrame'),
+            ('result3', 'Insecure plugin'),
+            ('result4', 'Insecure stylesheet'),
+        ]
+        self.locationbar = self.browser.navbar.locationbar
+
+    def tearDown(self):
+        self.marionette.set_context("chrome")
```
whimboo: We don't leave chrome scope in this test, so this line is not necessary. Also please do clean-up code in a try/finally so we can ensure to call the super teardown method.

firefox_ui_tests/remote/security/test_mixed_script_content_blocking.py
```diff
@@ -0,0 +1,100 @@
...
+                                       script_args=[self.locationbar.notification_popup])
+        FirefoxTestCase.tearDown(self)
+
+    def _expect_protection_status(self, enabled):
+        color, parts, state = (
+            "rgb(255, 0, 0)",
+            "identity-icons-https-mixed-active.png",
+            "unblocked"
+        )
+        if enabled:
```
whimboo: Lets make this an if/else.

whimboo: Or define all in a hash and have sub items for `enabled` and `disabled`.

firefox_ui_tests/remote/security/test_mixed_script_content_blocking.py
```diff
@@ -0,0 +1,100 @@
...
+            color, parts, state = (
+                "rgb(0, 136, 0)",
+                "identity-icons-https.png",
+                "blocked"
+            )
+
+        icon_id = "bad-content-%s-notification-icon" % state
+        icon = self.marionette.find_element(By.ID, icon_id)
```
whimboo: This is below the locationbar, so it should be the root element here. Also I really don't like that many find_element calls in our test. We really have to get those notification bar code implemented. :(

firefox_ui_tests/remote/security/test_mixed_script_content_blocking.py
```diff
@@ -0,0 +1,100 @@
...
+        self.locationbar = self.browser.navbar.locationbar
+
+    def tearDown(self):
+        self.marionette.set_context("chrome")
+        self.marionette.execute_script("arguments[0].hidePopup();",
+                                       script_args=[self.locationbar.notification_popup])
+        FirefoxTestCase.tearDown(self)
+
+    def _expect_protection_status(self, enabled):
+        color, parts, state = (
```
whimboo: Instead of `parts` I would suggest to use `icon_filename`.

firefox_ui_tests/remote/security/test_mixed_script_content_blocking.py
```diff
@@ -0,0 +1,100 @@
...
+        if enabled:
+            color, parts, state = (
+                "rgb(0, 136, 0)",
+                "identity-icons-https.png",
+                "blocked"
+            )
+
+        icon_id = "bad-content-%s-notification-icon" % state
+        icon = self.marionette.find_element(By.ID, icon_id)
+        self.wait_for_condition(lambda _: icon.get_attribute('showing') == 'true')
```
whimboo: I would suggest that we really start to use `Wait` for all the new tests. That way we can also pass-in a message, which would in most cases be very helpful to understand the failure.

firefox_ui_tests/remote/security/test_mixed_script_content_blocking.py
chmanchester: This fixes the failure in teardown that occurs because marionette leaves us in content scope when a test is skipped. Hiding this popup should never throw, if it does we want the test to fail, which wouldn't happen if we use try/finally.

firefox_ui_tests/remote/security/test_mixed_script_content_blocking.py
chmanchester: I'm not going to make this change.

firefox_ui_tests/remote/security/test_mixed_script_content_blocking.py
chmanchester: The uniqueness of this selector is determined by its id, I don't see any benefit to using a different root. The notification bar code is a thin abstraction that basically requires you to pass in a full id to get elements anyway, adding it later will be trivial.

firefox_ui_tests/remote/security/test_mixed_script_content_blocking.py
chmanchester: I prefer using the helper functions for this so we don't have to expose boolean parameters that might have unclear meaning to test code. Without the kwarg name it's not clear to me what "expect_protection_status(True)" would mean.

firefox_ui_tests/remote/security/test_mixed_script_content_blocking.py
chmanchester: This is a notification popup not a notification bar, so it's not clear to me it's covered by that bug.

firefox_ui_tests/remote/security/test_mixed_script_content_blocking.py
chmanchester: As above, the selector is already unambiguous so I don't see any value in this change.

firefox_ui_tests/remote/security/test_mixed_script_content_blocking.py
chmanchester: When a test is skipped we get left in content scope so getting the element fails. We've talked about this multiple times.

firefox_ui_tests/remote/security/test_mixed_script_content_blocking.py
whimboo: Why not? I would like to know what your opposite statement would be here.

firefox_ui_tests/remote/security/test_mixed_script_content_blocking.py
whimboo: Then please use kwargs. Adding more levels to the callstack is only confusing and I really don't see a value in those helpers here.

firefox_ui_tests/remote/security/test_mixed_script_content_blocking.py
whimboo: Oh, right. So that means that the identity popup is also of the same type, and we would have to convert the class for doorhanger notifications. Mind filing the bug and adding the TODO? We shouldn't forget about it.

firefox_ui_tests/remote/security/test_mixed_script_content_blocking.py
whimboo: We keep a hierarchy of elements and using find_element from the root, is killing the current level in the DOM. When we convert this to a new class, it would have to be done anyway. So please change that.

firefox_ui_tests/remote/security/test_mixed_script_content_blocking.py
whimboo: This does not correlate to other tests which are also skipped and we have clean-up code operating in chrome scope like https://github.com/mozilla/firefox-ui-tests/blob/master/firefox_ui_tests/remote/security/test_safe_browsing_notification.py#L44. There we do NOT switch to chrome scope again and it works. So the above code still looks very suspicious. Please show me the code in Marionette which sets us back into content when a test is skipped. I strongly feel this is a bug in Marionette then.

firefox_ui_tests/remote/security/test_mixed_script_content_blocking.py
whimboo: Please change it so it will be clear that this element is located beneath the locationbar. This will help at least contributors who would like to work on that. Also finding this element will be faster from a node which is closer in the tree as when you start with the root node.

firefox_ui_tests/remote/security/test_mixed_script_content_blocking.py
chmanchester: It's a subjective matter of style and I've already written the class in the way I find most readable. In the absence of a style guide that explicitly contradicts me on this point, and given it has no possible effect on the coverage provided by the test, I don't see why I would make this change.

firefox_ui_tests/remote/security/test_mixed_script_content_blocking.py
chmanchester: I would prefer not to make this change and I can't imagine why this would be a blocker for landing this test.

firefox_ui_tests/remote/security/test_mixed_script_content_blocking.py
chmanchester: https://bugzilla.mozilla.org/show_bug.cgi?id=1144873

firefox_ui_tests/remote/security/test_mixed_script_content_blocking.py
chmanchester: If this was in a UI library class I would use this style of selector (like I did in the toolbar classes), I don't see why this change is desirable.

firefox_ui_tests/remote/security/test_mixed_script_content_blocking.py
chmanchester: The code you've linked works by catching the exception. This code works by addressing the cause of the exception so it's never thrown. We noticed this here: https://bugzilla.mozilla.org/show_bug.cgi?id=1127831#c3 and I already noted it was a marionette bug, but again, easy to work around.

It looks like gather_debug in BaseMarionetteTestRunner does the context switch, and this isn't called when the test runs an succeeds.

firefox_ui_tests/remote/security/test_mixed_script_content_blocking.py
chmanchester: Finding the element with the inspector is easy enough to do with the id, and the actual hierarchy can be further examined from there. Finding the parent and then calling find_element would require two round trips to the server, I think that would end up being slower. I don't think this change is needed.

firefox_ui_tests/remote/security/test_mixed_script_content_blocking.py
chmanchester: https://bugzilla.mozilla.org/show_bug.cgi?id=1144907

firefox_ui_tests/remote/security/test_mixed_script_content_blocking.py
whimboo: I would agree if a method is wrapped which needs lots of specific parameters set, but here it's just a single parameter, which can even be set as a kwarg.

I'm still not a fan, but don't want to get a large debate started on it. So leave it as is.

firefox_ui_tests/remote/security/test_mixed_script_content_blocking.py
```diff
@@ -0,0 +1,116 @@
...
+    def tearDown(self):
+        try:
+            self.marionette.set_context("chrome")
+            self.marionette.execute_script("arguments[0].hidePopup();",
+                                           script_args=[self.locationbar.notification_popup])
+        finally:
+            FirefoxTestCase.tearDown(self)
+
+    def _expect_protection_status(self, enabled):
+
```
whimboo: nit: lets be consistent and lets get those empty lines removed from the beginning of a method.

firefox_ui_tests/remote/security/test_mixed_script_content_blocking.py
whimboo: Given that this will be fixed shortly, we can leave this out and land the test once the patch from that bug in on mozilla-central.

firefox_ui_tests/remote/security/test_mixed_script_content_blocking.py
```diff
@@ -0,0 +1,116 @@
...
+    @skip_under_xvfb
+    def test_mixed_content_page(self):
+        with self.marionette.using_context('content'):
+            self.marionette.navigate(self.url)
+
+        self.expect_protection_enabled()
+
+        popup = self.locationbar.notification_popup
+        self.assertEqual(popup.get_attribute('state'), 'closed')
```
whimboo: Please add at least a single TODO comment for bug 1144873 in this test, so we know that code has to be updated once the doorhanger notifications are getting implemented.

PR #131, opened 2015-03-18
Dv cert

chmanchester:


firefox_ui_tests/remote/security/manifest.ini
```diff
@@ -1,3 +1,5 @@
+[test_dv_certificate.py]
+[test_mixed_script_content_blocking.py]
```
whimboo: I think you didn't branch off from master but from your former integration branch.

firefox_ui_tests/remote/security/test_dv_certificate.py
```diff
@@ -0,0 +1,103 @@
...
+from marionette_driver import By, Wait
+
+from firefox_ui_harness.decorators import skip_under_xvfb
+from firefox_ui_harness.testcase import FirefoxTestCase
+
+class TestDVCertificate(FirefoxTestCase):
+
+    def setUp(self):
+        FirefoxTestCase.setUp(self)
+        self.test_url = "https://ssl-dv.mozqa.com"
```
whimboo: `self.url` similar to all the other tests please.

firefox_ui_tests/remote/security/test_dv_certificate.py
```diff
@@ -0,0 +1,103 @@
...
+        self.cert = None
```
whimboo: No need to define it at this level. We only had to do it in Mozmill Tests because of the callbacks. Lets make it a normal variable in the test.

firefox_ui_tests/remote/security/test_dv_certificate.py
```diff
@@ -0,0 +1,103 @@
...
+        self.locationbar = self.browser.navbar.locationbar
+        self.identity_popup = self.browser.navbar.locationbar.identity_popup
+
+    def tearDown(self):
+        self.marionette.set_context("chrome")
+        self.identity_popup.close(force=True)
+        FirefoxTestCase.tearDown(self)
```
whimboo: I miss try/finally, and there is no need for set_context given that we stay in chrome scope for this test.

firefox_ui_tests/remote/security/test_dv_certificate.py
```diff
@@ -0,0 +1,103 @@
...
+
+    @skip_under_xvfb
+    def test_larry_blue(self):
```
whimboo: Lets rip off `larry` from all our tests. It's misleading and hard to understand if you don't have the history. So just take DV here.

firefox_ui_tests/remote/security/test_dv_certificate.py
```diff
@@ -0,0 +1,103 @@
...
+
+        favicon_hidden = self.marionette.execute_script("""
+          return arguments[0].hasAttribute("hidden");
+        """, script_args=[self.locationbar.favicon])
+
+        self.assertFalse(favicon_hidden)
+
+        self.assertIn('identity-icons-https',
+                      self.locationbar.favicon.value_of_css_property('list-style-image'),
+                      'There is a lock icon')
```
whimboo: Lets not add the message here because it will rip off the actual values. :/ We might have to override those methods in our testclass, to allow a more detailed output.

firefox_ui_tests/remote/security/test_dv_certificate.py
```diff
@@ -0,0 +1,103 @@
...
+
+        self.assertEqual(self.identity_popup.box.get_attribute('className'),
+                         'verifiedDomain',
+                         'Identity is verified')
+
+        panel = self.identity_popup.popup
```
whimboo: This is not used so you can remove it.

firefox_ui_tests/remote/security/test_dv_certificate.py
```diff
@@ -0,0 +1,103 @@
...
+        self.identity_popup.box.click()
+        self.wait_for_condition(lambda _: self.identity_popup.is_open)
```
whimboo: Please use `Wait` with a message.

firefox_ui_tests/remote/security/test_dv_certificate.py
```diff
@@ -0,0 +1,103 @@
...
+
+        doorhanger_class = self.identity_popup.popup.get_attribute('className')
```
whimboo: Put this directly into assertEqual. There is no value to define an extra variable here.

firefox_ui_tests/remote/security/test_dv_certificate.py
whimboo: Or maybe leave the message so it's not lost and we can fix the output later.

firefox_ui_tests/remote/security/test_dv_certificate.py
```diff
@@ -0,0 +1,103 @@
...
+        self.assertEqual(doorhanger_class,
+                         'verifiedDomain',
+                         'The Larry UI is domain verified (aka Blue)')
+
+        encryption_icon = self.identity_popup.encryption_icon
+        lock_image = encryption_icon.value_of_css_property('list-style-image')
```
whimboo: No need for this variable. Use it directly, or does it exceed the line length of 100?

firefox_ui_tests/remote/security/test_dv_certificate.py
```diff
@@ -0,0 +1,103 @@
...
+
+        # Bug 443116
+        # Larry strips the 'www.' from the CName using the eTLDService
+        # This is expected behaviour for the time being
+        host_name = self.identity_popup.host.get_attribute('textContent')
+        expected_name = self.marionette.execute_script("""
+           let Cu = Components.utils;
+           Cu.import("resource://gre/modules/Services.jsm");
+           return Services.eTLD.getBaseDomainFromHost(arguments[0]);
+        """, script_args=[self.cert['commonName']])
```
whimboo: There is a method in the security module for that, which you can use here.

firefox_ui_tests/remote/security/test_dv_certificate.py
```diff
@@ -0,0 +1,103 @@
...
+                         'The Larry UI is domain verified (aka Blue)')
+
+        encryption_icon = self.identity_popup.encryption_icon
+        lock_image = encryption_icon.value_of_css_property('list-style-image')
+        self.assertNotEqual(lock_image, 'none', 'There is a lock icon')
+
+        # Bug 443116
+        # Larry strips the 'www.' from the CName using the eTLDService
+        # This is expected behaviour for the time being
+        host_name = self.identity_popup.host.get_attribute('textContent')
```
whimboo: No reason for this variable. Use the code in the assertEqual call directly.

firefox_ui_tests/remote/security/test_dv_certificate.py
```diff
@@ -0,0 +1,103 @@
...
+           Cu.import("resource://gre/modules/Services.jsm");
+           return Services.eTLD.getBaseDomainFromHost(arguments[0]);
+        """, script_args=[self.cert['commonName']])
+
+        self.assertEqual(host_name, expected_name,
+                         'The site identifier string is equal to the cert host')
+
+        verifier_label = self.browser.get_property('identity.identified.verifier')
+        verifier_label = verifier_label.replace("%S", self.cert['issuerOrganization'])
+        verifier_element = self.identity_popup.verifier
```
whimboo: No need for this extra variable

firefox_ui_tests/remote/security/test_dv_certificate.py
```diff
@@ -0,0 +1,103 @@
...
+
+        self.assertEqual(verifier_element.get_attribute('textContent'),
+                         verifier_label,
+                         'The "Verified by: %S" string is set')
```
whimboo: This might cause problems because %s is a placeholder in Python.

firefox_ui_tests/remote/security/test_dv_certificate.py
```diff
@@ -0,0 +1,103 @@
...
+            self.identity_popup.more_info_button.click()
+
+        page_info_window = self.browser.open_page_info_window(opener)
+        deck = page_info_window.deck
+
+        self.assertEqual(deck.selected_panel, deck.security,
+                         "The security tab is selected by default")
+
+        web_id_domain = deck.security.domain.get_attribute('value')
+        self.assertEqual(web_id_domain, self.cert['commonName'],
```
whimboo: Same as above for the extra variable. It also applies to the checks below. Just use one parameter per line, which is way better readable.

whimboo: Also why did you strip out the replacements in this string? Is that not necessary?

firefox_ui_tests/remote/security/test_dv_certificate.py
```diff
@@ -0,0 +1,103 @@
...
+        security_owner = page_info_window.get_property('securityNoOwner')
+        self.assertEqual(web_id_owner, security_owner,
+                         "Expected owner label found")
+
+        verifier_label = deck.security.verifier.get_attribute('value')
+        self.assertEqual(verifier_label,
+                         self.cert['issuerOrganization'])
+
+        page_info_window.close()
+        self.browser.switch_to()
```
whimboo: You will have to use a try/finally here starting right after the page info window has been opened. In case of test failures above, we won't hit this code.

firefox_ui_tests/remote/security/test_dv_certificate.py
whimboo: This method is missing the `close_all_windows()` call.

firefox_ui_tests/remote/security/manifest.ini
chmanchester: That was intentional so this file wouldn't have conflicts.

firefox_ui_tests/remote/security/test_dv_certificate.py
chmanchester: This sounds like a bug in the mozlog unittest adapter that should be filed.

firefox_ui_tests/remote/security/test_dv_certificate.py
chmanchester: It hasn't landed.

firefox_ui_tests/remote/security/test_dv_certificate.py
chmanchester: I don't think it will.

firefox_ui_tests/remote/security/test_dv_certificate.py
chmanchester: Moved to tearDown.

firefox_ui_tests/remote/security/manifest.ini
whimboo: You will have conflicts anyway when you rebase that branch to master, once the other PR has been merged. Or maybe that one would be earlier. So if that would be the case this PR is blocked by the other PR.

firefox_ui_tests/remote/security/test_dv_certificate.py
whimboo: This is actually a bug in `unittest` with all of those methods. I hope overriding those would be possible.

firefox_ui_tests/remote/security/test_dv_certificate.py
whimboo: So I assume that you have tested it?

chmanchester: This isn't going to cause problems, %S isn't meaningful to string formatting code, and this string isn't being formatted.

firefox_ui_tests/remote/security/test_dv_certificate.py
chmanchester: It doesn't seem necessary for this cert.

firefox_ui_tests/remote/security/test_dv_certificate.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def setUp(self):
+        FirefoxTestCase.setUp(self)
+        self.url = "https://ssl-dv.mozqa.com"
+        self.locationbar = self.browser.navbar.locationbar
+        self.identity_popup = self.browser.navbar.locationbar.identity_popup
+        self.page_info_window = None
+
+    def tearDown(self):
+        self.marionette.set_context("chrome")
```
whimboo: This will be fixed by tomorrow in Marionette. So I don't think we should add it anymore. Thanks btw. for that fix!

firefox_ui_tests/remote/security/test_dv_certificate.py
```diff
@@ -0,0 +1,106 @@
...
+        try:
+            if self.page_info_window is not None:
+                self.page_info_window.close()
```
whimboo: Please force close the window here.

whimboo: Well, we could even remove those two lines given that you call `self.windows.close_all()` right below.

firefox_ui_tests/remote/security/test_dv_certificate.py
```diff
@@ -0,0 +1,106 @@
...
+                            'There is a lock icon')
+
+        # Bug 443116
+        # Larry strips the 'www.' from the CName using the eTLDService
+        # This is expected behaviour for the time being
+        expected_name = self.marionette.execute_script("""
+           let Cu = Components.utils;
+           Cu.import("resource://gre/modules/Services.jsm");
+           return Services.eTLD.getBaseDomainFromHost(arguments[0]);
+        """, script_args=[cert['commonName']])
```
whimboo: This method is available now in the security module. So please update this code.

firefox_ui_tests/remote/security/test_mixed_script_content_blocking.py
```diff
@@ -0,0 +1,100 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
whimboo: For a final review please do not include this file. Thanks. Lets see if we can get the other PR landed soon.

firefox_ui_tests/remote/security/test_dv_certificate.py
```diff
@@ -0,0 +1,98 @@
...
+    def setUp(self):
+        FirefoxTestCase.setUp(self)
+        self.url = "https://ssl-dv.mozqa.com"
+        self.locationbar = self.browser.navbar.locationbar
+        self.identity_popup = self.browser.navbar.locationbar.identity_popup
+
+    def tearDown(self):
+        try:
+            self.browser.switch_to()
+            self.identity_popup.close(force=True)
```
whimboo: I think you also need to catch the NoElementFoundException similar to other security tests which are skipped on Linux. Travis is failing probably here.

firefox_ui_tests/remote/security/test_dv_certificate.py
```diff
@@ -0,0 +1,98 @@
...
+        with self.marionette.using_context('content'):
+            self.marionette.navigate(self.url)
+
+        cert = self.browser.tabbar.selected_tab.certificate
+
+        favicon_hidden = self.marionette.execute_script("""
+          return arguments[0].hasAttribute("hidden");
+        """, script_args=[self.locationbar.favicon])
+
+        self.assertFalse(favicon_hidden)
```
whimboo: Lets add a message here. Otherwise we won't be able to figure out what's broken without having a clean look at the test code.

firefox_ui_tests/remote/security/test_dv_certificate.py
```diff
@@ -0,0 +1,98 @@
...
+                      'There is a lock icon')
+
+        self.assertEqual(self.identity_popup.box.get_attribute('className'),
+                         'verifiedDomain',
+                         'Identity is verified')
+
+        self.identity_popup.box.click()
+        Wait(self.marionette).until(
+            lambda _: self.identity_popup.is_open,
+            message="The popup should be open"
```
whimboo: nit: single quotes please

firefox_ui_tests/remote/security/test_dv_certificate.py
```diff
@@ -0,0 +1,98 @@
...
+
+        encryption_icon = self.identity_popup.encryption_icon
+        self.assertNotEqual(encryption_icon.value_of_css_property('list-style-image'),
+                            'none',
+                            'There is a lock icon')
+
+        # Bug 443116
+        # Larry strips the 'www.' from the CName using the eTLDService
+        # This is expected behaviour for the time being
+        expected_name = self.security.get_domain_from_common_name(cert['commonName'])
```
whimboo: nit: used only once. It can directly be added to the assertEqual call.

firefox_ui_tests/remote/security/test_dv_certificate.py
```diff
@@ -0,0 +1,98 @@
...
+
+        self.assertEqual(self.identity_popup.host.get_attribute('textContent'),
+                         expected_name,
+                         'The site identifier string is equal to the cert host')
+
+        verifier_label = self.browser.get_property('identity.identified.verifier')
+        verifier_label = verifier_label.replace("%S", cert['issuerOrganization'])
```
whimboo: nit: Same here. Would perfectly match inside assertEqual.

firefox_ui_tests/remote/security/test_dv_certificate.py
```diff
@@ -0,0 +1,98 @@
...
+                         'The "Verified by: %S" string is set')
+
+        def opener(mn):
+            self.identity_popup.more_info_button.click()
+
+        page_info_window = self.browser.open_page_info_window(opener)
+        deck = page_info_window.deck
+
+        self.assertEqual(deck.selected_panel, deck.security,
+                         "The security tab is selected by default")
```
whimboo: This will break with e10s enabled. So you also have to skip this test for e10s.

firefox_ui_tests/remote/security/test_dv_certificate.py
chmanchester: This is only failing in with the known cert issue now.

whimboo: Maybe help me out. Which known cert issue do you mean here?

chmanchester: https://bugzilla.mozilla.org/show_bug.cgi?id=1146385

whimboo: Oh, you were talking about the full set of tests. Given that you commented here, I was wondering what's wrong with out teardown method. I have retriggered the Travis build, so lets see if it is green.

whimboo: All is passing now. If all review comments have been addressed its ready to get landed.

PR #132, opened 2015-03-18
Bug 1132698 - Convert Mozmill testSecurityInfoViaMoreInformation

galgeek:
https://bugzilla.mozilla.org/show_bug.cgi?id=1132698


whimboo:
All of those tests are indeed covered by the upcoming test for EV certificates via pull #120. So this test is obsolete.

PR #89, opened 2015-02-17
Bug 1132708 - Convert Mozmill test 'remote/testSecurity/testUntrustedConectionErrorPage.js' to Marionette

Ethcelon:

PR #133, opened 2015-03-18
Bug 1132708 - Convert Mozmill test 'remote/testSecurity/testUntrustedCon...

jmaher:
...ectionErrorPage.js' to Marionette.  This replaces https://github.com/mozilla/firefox-ui-tests/pull/89


jmaher:
I don't understand how my changes to add a new test is causing a different test to fail.  I spend 15 minutes trying to run the tests locally with no luck :(  I give up.

jmaher:
this failure on the test is in a different test/directory/etc.  Unrelated to adding a new test with no library or other modifications.

firefox_ui_tests/remote/security/test_untrusted_connection_error_page.py
```diff
@@ -0,0 +1,35 @@
...
+from marionette_driver.errors import MarionetteException
+
+from firefox_ui_harness.testcase import FirefoxTestCase
+
+
+class TestUntrustedConnectionErrorPage(FirefoxTestCase):
+    def setUp(self):
+        FirefoxTestCase.setUp(self)
+
+        self.test_url = 'https://ssl-selfsigned.mozqa.com'
```
whimboo: In all of our tests we name this `self.url` for single URLs. Please follow that so we can keep this style.

firefox_ui_tests/remote/security/test_untrusted_connection_error_page.py
```diff
@@ -0,0 +1,35 @@
...
+    def test_untrusted_connection_error_page(self):
+        # Test the GetMeOutOfHereButton from an Untrusted Error page
+        with self.marionette.using_context('content'):
+            self.assertRaises(MarionetteException,
+                              self.marionette.navigate, self.test_url)
+            # Wait for the DOM to receive events
+            time.sleep(1)
+
+            button = self.marionette.find_element(
+                By.ID, "getMeOutOfHereButton")
```
whimboo: This should fit into a single line. We have a max limit of 100 chars and not 80.

firefox_ui_tests/remote/security/test_untrusted_connection_error_page.py
```diff
@@ -0,0 +1,35 @@
...
+            self.assertTrue(button.is_displayed())
```
whimboo: I don't think this is necessary. find_element will fail with an exception if the button does not exist. That is what our Mozmill test is covering. The case when it is hidden we can get if the `click` event is not working. Also as we have seen lately `is_displayed` is not reliable and should not be used. So lets simply remove this line until we have a proper method to check for the displayed status.

firefox_ui_tests/remote/security/test_untrusted_connection_error_page.py
jmaher: thanks, I have updated my patch!

firefox_ui_tests/remote/security/test_untrusted_connection_error_page.py
jmaher: updated the PR with the feedback!

firefox_ui_tests/remote/security/test_untrusted_connection_error_page.py
```diff
@@ -0,0 +1,35 @@
...
+            button.click()
+            self.wait_for_condition(
+                lambda mn: self.browser.default_homepage in mn.get_url())
```
whimboo: This will not work for locales which have a redirect as their default home page. What we have to do here is to wait until the target page has been loaded, whereby this is still a missing piece in our tests. So maybe better is to run `self.browser.get_final_url(self.browser.default_homepage)` at the very first in this test and use the target page as reference for this check, which can then even be a `==` and not `in`.

firefox_ui_tests/remote/security/test_untrusted_connection_error_page.py
```diff
@@ -0,0 +1,31 @@
...
+    def test_untrusted_connection_error_page(self):
+        # Test the GetMeOutOfHereButton from an Untrusted Error page
+        with self.marionette.using_context('content'):
+            self.assertRaises(MarionetteException, self.marionette.navigate, self.url)
+            # Wait for the DOM to receive events
+            time.sleep(1)
+
+            button = self.marionette.find_element(By.ID, "getMeOutOfHereButton")
+            button.click()
+            self.wait_for_condition(lambda mn: self.browser.default_homepage in mn.get_url())
```
whimboo: Also please use `Wait()` here. The old `wait_for_condition` method is not that reliable and should not be used anymore.

jmaher:
this fails the tests and is easy to reproduce locally.  It is all related to the call to self.browser.get_final_url.

firefox_ui_tests/remote/security/test_untrusted_connection_error_page.py
```diff
@@ -0,0 +1,34 @@
...
+
+        # Test the GetMeOutOfHereButton from an Untrusted Error page
+        with self.marionette.using_context('content'):
+            self.assertRaises(MarionetteException, self.marionette.navigate, self.url)
+            # Wait for the DOM to receive events
+            time.sleep(1)
+
+            button = self.marionette.find_element(By.ID, "getMeOutOfHereButton")
+            button.click()
+            Wait(lambda mn: target_url == mn.get_url())
```
whimboo: The problem here is how you call this. `Wait()` is a constructor and requires a reference to marionette. On this class there is a `until()` method, which wants a callback for the condition. With this fixed it should work.

jmaher:
it looks like @galgeek has updated this in a different pull request!  Thanks :)

PR #136, opened 2015-03-20
Bug 1132624 - Convert Mozmill testEnablePrivilege

galgeek:
This short test conversion is blocked for the moment by security preference settings. There's more information in the IRC logs for today, March 19th, for both #ateam and #automation. 

I've pushed up my WIP here in case the block takes long to resolve.

https://bugzilla.mozilla.org/show_bug.cgi?id=1132624


chmanchester:
Cleaning up a few old PRs. We can take this up when https://bugzilla.mozilla.org/show_bug.cgi?id=1149618 is resolved.

PR #137, opened 2015-03-20
Minor fix to README.md - added correct directory path for running setup.py after Git cloning fire...

sr-murthy:
...fox-ui-tests


README.md
```diff
@@ -17,7 +17,7 @@ Then:
 Alternatively you may want to clone the repo for development:
 
     git clone https://github.com/mozilla/firefox-ui-tests.git
-    python setup.py develop
+    python firefox-ui-tests/setup.py develop
```
whimboo: If you want to work on this repository you would have to chdir into the directory. So I would leave the line as is but prefix it like `cd firefox-ui-tests && python setup.py develop`.

sr-murthy: OK

PR #138, opened 2015-03-20
Bug 1145677 - Ensure that the default new tab webpage is loaded in setUp() of FirefoxTestCase. r=chmanchester

whimboo:


firefox_ui_harness/testcase.py
```diff
@@ -23,6 +23,13 @@ def setUp(self, *args, **kwargs):
         self.marionette.set_context('chrome')
         self.browser = self.windows.current
 
+        with self.marionette.using_context(self.marionette.CONTEXT_CONTENT):
+            # Ensure that we have a default page opened
+            # Bug 1145668 - We currently cannot use 'about:blank' because it errors
+            # if about:error was loaded before
+            # self.marionette.navigate(self.prefs.get_pref('browser.newtab.url'))
```
chmanchester: Get rid of the commented line.

whimboo: I don't think that was the right solution. Instead we should tweak the preference value for our needs. I will have an update for it in a bit.

whimboo:
PR got merged as https://github.com/mozilla/firefox-ui-tests/commit/f8050f8de317f94fec953bf96618ae242f6324f5

PR #139, opened 2015-03-22
Bug 1132701 - Convert Mozmill testSSLStatusAfterRestart

galgeek:
https://bugzilla.mozilla.org/show_bug.cgi?id=1132701

It looks like this test should use the restart method added in PR #122. Until that lands, I've commented out that part of the code to let Travis test the rest.

The Mozmill test defines three separate but nearly identical helper methods to handle the page info window for each test item. I've combined them here, adding the item's url to each assertion's message. I wonder if I've missed some reason for separating these as the Mozmill test does.

### DIFF
--- a/firefox_ui_harness/runners/base.py
+++ b/firefox_ui_harness/runners/base.py
@@ -24,7 +24,7 @@ DEFAULT_PREFS = {
     'browser.search.update': False,
     'browser.sessionstore.resume_from_crash': False,
     'browser.shell.checkDefaultBrowser': False,
-    'browser.startup.page': 0,
+    'browser.startup.page': 3,
     'browser.tabs.animate': False,
     'browser.tabs.warnOnClose': False,
     'browser.tabs.warnOnOpen': False,
--- a/firefox_ui_harness/runners/base.py
+++ b/firefox_ui_harness/runners/base.py
@@ -24,7 +24,7 @@ DEFAULT_PREFS = {
     'browser.search.update': False,
     'browser.sessionstore.resume_from_crash': False,
     'browser.shell.checkDefaultBrowser': False,
-    'browser.startup.page': 3,
+    'browser.startup.page': 0,
     'browser.tabs.animate': False,
     'browser.tabs.warnOnClose': False,
     'browser.tabs.warnOnOpen': False,
--- a/firefox_ui_tests/remote/security/test_ssl_status_after_restart.py
+++ b/firefox_ui_tests/remote/security/test_ssl_status_after_restart.py
@@ -63,30 +63,32 @@ class TestSSLStatusAfterRestart(FirefoxTestCase):
 
         i = 0
         for item in self.test_data:
-            self.browser.tabbar.switch_to(i)
+            self.browser.tabbar.tabs[i].switch_to()
             self.verify_certificate_status(item)
             i = i + 1
         '''
 
     def verify_certificate_status(self, item):
-        url = item['url']
+        url, identity, type = item['url'], item['identity'], item['type']
 
         # Check the favicon
         # TODO: find a better way to check, e.g., mozmill's isDisplayed
-        self.assertEqual(self.browser.navbar.locationbar.favicon.get_attribute('hidden'), 'false',
-                         'Lock icon is visible in identity box for ' + url)
+        favicon_hidden = self.marionette.execute_script("""
+          return arguments[0].hasAttribute("hidden");
+        """, script_args=[self.browser.navbar.locationbar.favicon])
+        self.assertFalse(favicon_hidden)
 
         self.identity_popup.box.click()
         Wait(self.marionette).until(lambda _: self.identity_popup.is_open)
 
         # Check the type shown on the idenity popup doorhanger
         self.assertEqual(self.identity_popup.popup.get_attribute('className'),
-                         item['type'],
+                         type,
                          'Extended certificate is verified for ' + url)
 
         # Check the identity label
         self.assertEqual(self.identity_popup.organization_label.get_attribute('value'),
-                         item['identity'],
+                         identity,
                          'Identity name is correct for ' + url)
 
         # Get the information from the certificate
@@ -96,7 +98,7 @@ class TestSSLStatusAfterRestart(FirefoxTestCase):
         page_info = self.browser.open_page_info_window(
             lambda _: self.identity_popup.more_info_button.click())
 
-        if item['identity'] != '':
+        if identity != '':
             owner = cert['organization']
         else:
             owner = page_info.get_property('securityNoOwner')
--- a/firefox_ui_tests/remote/security/test_ssl_status_after_restart.py
+++ b/firefox_ui_tests/remote/security/test_ssl_status_after_restart.py
@@ -40,6 +40,7 @@ class TestSSLStatusAfterRestart(FirefoxTestCase):
         try:
             self.windows.close_all([self.browser])
             self.browser.tabbar.close_all_tabs([self.browser.tabbar.tabs[0]])
+            self.browser.switch_to()
             self.identity_popup.close(force=True)
         finally:
             FirefoxTestCase.tearDown(self)
@@ -116,3 +117,4 @@ class TestSSLStatusAfterRestart(FirefoxTestCase):
         self.assertEqual(page_info.deck.security.verifier.get_attribute('value'),
                          cert['issuerOrganization'],
                          'Verifier matches issuer of certificate for ' + url)
+        page_info.close()
--- a/firefox_ui_tests/remote/security/test_ssl_status_after_restart.py
+++ b/firefox_ui_tests/remote/security/test_ssl_status_after_restart.py
@@ -1,0 +1,123 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from marionette_driver import Wait
+from marionette_driver.errors import NoSuchElementException
+
+from firefox_ui_harness.decorators import skip_if_e10s, skip_under_xvfb
+from firefox_ui_harness.testcase import FirefoxTestCase
+
+
+class TestSSLStatusAfterRestart(FirefoxTestCase):
+
+    def setUp(self):
+        FirefoxTestCase.setUp(self)
+
+        self.test_data = [
+            {
+                'url': 'https://ssl-dv.mozqa.com',
+                'identity': '',
+                'type': 'verifiedDomain'
+            },
+            {
+                'url': 'https://ssl-ev.mozqa.com/',
+                'identity': 'Mozilla Corporation',
+                'type': 'verifiedIdentity'
+            },
+            {
+                'url': 'https://ssl-ov.mozqa.com/',
+                'identity': '',
+                'type': 'verifiedDomain'
+            }
+        ]
+
+        # Set browser to restore previous session
+        self.prefs.set_pref('browser.startup.page', 3)
+
+        self.identity_popup = self.browser.navbar.locationbar.identity_popup
+
+    def tearDown(self):
+        try:
+            self.identity_popup.close(force=True)
+            self.browser.tabbar.close_all_tabs([self.browser.tabbar.tabs[0]])
+        except NoSuchElementException:
+            # TODO: A NoSuchElementException may be thrown here when the test is skipped
+            # as under xvfb.
+            pass
+        finally:
+            FirefoxTestCase.tearDown(self)
+
+    @skip_if_e10s
+    @skip_under_xvfb
+    # Bug 995801: Test certificate status after browser restart
+    def test_ssl_status_after_restart(self):
+        for item in self.test_data:
+            with self.marionette.using_context('content'):
+                self.marionette.navigate(item['url'])
+            self.verify_certificate_status(item)
+            self.browser.tabbar.open_tab()
+
+        '''
+        self.restart()
+
+        i = 0
+        for item in self.test_data:
+            self.browser.tabbar.switch_to(i)
+            self.verify_certificate_status(item)
+            i = i + 1
+        '''
+
+    def verify_certificate_status(self, item):
+        url = item['url']
+
+        # Check the favicon
+        # TODO: find a better way to check, e.g., mozmill's isDisplayed
+        self.assertEqual(self.browser.navbar.locationbar.favicon.get_attribute('hidden'), 'false',
+                         'Lock icon is visible in identity box for ' + url)
+
+        self.identity_popup.box.click()
+        Wait(self.marionette).until(lambda _: self.identity_popup.is_open)
+
+        # Check the type shown on the idenity popup doorhanger
+        self.assertEqual(self.identity_popup.popup.get_attribute('className'),
+                         item['type'],
+                         'Extended certificate is verified for ' + url)
+
+        # Check the identity label
+        self.assertEqual(self.identity_popup.organization_label.get_attribute('value'),
+                         item['identity'],
+                         'Identity name is correct for ' + url)
+
+        # Get the information from the certificate
+        cert = self.browser.tabbar.selected_tab.certificate
+
+        # Open the Page Info window by clicking the More Information button
+        page_info = self.browser.open_page_info_window(
+            lambda _: self.identity_popup.more_info_button.click())
+
+        if item['identity'] != '':
+            owner = cert['organization']
+        else:
+            owner = page_info.get_property('securityNoOwner')
+
+        try:
+            # Verify that the current panel is the security panel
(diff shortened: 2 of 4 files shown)


firefox_ui_tests/remote/security/test_ssl_status_after_restart.py
```diff
@@ -0,0 +1,123 @@
...
+        except NoSuchElementException:
+            # TODO: A NoSuchElementException may be thrown here when the test is skipped
+            # as under xvfb.
+            pass
+        finally:
+            FirefoxTestCase.tearDown(self)
+
+    @skip_if_e10s
+    @skip_under_xvfb
+    # Bug 995801: Test certificate status after browser restart
```
whimboo: This is not a valid reference anymore. I don't think its necessary at all.

firefox_ui_tests/remote/security/test_ssl_status_after_restart.py
```diff
@@ -0,0 +1,123 @@
...
+                self.marionette.navigate(item['url'])
+            self.verify_certificate_status(item)
+            self.browser.tabbar.open_tab()
+
+        '''
+        self.restart()
+
+        i = 0
+        for item in self.test_data:
+            self.browser.tabbar.switch_to(i)
```
whimboo: switch_to() wants a handle but not an index. What you want instead is selected_index.

firefox_ui_tests/remote/security/test_ssl_status_after_restart.py
```diff
@@ -0,0 +1,123 @@
...
+            self.verify_certificate_status(item)
+            i = i + 1
+        '''
+
+    def verify_certificate_status(self, item):
+        url = item['url']
+
+        # Check the favicon
+        # TODO: find a better way to check, e.g., mozmill's isDisplayed
+        self.assertEqual(self.browser.navbar.locationbar.favicon.get_attribute('hidden'), 'false',
```
whimboo: Similar to the other PR. This has to be `hasAttribute`.

firefox_ui_tests/remote/security/test_ssl_status_after_restart.py
```diff
@@ -0,0 +1,123 @@
...
+
+        i = 0
+        for item in self.test_data:
+            self.browser.tabbar.switch_to(i)
+            self.verify_certificate_status(item)
+            i = i + 1
+        '''
+
+    def verify_certificate_status(self, item):
+        url = item['url']
```
whimboo: You can split this like `url, identity, type = item`

firefox_ui_tests/remote/security/test_ssl_status_after_restart.py
```diff
@@ -0,0 +1,123 @@
...
+        cert = self.browser.tabbar.selected_tab.certificate
+
+        # Open the Page Info window by clicking the More Information button
+        page_info = self.browser.open_page_info_window(
+            lambda _: self.identity_popup.more_info_button.click())
+
+        if item['identity'] != '':
+            owner = cert['organization']
+        else:
+            owner = page_info.get_property('securityNoOwner')
```
whimboo: This is a change in behavior compared to the mozmill tests. Can you explain why this is necessary?

whimboo: Oh, I see. You have combined all the individual helper methods. Nice!

firefox_ui_tests/remote/security/test_ssl_status_after_restart.py
```diff
@@ -0,0 +1,123 @@
...
+
+        try:
+            # Verify that the current panel is the security panel
+            self.assertEqual(page_info.deck.selected_panel, page_info.deck.security)
+
+            # Verify the domain listed on the security panel
+            self.assertIn(self.security.get_domain_from_common_name(cert['commonName']),
+                          page_info.deck.security.domain.get_attribute('value'),
+                          'Expected domain found in certificate for ' + url)
```
whimboo: So we don't need all this regex code for DV here anymore?

firefox_ui_tests/remote/security/test_ssl_status_after_restart.py
galgeek: I believe it was this bug that prompted the creation of this test.

I agree, it's not necessary here.

firefox_ui_tests/remote/security/test_ssl_status_after_restart.py
galgeek: tabbar's `switch_to` claims and looks like it will handle an index. It wraps `self.browser.tabbar.tabs[index].switch_to()` which is easy enough to use here instead.

firefox_ui_tests/remote/security/test_ssl_status_after_restart.py
galgeek: Sure. Until my last revision, I was using identity and type only once.

firefox_ui_tests/remote/security/test_ssl_status_after_restart.py
galgeek: I'm not sure, in fact.

The regex code in the DV method here and in the DV test looks like it tests for the entire host name against a (possibly wildcard) cert.

Here we're testing only that the domain name is in Page Info's domain value, so we don't need to worry about a wildcard cert, but we're also not making as stringent a test. This is the same test that's present in the EV test.

galgeek:
Thanks for your feedback, Henrik!

Note that I've made some comments in line notes above.

I'll request feedback/review again after self.restart() is available.

firefox_ui_tests/remote/security/test_ssl_status_after_restart.py
whimboo: This only switches the handle to the given tab but does not focus it! I think it would be better here to actually focus the underlying tab we test.

firefox_ui_tests/remote/security/test_ssl_status_after_restart.py
whimboo: I checked it myself and our https://ssl-ov.mozqa.com domain is using a wildcard certificate. So we would still need it in such a case.

whimboo:
> I'll request feedback/review again after self.restart() is available.

For testing you could copy the logic for now, which we can replace very easily then.

firefox_ui_tests/remote/security/test_ssl_status_after_restart.py
galgeek: The current check handles a wildcard certificate, and it's running fine for the -ov item. It checks that the **domain** displayed on page info's security tab matches the **domain** on the certificate, in its common name field. This is the same check that's used in both the mozmill and marionette -ev test. (The mozmill test's separate -ov check doesn't test the domain at all.)

The mozill -dv tests' regex code checks that the **hostname** displayed on page info's security tab matches the certificate's common name field as a whole, which can include either a specific host or a wildcard for the domain.

I'm not sure why we'd want to do something different for the -dv test than for the others. Do we need or want to check for hostname rather than just for domain for each item?

galgeek:
I've pulled upstream, run setup.py develop again, and rebased on master, but when I try using the code from the update tests PR, I get this error:

```
0:05.72 TEST_END: MainThread ERROR, expected PASS
Traceback (most recent call last):
  File "/Users/bara/.virtualenvs/ffuit/lib/python2.7/site-packages/marionette_client-0.9-py2.7.egg/marionette/marionette_test.py", line 290, in run
    testMethod()
  File "/Users/bara/Dev/ffuit/firefox_ui_harness/decorators.py", line 15, in wrapper
    return target(self, *args, **kwargs)
  File "/Users/bara/Dev/ffuit/firefox_ui_harness/decorators.py", line 23, in wrapper
    return target(self, *args, **kwargs)
  File "/Users/bara/Dev/ffuit/firefox_ui_tests/remote/security/test_ssl_status_after_restart.py", line 61, in test_ssl_status_after_restart
    self.marionette.restart(in_app = True)
TypeError: restart() got an unexpected keyword argument 'in_app'
```

firefox_ui_harness/runners/base.py
```diff
@@ -24,7 +24,7 @@
     'browser.search.update': False,
     'browser.sessionstore.resume_from_crash': False,
     'browser.shell.checkDefaultBrowser': False,
-    'browser.startup.page': 0,
+    'browser.startup.page': 3,
```
chmanchester: I'm assuming this was 0 before for a reason, so I don't think we can change this here.

firefox_ui_tests/remote/security/test_ssl_status_after_restart.py
```diff
@@ -0,0 +1,127 @@
...
+            },
+            {
+                'url': 'https://ssl-ov.mozqa.com/',
+                'identity': '',
+                'type': 'verifiedDomain'
+            }
+        ]
+
+        # Set browser to restore previous session
+        self.prefs.set_pref('browser.startup.page', 3)
```
chmanchester: If this isn't sufficient for the test because because the pref doesn't hang around after the restart, you can try `self.marionette.instance.prefs['browser.startup.page'] = 3` in addition (although you'll have to set that back at the end of the test yourself).

firefox_ui_tests/remote/security/test_ssl_status_after_restart.py
```diff
@@ -0,0 +1,127 @@
...
+            new_tab.select()
+
+        # TODO: Bug 1148220 add ability to store open tabs to restart method
+        self.marionette.execute_script("""
+          Components.utils.import("resource://gre/modules/Services.jsm");
+          let cancelQuit = Components.classes["@mozilla.org/supports-PRBool;1"]
+                                     .createInstance(Components.interfaces.nsISupportsPRBool);
+          Services.obs.notifyObservers(cancelQuit, "quit-application-requested", null);
+        """)
+        self.restart()
```
chmanchester: I don't think there's any harm in putting this in the firefox-ui-tests' restart method already.

firefox_ui_tests/remote/security/test_ssl_status_after_restart.py
```diff
@@ -0,0 +1,127 @@
...
+
+        i = 0
+        for item in self.test_data:
```
chmanchester: You can use python's `enumerate` here, which binds names to and indexes and items iterating over a loop, so:

```python
>>> arr = [1, 2, 3]
>>> for index, item in enumerate(arr):
...     print "index: %s, item: %s" % (index, item)
... 
index: 0, item: 1
index: 1, item: 2
index: 2, item: 3
```

firefox_ui_tests/remote/security/test_ssl_status_after_restart.py
```diff
@@ -0,0 +1,127 @@
...
+            self.browser.tabbar.tabs[i].select()
+            self.verify_certificate_status(item)
+            i = i + 1
+
+    def verify_certificate_status(self, item):
+        url, identity, type = item['url'], item['identity'], item['type']
```
chmanchester: If you use an array of tuples rather than an array of dicts above, this could just be
```python
url, identity, type = item
```
Of course that's a matter of style and personal discretion and totally optional :)

chmanchester: It's usually advisable not to have variables names shadow a builtin ("type" is a python builtin).

firefox_ui_tests/remote/security/test_ssl_status_after_restart.py
```diff
@@ -0,0 +1,127 @@
...
+            owner = page_info.get_property('securityNoOwner')
+
+        self.assertEqual(page_info.deck.security.owner.get_attribute('value'), owner,
+                         'Expected owner label found for ' + url)
+
+        # Verify the verifier listed on the security panel
+        self.assertEqual(page_info.deck.security.verifier.get_attribute('value'),
+                         cert['issuerOrganization'],
+                         'Verifier matches issuer of certificate for ' + url)
+        page_info.close()
```
chmanchester: Let's close this in tearDown.

firefox_ui_tests/remote/security/test_ssl_status_after_restart.py
chmanchester: This question seems unresolved, and significant to the meaning of this test. I think we ought to figure this out before landing.

firefox_ui_tests/remote/security/test_ssl_status_after_restart.py
```diff
@@ -0,0 +1,127 @@
...
+
+        # Verify that the current panel is the security panel
+        self.assertEqual(page_info.deck.selected_panel, page_info.deck.security)
+
+        # Verify the domain listed on the security panel
+        # If this is a wildcard cert, use only the domain
+        if '*' in cert['commonName'][0]:
+            cert_name = self.security.get_domain_from_common_name(cert['commonName'])
+        else:
+            cert_name = cert['commonName']
```
chmanchester: There was a note somewhere this is a little different from the mozmill test. I haven't had an opportunity to look through the mozmill test thoroughly but I will before reviewing this again.

firefox_ui_tests/remote/security/test_ssl_status_after_restart.py
galgeek: Thanks! I'm not sure I've seen enumerate before, that's very handy.

firefox_ui_harness/runners/base.py
galgeek: browser.startup.page: 3 restores the previous session. I added it here when I was initially troubleshooting restart... it's gone now.

firefox_ui_tests/remote/security/test_ssl_status_after_restart.py
galgeek: It's easier to see what's going on with this in place, but it's true, it's not apparently necessary here. I've added the window.close that the page info unit test uses to tearDown.

firefox_ui_tests/remote/security/test_ssl_status_after_restart.py
galgeek: I like the idea of using a tuple, but I wonder if the dict isn't more readable, especially if people are more familiar with JavaScript than Python.

firefox_ui_tests/remote/security/test_ssl_status_after_restart.py
galgeek: The mozmill test does something different for -ev vs. -dv and checks nothing for -ov.

I've reworked this a little, and it should now be doing at least as much as the mozmill test does and without defining three separate functions to do it.

firefox_ui_tests/remote/security/test_ssl_status_after_restart.py
galgeek: I'll leave it here for now.

galgeek:
Travis's Linux worker failed early in the build today, not in the tests.

firefox_ui_tests/remote/security/test_ssl_status_after_restart.py
chmanchester: I expect multiple tests will need this, why not move to a shared location?

firefox_ui_tests/remote/security/test_ssl_status_after_restart.py
```diff
@@ -0,0 +1,125 @@
...
+        # Open the Page Info window by clicking the More Information button
+        page_info = self.browser.open_page_info_window(
+            lambda _: self.identity_popup.more_info_button.click())
+
+        # Verify that the current panel is the security panel
+        self.assertEqual(page_info.deck.selected_panel, page_info.deck.security)
+
+        # Verify the domain listed on the security panel
+        # If this is a wildcard cert, check only the domain
+        if '*' == cert['commonName'][0]:
```
chmanchester: This could be `.startswith('*')`

firefox_ui_tests/remote/security/test_ssl_status_after_restart.py
```diff
@@ -0,0 +1,125 @@
...
+            self.assertIn(self.security.get_domain_from_common_name(cert['commonName']),
+                          page_info.deck.security.domain.get_attribute('value'),
+                          'Expected domain found in certificate for ' + url)
```
chmanchester: I checked on the eTLD service and this looks like it does the right thing. Nice work!

chmanchester:
Travis is failing due to a change we made in tree (unrelated to this patch). I'll get this straightened out.

firefox_ui_tests/remote/security/test_ssl_status_after_restart.py
chmanchester: Sorry, the change I requested here isn't right. This should be left in and we should additionally close errant windows in teardown.

chmanchester:
https://github.com/mozilla/firefox-ui-tests/commit/d14a060c9e8f0ca11024bfbb1ec9f9bab9e1d227

PR #146, opened 2015-03-31
Bug 1149818 - Set the homepage_override_url to 'about:blank' so we don't...

chmanchester:
... leak a tab when updating across verion numbers.;r=ahal


chmanchester:
https://github.com/mozilla/firefox-ui-tests/commit/f0b62348c0093e39041263e3f014ce18339f5a61

PR #148, opened 2015-04-01
Bug 1149720 update for intermittent OS X failures in test_bookmarks

galgeek:
This replaces a wait_for_condition that's timing out and waits for self.places.is_bookmark_star_button_ready() before the button click, instead of asserting it after.

But maybe we also need to wait for self.places.is_bookmark_star_button_ready() after the click?


chmanchester:
https://github.com/mozilla/firefox-ui-tests/commit/6ad328d77e62f1ae6a5e0dc61b729bec7dc44c33

PR #152, opened 2015-04-15
Bug 1154810 - Update marionette dependencies

mjzffr:
@chmanchester r?


chmanchester:
For some reason this breaks update tests, I'll look into that this afternoon.

chmanchester:
Hm, tried again and this works fine..

PR #154, opened 2015-04-16
Create development version until we release the next version

armenzg:
This helps me to prepare for the 0.3 release, which, I would like it to contain all changes needed for release update process (among other changes).

That way I can keep a 0.3.dev0 python package on pypi.pub.build.m.o without causing confusion since everyone would know that is wip.


armenzg:
We will open another request once we have the final version ready.

PR #156, opened 2015-04-20
Bug 1156475 - Avoid global timeout for update tests that fail when the a...

chmanchester:
...bout window is focused by switching back to the main browser window before allowing the failure to propagate.


armenzg:
My apologies about this delay. For some reason I saw a review request on the bug for ato and I thought it was waiting on him. Please ship it!

PR #159, opened 2015-04-25
Add usage examples to toolbars documentation

utvar:
Hello, @chmanchester, @galgeek suggested I could get started contributing to Mozilla by adding usage examples to the Firefox Puppeteer documentation. Would this be helpful, and does this look like a good start?


galgeek:
Hi, @chmanchester and/or @whimboo!

I know that getting update tests working is high priority now. 

Maybe one of you could let me or @utvar know when you have a bit more time if this is a good first project? Thank you!

whimboo:
That's a good idea. The problem right now is that we only generate API reference documentation. If you compare our docs with e.g. [mozbase](http://mozbase.readthedocs.org/) you will see that. So for examples I would more prefer to get all of those added with extra documentation directly in sphinx files and not in the code. That would be a clean separation, and most likely easier to implement.

galgeek:
Thank you, @whimboo!

I may see @utvar Saturday, and we'll see if we can get started on something more like what you describe.

Let us know here if you have any more specific ideas.

utvar:
@whimboo, is this closer to what you had in mind?

whimboo:
The code examples we should directly add to the documentation and not reference it to a unittest in the firefox-ui-tests repository. For example see the marionette-client docs: http://marionette-client.readthedocs.org/en/latest/basics.html#session-management

It would also be good to add some introductory text to each of the classes, which describes what it is and e.g. which properties (mostly the ui ones) it contains.

utvar:
Sorry, this is more than I can do, given my current understanding of the project, so I am closing this PR. Instead I'll work on updating the TOC in PR #210.

whimboo:
Ok, thanks a lot anyway.

PR #161, opened 2015-05-06
Bug 1162070 - TestSafeBrowsingWarningPages.test_warning_pages doesn't test the expected support URL anymore. r=chmanchester

whimboo:


firefox_ui_tests/remote/security/test_safe_browsing_warning_pages.py
```diff
@@ -68,31 +68,28 @@ def check_get_me_out_of_here_button(self, unsafe_page):
...
-            else:
-                url = self.marionette.execute_script("""
-                  Components.utils.import("resource://gre/modules/Services.jsm");
-                  return Services.urlFormatter.formatURLPref(
-                  "browser.safebrowsing.malware.reportURL") + arguments[0];
-                """, script_args=[unsafe_page])
+        # TODO: Bug 1140470: use replacement for mozmill's waitforPageLoad
+        self.wait_for_condition(lambda mn: mn.execute_script("""
+          return document.readyState == 'complete';
+        """))
```
chmanchester: I don't see how this is achieving something different from waiting for the urls to be the same, but either way this reflects the browser change and is still effective.

whimboo: The problem is the part where you called `get_final_url()`. This helper method loads the specified url and waits for the loading be completed. But exactly that unloads the webpage which was loaded by the button click. So your `Wait().until()` call operates on the exact same loaded URL and always passes in the first iteration. With the assert it first retrieves the current URL, and then constructs the final URL. That way we will notice if the wanted URL has been loaded or not.

PR #168, opened 2015-05-13
Bug 1164450 - Create Python helper script to setup a virtual environment to run Firefox UI Tests. r=chmanchester

whimboo:


create_venv.py
```diff
@@ -0,0 +1,124 @@
...
+
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+"""
+The script can be used to setup a virtual environment for running Firefox UI Tests.
+It will automatically install the firefox ui test package, all its dependencies, and
+auxiliary packages like mozdownload.
+"""
```
jmaher: I don't understand what dependencies are required and which are optional.  It looks like mozdownload is optional.

README.md
```diff
@@ -16,6 +16,8 @@ Then:
     cd firefox-ui-tests
     python setup.py develop
 
+If you do not have installed virtualenv and virtualenvwrapper you can also use the create_venv.py script to let it automatically create a virtual environmnet for you.
+
```
jmaher: this isn't clear on how to get into a virtualenv.  The script being added is great for downloading and installing virtualenv, but for actually creating  a virtualenv it doesn't look simple.  Could we extend this new script to use an existing one, create one if virtualenv exists, or do what it does if all else fails?

create_venv.py
whimboo: Please see the requirements.txt file, where optional packages are listed.

README.md
whimboo: If you already have a virtual environment this script will not be helpful for you, unless you want to replace it. This is really for the CI system and basic starters who do not know how to handle virtual environments. Most of us will even work with virtualenvwrapper, and in that case the location can even not be easily determined by the script. So I don't think it makes sense so cover all those cases.

I will tweak the docs here and add an example.

create_venv.py
```diff
@@ -0,0 +1,127 @@
...
+        try:
+            os.remove(zip_path)
+        except OSError:
+            pass
+
+        shutil.rmtree(os.path.dirname(script_path), ignore_errors=True)
+
+
+def main():
+    parser = optparse.OptionParser('Usage: %prog [options] path_to_venv')
```
chmanchester: Is there some reason optparse is necessary? We should use argparse in new code whenever possible.

create_venv.py
```diff
@@ -0,0 +1,127 @@
...
+    parser.add_option('-p', '--python',
```
chmanchester: Is exposing this option necessary? Are we going to end up with a more trusted python installation on the machine than the one we're running this script with?

create_venv.py
```diff
@@ -0,0 +1,127 @@
...
+                      dest='with_requirements',
+                      action='store_true',
+                      help='Install optional packages as listed in requirements.txt')
+
+    (options, args) = parser.parse_args(args=None, values=None)
+
+    if len(args) != 1:
+        parser.error('Path to the environment has to be specified')
+    target = args[0]
+    assert(target)
```
chmanchester: assert is a syntactic form in python, not a function. Lose the parens.

create_venv.py
```diff
@@ -0,0 +1,127 @@
...
+    parser = optparse.OptionParser('Usage: %prog [options] path_to_venv')
+    parser.add_option('-p', '--python',
+                      type='string',
+                      dest='python',
+                      metavar='BINARY',
+                      help='The Python interpreter to use.')
+    parser.add_option('--with-requirements',
+                      dest='with_requirements',
+                      action='store_true',
+                      help='Install optional packages as listed in requirements.txt')
```
chmanchester: It's funny to refer to these as "optional requirements". I don't have a problem making pep8 and mozdownload requirements for the firefox-ui-tests, but if not maybe find a way to refer to them that indicates they're for setting up the environment rather than running the tests themselves.

create_venv.py
```diff
@@ -0,0 +1,127 @@
...
+    tps_env = os.path.join(target, venv_activate_this)
+    execfile(tps_env, dict(__file__=tps_env))
+
+    # Install Firefox UI Tests package
+    subprocess.check_call([os.path.join(target, venv_python_bin),
+                           os.path.join(here, 'setup.py'), 'develop'])
+
+    # Install optional packages via requirements.txt
+    if options.with_requirements:
+        subprocess.check_call(['pip', 'install', '-r',
```
chmanchester: So we still have to run a script to install pip before we run this script? If we're going to bootstrap with this script (which doesn't use pip to get virtualenv), should we start by installing pip and then use if for everything else?

create_venv.py
whimboo: I don't want to mix optparse and argparse in the same project. As long as Marionette is not using argparse I would also stay with optparse here.

create_venv.py
whimboo: This gives us the option to setup a virtual environment for different versions of Python. Very helpful on Travis if we decide to run with e.g. Python3.

create_venv.py
whimboo: If that is the case, I will make them hard requirements. I only thought that you might have problems with that. Looks like I was wrong. :)

create_venv.py
whimboo: No, you don't have. At this point we are inside the virtualenv and setuptools and pip will be available.

create_venv.py
chmanchester: I don't think this is a valid reason not to use argparse. Marionette is moving soon anyway.

PR #169, opened 2015-05-13
Bug 1164637 - Create entry scripts for functional and remote tests. r=chmanchester

whimboo:


firefox_ui_harness/options/functional.py
```diff
@@ -0,0 +1,18 @@
...
+
+from . import FirefoxUIOptions
+
+
+class FunctionalOptions(FirefoxUIOptions):
+
+    def parse_args(self, *args, **kwargs):
+        options, tests = FirefoxUIOptions.parse_args(self, *args, **kwargs)
+
+        tests = [firefox_puppeteer.manifest, firefox_ui_tests.manifest_functional]
```
chmanchester: So the new entry point never allows specifying tests? Can we just specify a manifest directly instead of making a new entry point? At least error out if some are specified since they're being overwritten.

whimboo: Different test types can require different options. You can see this with the update tests. For consistency we should have entry points for each test type, and not start to do a mix-up of entry points and CLI arguments. Good point regarding error out. I will see what I can do.

chmanchester: We already have the ability to specify tests from the command line. These entry points don't provide anything further. This is a configuration change, it doesn't need to be a code change.

whimboo: Actually a check for specified tests is not possible because FirefoxUIOptions is already setting the default tests in case none are given. So we will always have tests specified. We would have to do a refactoring of the classes, and move code around.

whimboo: Just to note... when it comes to reporting the results should end-up in different buckets for functional and remote tests. If we do not have separate entry scripts and classes, there will be no way to distinct those on the treeherder side. Keep in mind that we also have an entry script for update tests, and those also overwrite possible arguments. I'm going to update the command line usage docstring and strip of the arguments for the new entry scripts.

chmanchester: We can report them differently if they're run at different times. Mochitest devtools test use the same runtests.py script as mochitest plain and show up as different jobs in the UI. This made enough sense for update tests because those have additional setup steps and use a different runner.

whimboo: In that case I trust your experience in those other test suites and that it will work well with treeherder. That said lets keep what we have right now in our tests. I don't see value in changing it then unless we need a specific runner for a group of tests.

whimboo:
Closing this PR as wontfix for now given that it seems to make no sense to have those additional entry scripts.

PR #170, opened 2015-05-15
Move mozdownload and pep8 to optional downloads

whimboo:


jmaher:
drive by- this looks good.

PR #173, opened 2015-05-19
Bug 1166411 - firefox-ui-update should not require --update-target-version to be set if --update-target-buildid has been specified. r=jmaher

whimboo:


jmaher:
this is very simple and looks very safe!

PR #176, opened 2015-05-21
Bug 1167126 - Fix testCheckAboutPrivateBrowsing for correct window and tab handling. r=jmaher

whimboo:


firefox_ui_tests/functional/private_browsing/test_about_private_browsing.py
```diff
@@ -2,65 +2,54 @@
...
 class TestAboutPrivateBrowsing(FirefoxTestCase):
 
     def setUp(self):
         FirefoxTestCase.setUp(self)
-        self.pb_url = self.marionette.absolute_url('private_browsing/about.html?')
 
-    def tearDown(self):
-        self.marionette.close()
+        self.support_url = 'about:blank?'
+        self.pb_url = self.support_url + 'private-browsing'
```
jmaher: this seems overkill to have support_url and pb_url where as both are only used once.  I would rather see no variables here, but one would be fine.

Can you also comment why we need to set the preference to this value on teardown?  It isn't clear to me by reading the code

jmaher: oh, this is in setup- my bad.  why does support_url need a ? after it, that seems incorrect.

firefox_ui_tests/functional/private_browsing/test_about_private_browsing.py
```diff
@@ -2,65 +2,54 @@
...
+
+        self.prefs.set_pref('app.support.baseURL', self.support_url)
 
     @skip_if_e10s
```
jmaher: I know you didn't change this, but can we not do this in the manifest.ini file?  that is what we do for mochitests.

firefox_ui_tests/functional/private_browsing/test_about_private_browsing.py
```diff
@@ -2,65 +2,54 @@
...
-            top_html = self.marionette.find_element(By.TAG_NAME, 'html')
-            top_html.send_keys(self.keys.SHIFT, self.keys.ACCEL, access_key)
-
-        self.wait_for_condition(lambda mn: len(self.windows.all) == 2)
-        self.browser_pb = self.windows.switch_to(lambda win: win.is_private)
-        self.assertTrue(self.browser_pb.is_private)
+            status_node = self.marionette.find_element(By.CSS_SELECTOR, 'p.showNormal')
+            self.assertEqual(status_node.text,
+                             self.browser.get_entity('aboutPrivateBrowsing.subtitle.normal'),
+                             'Status text indicates we are in private browsing mode')
```
jmaher: do we care about this text being localized at all?

firefox_ui_tests/functional/private_browsing/test_about_private_browsing.py
whimboo: Both variables are used twice. I would agree to not bind ´support_url´ to the class but make it just a local variable. But we should really keep it for maintenance. You don't want to have to check the file which places hard-coded strings need to be replaced.

The question mark is part of the original URL which is stored in that pref. We overwrite that one, so we have to mimic that part. Its the parameter, which decides to load which site.

firefox_ui_tests/functional/private_browsing/test_about_private_browsing.py
whimboo: This is something we might indeed wanna do in the manifest. Feel free to file a bug with that proposal. Maybe we can find someone who wanna work on it.

firefox_ui_tests/functional/private_browsing/test_about_private_browsing.py
whimboo: This tests the localized text indirectly, but yes we cannot figure out if it is localized or not. We can also not check if the localized text has been correctly translated.

jmaher: thanks for replying to my questions; I will file a bug for the e10s in the
manifests bit.  all else is good assuming we have green tests :)

PR #180, opened 2015-05-26
Bug 1168545 - Add default preference to disable reader view intro

galgeek:
Adding this default preference may still be a good idea, though it doesn't alone enable functional/locationbar/test_access_locationbar to pass.

### DIFF
--- a/firefox_ui_harness/runners/base.py
+++ b/firefox_ui_harness/runners/base.py
@@ -19,6 +19,7 @@ DEFAULT_PREFS = {
     # can correctly handle error pages
     'browser.newtab.url': 'about:newtab',
     'browser.newtabpage.enabled': False,
+    'browser.reader.detectedFirstArticle': True,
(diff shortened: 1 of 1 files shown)


whimboo:
We definitely need that! Can you check when this came in? I wonder if we have to get this backported to Aurora too.

galgeek:
It looks like it's in Aurora, too:

http://mxr.mozilla.org/mozilla-aurora/search?string=readerview&find=\.js&findi=\.js&filter=^[^\0]*%24&hitlimit=&tree=mozilla-aurora

whimboo:
It does and it will even be included in the upcoming 38.0.5 release of Firefox. So we should get this backported down to the beta branch, which will be merged to release soon.

whimboo:
The PR got merged as https://github.com/mozilla/firefox-ui-tests/commit/594b6105b3198042288efd3e1a2c32cc7d609ae9.  Thanks Barbara!

PR #183, opened 2015-05-28
Bug 1131361 - Convert mozmill testFaviconInAutocomplete to marionette

galgeek:
WIP and not quite sure when I'll be able to look at it again. I've left details in comments in the code. 

The mozmill test runs and passes locally for me.

This marionette version doesn't find anything in the favicon element's src attribute; is it checking the same thing that the mozmill test checks? Or is there a significant difference in the environment that must be accounted for otherwise?


firefox_ui_tests/functional/locationbar/test_favicon_in_autocomplete.py
```diff
@@ -0,0 +1,87 @@
...
+        finally:
+            FirefoxTestCase.tearDown(self)
+
+    @skip_under_xvfb
+    def test_favicon_in_autocomplete(self):
+        # Open the test page
+        def load_urls():
+            with self.marionette.using_context('content'):
+                for url in self.test_urls:
+                    self.marionette.navigate(url)
```
whimboo: nit: There is only a single URL, so this for loop can be ripped off.

firefox_ui_tests/functional/locationbar/test_favicon_in_autocomplete.py
```diff
@@ -0,0 +1,87 @@
...
+        self.places.wait_for_visited(self.test_urls, load_urls)
+
+        locationbar = self.browser.navbar.locationbar
+        favicon = self.browser.navbar.locationbar.favicon
+        # this code assigns locationbar_favicon_url an empty string; maybe a sleep will help? no
+        time.sleep(5)
```
whimboo: Do you mean the favicon cannot be found?

firefox_ui_tests/functional/locationbar/test_favicon_in_autocomplete.py
```diff
@@ -0,0 +1,87 @@
...
+        locationbar_favicon_url = self.marionette.execute_script("""
+          return arguments[0].getAttribute("src");
+        """, script_args=[favicon])
```
whimboo: Wait, I see now. An empty string for the src attribute? For that you should also be able to use `favicon.get_attribute()`.

firefox_ui_tests/functional/locationbar/test_favicon_in_autocomplete.py
```diff
@@ -0,0 +1,87 @@
...
+        print 'locationbar_favicon_url: %s' % locationbar_favicon_url
+        print 'favicon_image: %s' % favicon_image
+
+        # Clear the location bar, type the test string, check that autocomplete list opens
+        locationbar.clear()
+        locationbar.urlbar.send_keys(self.test_string)
+        self.assertEqual(locationbar.value, self.test_string)
+        Wait(self.marionette).until(lambda _: self.autocomplete_results.is_open)
+
+        result = self.autocomplete_results.visible_results[0]
```
whimboo: This won't work anymore given that the first entry will be the search now due to the landing of the unified urlbar. See more details in bug 1170148.

firefox_ui_tests/functional/locationbar/test_favicon_in_autocomplete.py
```diff
@@ -0,0 +1,87 @@
...
+        # result_favicon_url = self.marionette.execute_script("""
+        #   return arguments[0].boxObject.firstChild.childNodes[0].getAttribute("src");
+        # """, script_args=[result])
+        # print 'result_favicon_url--unneeded?: %s' % result_favicon_url
+        #
+        # this variable should store the same value the mozmill test compares to
+        # locationbar_favicon_url
+        result_icon = self.marionette.execute_script("""
+          return arguments[0].image;
+        """, script_args=[result])
```
whimboo: Here we should also keep the src attribute.

firefox_ui_tests/functional/locationbar/test_favicon_in_autocomplete.py
galgeek: hi, @whimboo!

favicon.get_attribute('src') also produces an empty string here. I wonder if the underlying code for this has changed, too, with the landing of the unified urlbar...

whimboo: Hm, does the code work with Aurora? I don't think that the unified auto-complete is responsible for it. This may be something else.

galgeek:
It looks like the mozmill version of this test is simply comparing two blank strings.

A marionette version that actually tests something worth testing, with the unified urlbar, likely requires a rethink....

I've left somewhat more detailed comments in the code.

firefox_ui_tests/functional/locationbar/test_favicon_in_autocomplete.py
```diff
@@ -0,0 +1,82 @@
...
+
+        locationbar = self.browser.navbar.locationbar
+        favicon = self.browser.navbar.locationbar.favicon
+
+        # this code assigns locationbar_favicon_url an empty string.
+        # a sleep doesn't help, and there's no change if run again later in test.
+        # is this a prefs difference from the mozmill tests?
+        # mozmill test's expect.contain, line 73, succeeds with comparison in either order...
+        # aha! mozmill --console-level=DEBUG shows line 73 comparing empty strings!?
+        locationbar_favicon_url = favicon.get_attribute('src')
```
whimboo: I just checked that with the DOM inspector and noticed that this is true. It looks like that the favicon icon is now used as `list-style-image`. So we would have to retrieve this CSS property from the element.

firefox_ui_tests/functional/locationbar/test_favicon_in_autocomplete.py
```diff
@@ -0,0 +1,82 @@
...
+
+        # this code assigns favicon_image a value different from the compared result_icon
+        favicon_image = favicon.value_of_css_property('list-style-image')
```
whimboo: I don't think that we ever show the real underlying favicon anymore. It seems to always be the lock (secure) or globe (insecure) icon now.

firefox_ui_tests/functional/locationbar/test_favicon_in_autocomplete.py
```diff
@@ -0,0 +1,82 @@
...
+
+        # this variable, equivalent to mozmill's listFaviconUrl, never gets used again.
+        # here, it's identical to result_icon.
+        # result_favicon_url = self.marionette.execute_script("""
+        #   return arguments[0].boxObject.firstChild.childNodes[0].getAttribute("src");
+        # """, script_args=[result])
+        # print 'result_favicon_url--unneeded?: %s' % result_favicon_url
+
+        # this variable should store the same value the mozmill test compares to
+        # locationbar_favicon_url... but see above re: line 73
```
whimboo: No. See above why this is not the case anymore.

galgeek:
Thanks, Henrik!

It looks like a page's favicon now gets displayed on the page's tab, not in the location bar, and may be stored in a xul:image object with anonid tab-icon-image. 

A page's favicon still shows in its item within the autocomplete results.

So this may be an approach to updating this test. I may have a chance look harder again soon, or someone else is welcome to take it.

firefox_puppeteer/tests/test_software_update.py
```diff
@@ -9,6 +11,7 @@
 from firefox_puppeteer.api.software_update import SoftwareUpdate
 
 
+@unittest.skip('Bug 1196006 - about and page_info window tests time out on Travis Linux worker')
```
whimboo: Please remove all changes to this file. Skipping should happen in the other bug. But if this alone doesn't help lets skip all the hanging tests.

firefox_ui_tests/functional/locationbar/test_favicon_in_autocomplete.py
```diff
@@ -0,0 +1,58 @@
...
+        # Clear the location bar, type the test string, check that autocomplete list opens
+        locationbar.clear()
+        locationbar.urlbar.send_keys(self.test_string)
+        self.assertEqual(locationbar.value, self.test_string)
+        Wait(self.marionette).until(lambda _: self.autocomplete_results.is_complete)
+
+        result = self.autocomplete_results.visible_results[1]
+
+        result_icon = self.marionette.execute_script("""
+          return arguments[0].image;
```
whimboo: Maybe get_attribute() works here? Or is that a JS property?

firefox_ui_tests/functional/locationbar/test_favicon_in_autocomplete.py
```diff
@@ -0,0 +1,58 @@
...
+        """, script_args=[result])
+
+        self.assertIn(self.test_favicon, result_icon)
```
whimboo: Why `assertIn()` and not `assertEqual`?

galgeek: result_icon contains the favicon's absolute URL (e.g., http://[hostname]/images/mozilla_favicon.ico); we could use assertEqual if we parse out the favicon's file name.

firefox_ui_tests/functional/locationbar/test_favicon_in_autocomplete.py
galgeek: get_attribute() doesn't work on result.
Trying to index result, or autocomplete.visible_result[1], in Python gives...
`TypeError: 'HTMLElement' object does not support indexing`
Maybe this is the library work you wanted done with this test? visible_results should maybe return something Python-friendlier.

galgeek:
The Travis failure is one of the intermittent failures of test_safe_browsing_warning_pages.py, not this test.

firefox_ui_tests/functional/locationbar/test_favicon_in_autocomplete.py
whimboo: Then we will have to wait for the `get_property()` method which will be added to Marionette hopefully soonish. Thanks for checking that.

firefox_ui_tests/functional/locationbar/test_favicon_in_autocomplete.py
```diff
@@ -0,0 +1,58 @@
...
+from firefox_ui_harness import FirefoxTestCase
+
+
+class TestFaviconInAutocomplete(FirefoxTestCase):
+
+    def setUp(self):
+        FirefoxTestCase.setUp(self)
+
+        # Location bar suggests "History"
+        self.prefs.set_pref('browser.urlbar.default.behavior', 1)
```
whimboo: This pref is not used anymore. You should make use of one of the newer ones similar to https://github.com/mozilla/firefox-ui-tests/blob/master/firefox_ui_tests/functional/locationbar/test_suggest_bookmarks.py.

firefox_ui_tests/functional/locationbar/test_favicon_in_autocomplete.py
```diff
@@ -0,0 +1,58 @@
...
+
+        self.places.remove_all_history()
+
+        self.test_urls = [self.marionette.absolute_url('layout/mozilla.html')]
+
+        self.test_string = 'mozilla'
+        self.test_favicon = 'images/mozilla_favicon.ico'
```
whimboo: nit: I would leave out the folder. Just take the filename.

whimboo:
Can you please merge in (not rebase) latest master, so we get passing tests? Right now it fails due to the travis changes. Thanks.

galgeek:
Thanks for your feedback, Henrik!  I've made the changes you requested.

I apologize, after amending the commit message, I am not leaving you a tidy review package, but for what it's worth, here's the git diff I did before committing today's updates: https://pastebin.mozilla.org/8843992

PR #184, opened 2015-05-28
Bug 1169374 - Use non-strict version dependencies for mozbase modules. r=armenzg

whimboo:


armenzg:
All those dependencies show up on the internal pypi. Go for it when ready. r=armenzg

whimboo:
Thanks Armen. Travis on OS X is green, so I'm going to merge it now and backport soon to all other branches.

PR #186, opened 2015-05-29
Bug 1169823 - update safe browsing notification test to support new unwanted software feature

mwobensmith:


firefox_ui_tests/remote/security/test_safe_browsing_notification.py
```diff
@@ -15,6 +15,12 @@ def setUp(self):
         FirefoxTestCase.setUp(self)
 
         self.test_data = [
+            # Unwanted software URL
+            {
+                'button_property': None,
+                'report_page': None,
```
whimboo: Can you please add as a comment that there is currently no way to report it. That's why we do not need those two properties.

firefox_ui_tests/remote/security/test_safe_browsing_notification.py
```diff
@@ -56,11 +62,16 @@ def test_notification_bar(self):
...
+                # Only do this if feature supports it
+                if label is not None:
+                    self.marionette.navigate(unsafe_page)
+                    # Wait for the DOM to receive events for about:blocked
+                    time.sleep(1)
+                    self.check_ignore_warning_button(unsafe_page)
+                    self.check_not_badware_button(label, report_page)
+                else:
+                    # Must reset permissions if this test did not run previously
+                    self.utils.remove_perms('www.itisatrap.org', 'safe-browsing')
```
whimboo: Removing the permissions should only be necessary when the ignore warning link gets clicked. And therefore we already have the removal of permissions in `check_ignore_warning_button`. So I don't see why this is necessary here.

firefox_ui_tests/remote/security/test_safe_browsing_notification.py
```diff
@@ -56,11 +62,16 @@ def test_notification_bar(self):
...
-                # Wait for the DOM to receive events for about:blocked
-                time.sleep(1)
-                self.check_ignore_warning_button(unsafe_page)
-                self.check_not_badware_button(label, report_page)
+                # Only do this if feature supports it
+                if label is not None:
+                    self.marionette.navigate(unsafe_page)
+                    # Wait for the DOM to receive events for about:blocked
+                    time.sleep(1)
+                    self.check_ignore_warning_button(unsafe_page)
```
whimboo: The ignore warning link is still present for the unwanted software case. So you should not exclude that from testing. It's really only the not badware link which you cannot test for that page type.

mwobensmith: The "ignore warning" link gets tested for unwanted software on line 80, so it does not go untested. It cannot be tested here on line 70 because this part is used to test the report functionality, which unwanted software blocking does not have. So, I believe my logic is correct here.

firefox_ui_tests/remote/security/test_safe_browsing_notification.py
mwobensmith: Done.

firefox_ui_tests/remote/security/test_safe_browsing_notification.py
whimboo: Ok, you are right. The `label` variable is kinda misleading here. I had to triple check what this is. Would you mind updating the 3 instances of it to `button_property`? That is what it actually is. The label itself will be retrieved later via `get_property()`.

mwobensmith: Done.

whimboo:
PR got merged as https://github.com/mozilla/firefox-ui-tests/commit/5e50fbf3e2bd2e6a3728dee5a1cdb43b77ef04ee.

PR #192, opened 2015-06-04
Bug 1171400 - Disable test_bookmarks in test_places.py for XVFB a=bustage

whimboo:


whimboo:
For bug 1171400 we landed simply the screen resolution change.

PR #198, opened 2015-06-14
Bug 1170148 - Fix test_toolbars.py for unified auto-complete feature

galgeek:
This patch adds the property is_complete to the autocomplete results class and replaces ...is_open with ...is_complete in problem tests in test_toolbars.py.

Perhaps it would be better to replace the code in the ...is_open property with code similar to that in ...is_complete, and skip adding a separate property?

https://bugzilla.mozilla.org/show_bug.cgi?id=1170148

### DIFF
--- a/firefox_puppeteer/tests/test_toolbars.py
+++ b/firefox_puppeteer/tests/test_toolbars.py
@@ -137,9 +137,6 @@ class TestAutoCompleteResults(FirefoxTestCase):
             all_matches = title_matches + url_matches
             self.assertTrue(len(all_matches) > 0)
             for match_fragment in all_matches:
-                # skip the new unified search item, if present
-                if match_fragment == '- Search with Yahoo':
-                    continue
                 self.assertIn(input_text, match_fragment.lower())
 
 
--- a/firefox_puppeteer/tests/test_toolbars.py
+++ b/firefox_puppeteer/tests/test_toolbars.py
@@ -72,7 +72,6 @@ class TestLocationBar(FirefoxTestCase):
         self.assertEqual('toolbarbutton', stop_button.get_attribute('localName'))
 
 
-@unittest.skip('Bug 1170148 - Fix test_toolbars.py for unified auto-complete feature')
 class TestAutoCompleteResults(FirefoxTestCase):
     def setUp(self):
         FirefoxTestCase.setUp(self)
@@ -97,7 +96,7 @@ class TestAutoCompleteResults(FirefoxTestCase):
         self.assertFalse(autocompleteresults.is_open)
         self.browser.navbar.locationbar.urlbar.send_keys('a')
         results = autocompleteresults.results
-        self.wait_for_condition(lambda _: autocompleteresults.is_open)
+        self.wait_for_condition(lambda _: autocompleteresults.is_complete)
         visible_result_count = len(autocompleteresults.visible_results)
         self.assertTrue(visible_result_count > 0)
         self.assertEqual(visible_result_count,
@@ -107,7 +106,7 @@ class TestAutoCompleteResults(FirefoxTestCase):
     def test_close(self):
         autocompleteresults = self.browser.navbar.locationbar.autocomplete_results
         self.browser.navbar.locationbar.urlbar.send_keys('a')
-        self.wait_for_condition(lambda _: autocompleteresults.is_open)
+        self.wait_for_condition(lambda _: autocompleteresults.is_complete)
         # The Wait in the library implementation will fail this if this doesn't
         # end up closing.
         autocompleteresults.close()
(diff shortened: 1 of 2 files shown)


galgeek:
The Travis failure results from failures in other locationbar tests, likely related to the new unified autocomplete.

firefox_puppeteer/tests/test_toolbars.py
```diff
@@ -138,7 +137,10 @@ def test_matching_text(self):
             all_matches = title_matches + url_matches
             self.assertTrue(len(all_matches) > 0)
             for match_fragment in all_matches:
-                self.assertIn(match_fragment, (input_text, input_text.upper()))
+                # skip the new unified search item, if present
+                if match_fragment == '- Search with Yahoo':
+                    continue
```
whimboo: We cannot hard-code any search engine given that this will break localized builds of Firefox. Instead we should check for the special CSS class assigned to the result.

firefox_puppeteer/ui/toolbars.py
```diff
@@ -315,6 +315,22 @@ def is_open(self):
...
 
     @property
+    def is_complete(self):
+        """Returns when this popup is open and search results are complete.
+
+        :returns: True when the popup is open including any search results, otherwise false.
+        """
+        return self.marionette.execute_script("""
+          Components.utils.import("resource://gre/modules/Services.jsm");
+          let win = Services.focus.activeWindow;
```
whimboo: You should pass-in the browser window into the `execute_script` call. Otherwise we have an inconsistency between the Marionette known active window and another focused one.

whimboo:
Thanks for taking care of this bug, and it'ss great to see that had time to look at this. The code looks fine but needs some more fixes.

The other failure is clearly a new regression since yesterday. I will file a new bug for it to get it temporarily solved, so it doesn't call failures on all platforms.

galgeek:
Thanks for your feedback, Henrik!

Maybe this is the best we can do for now with the toolbars.py update?

For test_toolbars.py and its test_matching_text, the current get_matching_text method doesn't support the new search item... the new search item looks like it shows up last in the results, so I skip a check on the last item for now, and include a comment.

firefox_puppeteer/tests/test_toolbars.py
```diff
@@ -137,8 +136,9 @@ def test_matching_text(self):
             url_matches = autocompleteresults.get_matching_text(result, "url")
             all_matches = title_matches + url_matches
             self.assertTrue(len(all_matches) > 0)
-            for match_fragment in all_matches:
-                self.assertIn(match_fragment, (input_text, input_text.upper()))
+            # check all but the last item (the new unified search item from Bug 1168811)
+            for match_fragment in all_matches[:-1]:
+                self.assertIn(input_text, match_fragment.lower())
```
whimboo: We do not know how many search items will be visible at the end. So my proposal here would be to limit the auto-completion to history items only. Similar to what we do in the actual tests. That's a safer method.

firefox_puppeteer/ui/toolbars.py
```diff
@@ -315,6 +315,22 @@ def is_open(self):
         return self.popup.get_attribute('state') == 'open'
 
     @property
+    def is_complete(self):
+        """Returns when this popup is open and search results are complete.
```
whimboo: I would not mention `search results` but better `autocomplete results`. Search will confuse here because of the shown search engine entries at the end.

firefox_puppeteer/ui/toolbars.py
```diff
@@ -315,6 +315,22 @@ def is_open(self):
...
+
+        :returns: True when the popup is open including any search results, otherwise false.
```
whimboo: Simply: `True, when auto-complete results have been populated`.

galgeek:
Thanks again for your feedback, Henrik!

I poked around more... realized the get_matching_text method was not the problem, and found a better way to check only particular autocomplete results.

firefox_puppeteer/tests/test_toolbars.py
```diff
@@ -129,16 +128,19 @@ def test_matching_text(self):
...
+            # check matching text only for results of type bookmark
+            if result.get_attribute('type') != 'bookmark':
+                continue
             title_matches = autocompleteresults.get_matching_text(result, "title")
             url_matches = autocompleteresults.get_matching_text(result, "url")
             all_matches = title_matches + url_matches
             self.assertTrue(len(all_matches) > 0)
             for match_fragment in all_matches:
-                self.assertIn(match_fragment, (input_text, input_text.upper()))
+                self.assertIn(input_text, match_fragment.lower())
```
whimboo: Looks fine to me. Just a nit here, the value we want to compare it with is `input_text`. So it still has to be the second parameter. I will do this flip before I merge the commit.

whimboo:
Merged as https://github.com/mozilla/firefox-ui-tests/commit/0310d279ecb2944130a6d3e18071330ccd589da2

PR #205, opened 2015-06-24
Update marionette requirements to get browsermob-proxy feature.

sydvicious:
Need to be able to use the browser mob feature, which requires newer versions of marionette-driver and marionette-client


whimboo:
@sydvicious, our policy is to file bugs in Bugzilla to get changes included. A pull request is not enough. Please file an appropriate bug and attach a link-file to this PR as attachment to it. Ask for review that way. Thanks.

setup.py
```diff
@@ -7,8 +7,8 @@
 PACKAGE_VERSION = '0.2'
 
 deps = [
-    'marionette-client >= 0.12',
-    'marionette-driver >= 0.7',
+    'marionette-client >= 0.15',
+    'marionette-driver >= 0.8',
```
whimboo: Versions have been updated. Please use the latest driver 0.9, what I also wanted to update in a bit.

whimboo:
I need some features which are available in newer releases. So I created [bug 1178792](https://bugzilla.mozilla.org/show_bug.cgi?id=1178792)

PR #210, opened 2015-06-30
Bug 1180429 - Missing headers for classes in TOC of docs

utvar:
Hi, @whimboo, this changes files in the docs/ui directory so that all classes are listed in the TOC (under the library containing them). If you like this, I could do the same for the docs/api directory.

### DIFF
--- a/firefox_puppeteer/docs/api/appinfo.rst
+++ b/firefox_puppeteer/docs/api/appinfo.rst
@@ -8,8 +8,8 @@ Firefox and provides access to a subset of its members.
 
 .. _nsIXULAppInfo: https://developer.mozilla.org/docs/Mozilla/Tech/XPCOM/Reference/Interface/nsIXULAppInfo
 
-API reference
--------------
+AppInfo
+-------
 
 .. autoclass:: AppInfo
    :members:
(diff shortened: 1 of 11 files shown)


whimboo:
I think that's a good idea. So yes, if we also miss those headings for api modules, lets also get them added there. Btw. does a bug in bugzilla exist for those changes? We usually track review process over there.

utvar:
Thanks for the feedback, @whimboo. There doesn't seem to be a bug report on Bugzilla for this. Should I enter one there under Mozilla QA? Then, would you prefer that I reference the bug in a comment, or in the title of this PR?

utvar:
It looks like the Travis build failed because of test failures, not because of my code.

chmanchester:
@utvar, this looks good. Please squash your commits into a single commit and update the commit message to reflect the bug number and reviewer so I can merge. Thanks!

utvar:
@chmanchester, here is the updated commit. Do you want me to change the PR title to match that of the Bug and commit message?

whimboo:
@utvar thanks for your contribution! If you are interested in something else please let us know.

PR #215, opened 2015-07-08
Bug 1181788 - Use uri objects instead of strings in utils.remove_perm…

chmanchester:
… for api change in nsIPermissionManager


whimboo:
As the bug status says this PR got merged a while ago.

PR #216, opened 2015-08-05
Bug 1170576 - Upgrade mozdownload to version 1.17. r=automatedtester

whimboo:


jmaher:
this is a fairly simple change, I assume the checks which have failed are intermittent.  Please confirm that before merging.

whimboo:
The failures happened here are due to missing packages from our internal pypi. I filed [bug 1191331](https://bugzilla.mozilla.org/show_bug.cgi?id=1191331) to get them updated. Otherwise we see test failures if using the latest packages. So none of those are related.

PR #217, opened 2015-08-05
Bug 1146296 - Add IdentityPopup.open() as helper method

gbrmachado:
I made a mistake with the git command line that closed the earlier pull request. Sorry.


firefox_puppeteer/tests/test_toolbars.py
```diff
@@ -223,8 +223,7 @@ def tearDown(self):
 
     @skip_under_xvfb
     def test_elements(self):
-        self.locationbar.identity_box.click()
-        Wait(self.marionette).until(lambda _: self.identity_popup.is_open)
+        self.open_identity_popup()
```
whimboo: Please keep in mind that `self` in this context is the test class but not the instance of the locationbar. What you have to use here is `self.locationbar.open_identity_popup()`. Please ensure to really run all tests locally and check travis results before asking for review/feedback. If there are questions please let me know via a need-info? request in Bugzilla.

firefox_puppeteer/ui/toolbars.py
```diff
@@ -223,6 +223,10 @@ def notification_popup(self):
         """
         return self.marionette.find_element(By.ID, "notification-popup")
 
+    def open_identity_popup(self):
+        self.locationbar.identity_box.click()
```
whimboo: `self.locationbar` does not exist in the context of the locationbar class. It is simply `self`.

whimboo:
So this all looks fine. I did a quick check locally and as it looks like there is an instance missing to be replaced in test_mixed_script_content_blocking.py. Once you included this we can get the patch landed.

gbrmachado:
Hey @whimboo . I've added the new method into the test_mixed_script_content_blocking.py file :)
I think everything is fine now :)

whimboo:
Thanks for the update. I will have a look in a minute. Just one thing which would be nice for future PRs. If you push an update please try to avoid squashing them all. We only want to do that at the very end. Otherwise reviewing updates will be much harder due to I have to go through the full list of changes again. Thanks.

whimboo:
Merged PR as https://github.com/mozilla/firefox-ui-tests/commit/98cfdc40e7e32353da0aeac2126e73b803f04e64.

PR #225, opened 2015-08-20
Bug 1196006 - skip window tests that are timing out or failing

galgeek:
test_about_window.py has been timing out on Travis Linux worker, and when skipped, test_page_info_window, too, times out...

### DIFF
--- a/firefox_puppeteer/tests/test_about_window.py
+++ b/firefox_puppeteer/tests/test_about_window.py
@@ -4,9 +4,11 @@
 
 from marionette_driver import By
 
+from firefox_ui_harness.decorators import skip_under_xvfb
 from firefox_ui_harness import FirefoxTestCase
 
 
+@skip_under_xvfb
 class TestAboutWindow(FirefoxTestCase):
 
     def setUp(self):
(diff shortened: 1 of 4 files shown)


whimboo:
I think it should be enough to skip those tests for XVFB only. Taking the current skip logic will completely skip them even they pass in non-headless mode.

firefox_puppeteer/tests/test_about_window.py
```diff
@@ -4,9 +4,11 @@
 
 from marionette_driver import By
 
+from firefox_ui_harness.decorators import skip_under_xvfb
 from firefox_ui_harness import FirefoxTestCase
```
whimboo: For the future I would prefer to add sub-package inclusions below the upper package. But that's fine for now given that it is temporary anyway.

firefox_ui_tests/functional/private_browsing/test_about_private_browsing.py
```diff
@@ -21,6 +21,7 @@ def setUp(self):
 
         self.pb_url = support_url + 'private-browsing'
 
+    @skip_under_xvfb
     @skip_if_e10s
```
whimboo: same for the ordering here. but its ok.

whimboo:
I will land the patch after squashing it and updating the reviewer name to 'whimboo' given that it should be the IRC handle.

Merged as https://github.com/mozilla/firefox-ui-tests/commit/b196f5bdd3445e8599089fee5a00bcfd7391ada0.

PR #233, opened 2015-09-03
Bug 1201668 - Allow running Firefox UI update tests through a python wrapper script

armenzg:
In bug 1198448 [1] we found that calling the generated 'firefox-ui-update' binary on the Release Engineering Windows test machines triggers a UAC prompt.

This pull request adds two python scripts which act as the two generated binaries.

[1] https://bugzilla.mozilla.org/show_bug.cgi?id=1198448


whimboo:
Please create a bug in Bugzilla under Mozilla QA / Firefox UI Tests, so we can track the necessary work and review process.

tests_script.py
```diff
@@ -0,0 +1,4 @@
+from firefox_ui_harness import cli
+
+if __name__ == '__main__':
+    cli()
```
whimboo: I'm not that happy with those scripts especially in the root folder. So I would suggest to place them under `firefox_ui_harness/update.py` and respectively. Then we can move out the `cli` code from runtests.py, and update setup.py to make use of the new cli wrapper scripts.

armenzg:
@whimboo is this better?

(venv1)armenzg-thinkpad firefox-ui-tests git:[origin/python_script*] $ firefox-ui-update --help | wc -l
151
(venv1)armenzg-thinkpad firefox-ui-tests git:[origin/python_script*] $ python firefox_ui_harness/update.py --help | wc -l
151

firefox_ui_harness/update.py
```diff
@@ -0,0 +1,4 @@
+from firefox_ui_harness import cli_update
+
+if __name__ == '__main__':
+    cli_update()
```
whimboo: I was thinking that we move the whole method from runtests.py to this file: https://github.com/mozilla/firefox-ui-tests/blob/master/firefox_ui_harness/runtests.py#L81. Maybe its better to also rename this file as `cli_update.py`.

firefox_ui_harness/__init__.py
```diff
@@ -2,9 +2,9 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-from runtests import (cli,
-                      cli_update,
-                      )
+from cli_update import cli_update
+
```
whimboo: nit: Kill this extra line and all will be fine.

armenzg:
Thank you!

PR #235, opened 2015-09-09
Bug 1203267 - Add a requirements.txt file for pinned mozbase packages

whimboo:


whimboo:
The reported failure from Travis is only for an unrelated test. So nothing to worry about here.

AutomatedTester:
r+

PR #240, opened 2015-09-21
Bug 1196006 - Fix skips for tests that are timing out

galgeek:
Skips should generally be put on tests, not on classes. 

(Adding a skip to a class disables all its tests, for ALL test runs.)


galgeek:
We want to revert the earlier PR first...

PR #253, opened 2015-09-29
Bug 1196006 - Focus window created in open_window()

galgeek:


firefox_puppeteer/ui/windows.py
```diff
@@ -349,7 +349,8 @@ def window_opened(mn):
 
         window = self._windows.create_window_instance(new_handle, expected_window_class)
 
-        Wait(self.marionette).until(lambda _: window.focused == expect_focus)
+        if expect_focus:
+            window.focus()
```
whimboo, line 353: I just noticed the discrepancy of the parameter name and our action. Right now we indeed only check for an expected state. So that is correct. But it looks like (as you found out) that the new window is not getting correctly focused. It might be a Firefox or a Marionette issue. 

What if we rename the parameter to `focus` and still let it default to `True`. All other modules would have to be checked to ensure we don't break anything. But then it's clear what we want here.

whimboo:
That looks great. Please squash all commits so I  can land it later today. Thanks Barbara!

whimboo:
PR got merged with https://github.com/mozilla/firefox-ui-tests/compare/c0d78e27baa1...57cbeac9df26.

PR #259, opened 2015-10-03
Bug 1136902 - Added test file and method for utils sanitize method

sr-murthy:
Bug 1136902 - Added test file and method for utils sanitize method.


firefox_puppeteer/tests/test_utils.py
```diff
@@ -0,0 +1,69 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from firefox_ui_harness import FirefoxTestCase
+
+
+class TestUtils(FirefoxTestCase):
```
whimboo, line 8: I would add a different class for each helper method. Otherwise it will become too big once more helper methods are contained in that file.

firefox_puppeteer/tests/test_utils.py
```diff
@@ -0,0 +1,69 @@
...
+          root.containerOpen = true;
+          for (let i = 0; i < root.childCount; i++) {
+            urls.push(root.getChild(i).uri)
+          }
+          root.containerOpen = false;
+
+          return urls;
+        """)
+
+    def test_sanitize(self):
```
whimboo: It would be better to have different test methods for each individual data type.

firefox_puppeteer/tests/test_utils.py
```diff
@@ -0,0 +1,69 @@
...
+
+        # Call the utils sanitize().
+        self.utils.sanitize(data_type={"history": True, "cookies": True})
```
whimboo: We should ensure that the cookie is set. Same for the history which should contain x items.

firefox_puppeteer/tests/test_utils.py
```diff
@@ -0,0 +1,69 @@
...
+
+        # Check that the browser history has been cleared
+        self.assertEqual(len(self.get_all_urls_in_history()), 0)
```
whimboo, line 65: It would be better to to leave out the length and compare the arrays directly. That way we could see which history items are still present.

firefox_ui_tests/resources/cookie_single.html
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE html>
```
whimboo, line 1: Please put this file into the `cookies` sub folder so we keep those files separate by module.

firefox_puppeteer/tests/test_utils.py
sr-murthy, line 8: I'm not sure I understand.  Do you mean `get_all_urls_in_history`?  That is the only helper method I see, but you want me to put that in a class of its own?  Where should the class live, or is there an example of something similar?

firefox_puppeteer/tests/test_utils.py
sr-murthy: You mean a different sanitize method for different types of resource, like pages, cookies, etc.

firefox_puppeteer/tests/test_utils.py
whimboo, line 8: Gna, partly scratch my last comment. It should not have been `helper method` but methods in utils.js. Means each of those methods in utils.js should better get its own class due to complexity. Alone for sanitize we need a couple of test methods in the class.

firefox_puppeteer/tests/test_utils.py
whimboo: Exactly. Otherwise it will be very hard to follow once we have tests for all types.

firefox_puppeteer/tests/test_utils.py
sr-murthy, line 8: Sure, no problem.

PR #265, opened 2015-10-21
Bug 1217046 - Move FirefoxTestCase from firefox_ui_harness into firefox_puppeteer

ujjwalwahi:


firefox_puppeteer/testcase.py
```diff
@@ -4,8 +4,8 @@
```
whimboo: Great. But I would really suggest to keep the subfolder and put this file into firefox_puppeteer/testcases/. Maybe we can keep using `base.py`?

firefox_puppeteer/testcase.py
```diff
@@ -4,8 +4,8 @@
 
 from marionette import MarionetteTestCase
 
-from firefox_puppeteer import Puppeteer
-from firefox_puppeteer.ui.browser.window import BrowserWindow
+from . import Puppeteer
```
whimboo: Please keep the absolute references for imports.

firefox_ui_harness/runners/base.py
```diff
@@ -7,8 +7,8 @@
 import mozinfo
 from marionette import BaseMarionetteTestRunner
 
-from firefox_ui_harness.testcases import FirefoxTestCase
 import firefox_ui_tests
+from firefox_puppeteer.testcases import FirefoxTestCase
```
whimboo, line 11: In a follow-up bug we will have to get rid of this import given that the harness should not reference puppeteer. For now this looks fine given that we at least have extracted the testcases from the harness.

firefox_ui_tests/testcases/update.py
```diff
@@ -4,9 +4,10 @@
 
 import pprint
 
+from firefox_puppeteer.testcases import FirefoxTestCase
```
whimboo: Please sort this line alphabetically into the next block.

whimboo:
Just a nit to fix. If you want you can already squash all the commits into a single one and give it a proper comment (see git log for examples), so it's easier for landing later today. Thanks!

whimboo:
@ujjwalwahi if you do changes to your pull I would recommend that you add a comment to the appropriate bug in Bugzilla or at least here on the PR. Sadly Github doesn't inform anyone if new commits have been pushed. So your latest commits went unnoticed. Sorry.

ujjwalwahi:
@whimboo actually I don't know why but one test is failing during Travis CI build, I was trying  to fix that, that's why I didn't added a comment. Yet, I am not unable to find a solution. It will be great if you can suggest  something. Code seems okay to me.

whimboo:
I see. So the failure here is only happening for a single Travis build.  All three others are passing. By checking the failure I see that this is [bug 1212811](https://bugzilla.mozilla.org/show_bug.cgi?id=1212811) and not related to this PR at all. So all fine here. I will have a look at the PR tomorrow morning.

whimboo:
Looks all great to me now! Even the commit message is perfect so that the PR can directly be merged. I will do it right away.

whimboo:
Merged the PR as https://github.com/mozilla/firefox-ui-tests/commit/50527c612f3c063ab52046d0d01f610b64f5f003

PR #267, opened 2015-10-23
Bug 1121710 - Fix MenuBar class for correct handling of menus r=whimboo

tareqak:
From here: https://bugzilla.mozilla.org/show_bug.cgi?id=1121710


tareqak:
@whimboo, Please have a look.

firefox_puppeteer/tests/test_menubar.py
```diff
@@ -20,7 +17,8 @@ def test_click_item_in_menubar(self):
 
         def opener(_):
             # Hard-coded labels will not work in localized builds
```
whimboo: nit: lets remove that line. It was only a temporary placeholder to visualize that this won't work.

firefox_puppeteer/ui/menu.py
```diff
@@ -37,6 +37,22 @@ def get_menu(self, label):
...
 
+    def get_menu_by_id(self, menu_id):
+        """Get a :class:`MenuElement` instance corresponding to the specified
+        ID.
+
+        :param menu_id: The ID of the menu, e.g., **file-menu** or
+        **view-menu**.
+        :returns: A :class:`MenuElement` instance.
+        """
+        menu = [m for m in self.menus if m.get_attribute('id') == menu_id]
```
whimboo, line 48: So as of now the current menu.py can only handle the mainmenu. It's because we rely on self.menus here. We have to make it more flexible similar to the link to Mozmill which I gave lately. We have to request the top element of any menu as best outside of the Menu class and pass it into the constructor. Lets base the Menu class also on UIBaseLib so that we have a reference back to the window.

Menus we should test are `main-menubar` and `contentAreaContextMenu`.

firefox_puppeteer/ui/menu.py
```diff
@@ -45,6 +61,15 @@ def select(self, label, item):
...
         return self.get_menu(label).select(item)
 
+    def select_by_id(self, menu_id, item_id):
+        """Select an item in a menu.
+
+        :param menu_id: The ID of the menu, e.g., **file-menu** or **view-menu**.
+        :param item_id: The ID of the item in the menu, e.g.,
+        **menu_newNavigatorTab**.
+        """
+        return self.get_menu_by_id(menu_id).select_by_id(item_id)
```
whimboo, line 71: Keep in mind that menu items of some menus in the main menubar are getting populated when the menu is opened. If we do not open them, you might miss entries, e.g. check that by clicking the file menu, or opening it via Alt+F. Different items are shown. This also applies to dynamic lists like the bookmarks menu.

firefox_puppeteer/ui/menu.py
```diff
@@ -45,6 +61,15 @@ def select(self, label, item):
...
+
     class MenuElement(DOMElement):
         """Wraps a menu element DOM element."""
```
whimboo, line 74: Best would be to have that as in independent class outside of the menubar class similar to other UIBaseLib based classes. For example see the deck class and its other classes in https://github.com/mozilla/firefox-ui-tests/blob/mozilla-central/firefox_puppeteer/ui/about_window/deck.py.

firefox_puppeteer/ui/menu.py
```diff
@@ -70,3 +95,19 @@ def select(self, label):
...
             return item[0].click()
+
+        def select_by_id(self, menu_item_id):
+            """Click on a menu item within this menu.
+
+            :param menu_item_id: The ID of the menu item, e.g.,
+            **menu_newNavigatorTab**.
+            """
+            item = [l for l in self.items if l.get_attribute('id') ==
+                    menu_item_id]
```
whimboo, line 106: I don't think it will handle multi-level menus like `View -> Toolbars -> Menu Bar`. We would need a recursive strategy to go through each menu hierarchy.

whimboo:
Thanks for the update.  So I checked for completeness and found the following line remaining:

https://github.com/mozilla/firefox-ui-tests/blob/mozilla-central/firefox_puppeteer/tests/test_page_info_window.py#L64

Once this is updated, feel free to squash all commits into a single one by also rebasing against latest changes on mozilla-central first.

tareqak:
This PR is ready to be merged. The test that failed succeeded in the next run in the same set of tests.

whimboo:
This PR got merged to mozilla-central as:
https://github.com/mozilla/firefox-ui-tests/commit/b1ea646a38f8373364aceac895083990bdff7e80

PR #277, opened 2015-11-02
Bug 1121710 - Separate MenuElement class r=whimboo

tareqak:
@whimboo Please have a look.


whimboo:
@tareqak mind creating a review request on Bugzilla so I can act on? Thanks.

whimboo:
The changes here look fine to me but as discussed earlier I would like to also see that we make both classes more general so it also works for context menus.

I simply don't want to land too small changes in case we have to backport patches on that bug to older branches. It could become a lot of work due to micro-changesets landing on mozilla-central.

tareqak:
Ok, I'll add the Menu refactoring work to this PR and make a review request on Bugzilla once it is ready.

firefox_puppeteer/ui/browser/window.py
```diff
@@ -70,6 +72,20 @@ def is_private(self):
...
+    def menubar(self):
+        """Provides access to the menu bar.
+
+        See the :class:`~ui.menu.MenuBar` reference.
+        """
+        self.switch_to()
+
+        if not self._menubar:
+            menubar = self.window_element.find_element(By.ID, 'main-menubar')
+            self._menubar = MenuBar(lambda: self.marionette, self, menubar)
```
whimboo, line 84: :+1: Kinda love that to have it finally as a property on the window object.

firefox_puppeteer/ui/menu.py
```diff
@@ -6,10 +6,10 @@
 from marionette_driver.errors import NoSuchElementException
 
 from firefox_puppeteer import DOMElement
-from firefox_puppeteer.base import BaseLib
+from firefox_puppeteer.ui_base_lib import UIBaseLib
 
 
-class MenuBar(BaseLib):
+class MenuBar(UIBaseLib):
```
whimboo, line 12: This should not only stand for menubar like elements but a menu in general. The `menubar` property of the browser window is just a specific instance of a `Menu` class.

firefox_puppeteer/ui/menu.py
```diff
@@ -21,7 +21,7 @@ def menus(self):
         """
         menus = (self.marionette.find_element(By.ID, 'main-menubar')
                                 .find_elements(By.TAG_NAME, 'menu'))
```
whimboo, line 23: You also want to get rid of the `main-menubar` line here. You can retrieve the element via `self.element` now, which makes this class really general.

whimboo, line 23: Given the new generic state of the class it would be great to have an additional unit test in the test_menu.py class which will cover a context menu like the #contentAreaContextMenu which exists for all pages in the content scope.

whimboo:
Oh, and please make sure to not amend commits or doing a rebase for the next push. I would like to see the individual changes for each commit. A squash will happen at the very end right before the merge to master. Thanks.

firefox_puppeteer/ui/menu.py
Adrianzatreanu: could you please clarify what is needed to be done here? i cannot find anything more general than a menubar as a menu

whimboo: There are menubar and context menus. Both are menu like types.

Adrianzatreanu: and where exactly can the context menus be found? cant quite find them

whimboo: I would suggest that we continue our discussion on the appropriate bug in Bugzilla given that this PR is closed. One last comment here to answer your question. Please have a look at https://dxr.mozilla.org/mozilla-central/search?q=contentareacontextmenu&redirect=false.

PR #279, opened 2015-11-10
Bug 1204093 Add a non-ascii locale to Travis CI testing matrix for mozilla-aurora

galgeek:


whimboo:
This solution will not work. More details I will list on the bug.

PR #280, opened 2015-11-10
Bug 1204093 Add a non-ascii locale *improved*. r=whimboo

galgeek:


whimboo:
This is not the PR against mozilla-aurora but mozilla-central. So it won't apply correctly.

PR #281, opened 2015-11-11
Bug 1204093 - Add a non-ascii locale to Travis for mozilla-aurora

galgeek:


whimboo:
To get this properly fixed I would propose you run the following on the mozilla-aurora branch:

> git cherry-pick -x 57d76af

You will get merge conflicts in before_script.sh which you will have to fix. Once done commit the changes on a new branch like that one. Reason is that I see differences in .travis.yml compared to master. Also cherry-picking leaves a reference of the original commit around, so we can see in a later merge if that has already been landed.

galgeek:
@whimboo Is there some additional step to take to preserve a reference to the cherry-picked commit? I don't see such a reference when I do my best to follow your instructions above.

whimboo:
No need to close this PR.  You can still work on the same local branch and force push it.

If you run cherry-pick with the -x option, it will add an additional line to the commit message which includes the revision id of the original changeset.

galgeek:
@whimboo I guess this isn't ideal, but maybe you can use it?
The second commit is the cherry-pick, leaving these two files identical to mozilla-central.
The third commit adds the only essential changes for mozilla-aurora, to before_script.sh.
I leave it to you to squash commits.

.travis/before_script.sh
```diff
@@ -1,4 +1,9 @@
 #!/bin/bash
 set -ev
 
-mozdownload --type daily --branch mozilla-aurora
+if [[ $LOCALE = 'en-US' ]]
```
whimboo, line 4: Actually we wouldn't need this if condition. mozdownload also accepts `--locale en-US`. But anyway I think that this is fine for now. Thanks.

.travis/before_script.sh
```diff
@@ -1,4 +1,9 @@
...
+then
+    mozdownload --type daily --branch mozilla-aurora
+else
+    mozdownload --type daily --branch mozilla-aurora --locale ru
```
whimboo, line 8: Also we shouldn't have hard-coded the locale here but using $LOCALE.

whimboo, line 8: Maybe we can also move the branch name as global env variable so this file would look identical across all branches.

whimboo:
Merged as https://github.com/mozilla/firefox-ui-tests/commit/6110eea4dc048fb261f500fbdbb46790148c066d

PR #282, opened 2015-11-12
Bug 1204093 Add a non-ascii locale *improved*

galgeek:
This version of the PR should be suitable for both mozilla-central and for mozilla-aurora. (For the beta and release branches, it looks like mozdownload should use --type release, rather than --type daily.)

This PR removes .travis/before_script.sh.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,9 +13,10 @@ compiler:
 
 env:
   global:
+    - MOZ_BRANCH='mozilla-central'
     - DISPLAY=$(if [[ $TRAVIS_OS_NAME = 'linux' ]]; then echo ':99'; fi)
     - MOZ_XVFB=$(if [[ $TRAVIS_OS_NAME = 'linux' ]]; then echo 1; fi)
-    - MOZ_BRANCH='mozilla-central'
+    - PIP_STRICT_VERSIONS=
   matrix:
     # Test with our internal pypi mirror / locale en-US
     - LOCALE='en-US'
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -50,7 +50,7 @@ before_script:
     - pep8 --config ./.pep8.rc --exclude=client .
 
     # Prepare the build and everything else
-    - ./.travis/before_script.sh
+    - mozdownload --type daily --branch $TRAVIS_BRANCH --locale $LOCALE
```
whimboo: As I have said on another PR for this bug, we cannot use the $TRAVIS_BRANCH environment variable. Instead we will have to define our own one.

.travis/before_script.sh
```diff
@@ -1,9 +0,0 @@
-#!/bin/bash
-set -ev
-
-if [[ $LOCALE = 'en-US' ]]
-then
-    mozdownload --type daily --branch mozilla-central
-else
-    mozdownload --type daily --branch mozilla-central --locale ru
-fi
```
whimboo, line 9: Hurray!

.travis.yml
```diff
@@ -12,25 +12,22 @@ compiler:
     - gcc
 
 env:
+  global:
+    - MOZ_BRANCH='mozilla-central'
+    - DISPLAY=$(if [[ $TRAVIS_OS_NAME = 'linux' ]]; then echo ':99'; fi)
+    - MOZ_XVFB=$(if [[ $TRAVIS_OS_NAME = 'linux' ]]; then echo 1; fi)
```
whimboo, line 18: Lovely! This cleans up the 3rd column on Travis a lot and makes it way better readable.

whimboo:
Merged as https://github.com/mozilla/firefox-ui-tests/commit/27544cb1945abd4dbaa15d52015f45807210cd70

PR #283, opened 2015-11-13
Bug 1224602 - Update libary and tests for identity and connection icon changes

whimboo:


whimboo:
The Travis failure here is an intermittent orange happened by loading a remote web page. So it timed out for 1 of 6 builds.

sydvicious:
I am curious as to the server/page that is failing, but am convinced that this patch has nothing to do with it. r+

PR #286, opened 2015-11-16
Bug 1223543 - Make use of new MarionetteHarness; r=whimboo

mjzffr:
This involves bumping marionette versions to
marionette-client 2.0.0 and marionette-driver 1.1.1

Tested locally:
- `firefox-ui-tests --binary $FF_NIGHTLY_PATH --workspace blargh firefox_ui_tests/functional/locationbar/test_favicon_in_autocomplete.py`
- `firefox-ui-update --installer "temp/Firefox 41.0.dmg" --gecko-log logs/gecko.log --address localhost:2828 --workspace build`
  - When the test completes, ./build contains a 'binary' dir and two temporary
    Firefox profiles.

### DIFF
--- a/firefox_ui_harness/runners/base.py
+++ b/firefox_ui_harness/runners/base.py
@@ -11,63 +11,13 @@ import firefox_ui_tests
 from firefox_puppeteer.testcases import FirefoxTestCase
 
 
-DEFAULT_PREFS = {
-    'app.update.auto': False,
-    'app.update.enabled': False,
-    'browser.dom.window.dump.enabled': True,
-    # Bug 1145668 - Has to be reverted to about:blank once Marionette
-    # can correctly handle error pages
-    'browser.newtab.url': 'about:newtab',
-    'browser.newtabpage.enabled': False,
-    'browser.reader.detectedFirstArticle': True,
-    'browser.safebrowsing.enabled': False,
-    'browser.safebrowsing.malware.enabled': False,
-    'browser.search.update': False,
-    'browser.sessionstore.resume_from_crash': False,
-    'browser.shell.checkDefaultBrowser': False,
-    'browser.startup.page': 0,
-    'browser.tabs.animate': False,
-    'browser.tabs.warnOnClose': False,
-    'browser.tabs.warnOnOpen': False,
-    'browser.uitour.enabled': False,
-    'browser.warnOnQuit': False,
-    'dom.ipc.reportProcessHangs': False,
-    'dom.report_all_js_exceptions': True,
-    'extensions.enabledScopes': 5,
-    'extensions.autoDisableScopes': 10,
-    'extensions.getAddons.cache.enabled': False,
-    'extensions.installDistroAddons': False,
-    'extensions.logging.enabled': True,
-    'extensions.showMismatchUI': False,
-    'extensions.update.enabled': False,
-    'extensions.update.notifyUser': False,
-    'focusmanager.testmode': True,
-    'geo.provider.testing': True,
-    'javascript.options.showInConsole': True,
-    'marionette.logging': False,
-    'security.notification_enable_delay': 0,
-    'signon.rememberSignons': False,
-    'startup.homepage_welcome_url': 'about:blank',
-    'toolkit.startup.max_resumed_crashes': -1,
-    'toolkit.telemetry.enabled': False,
-}
-
-
 class FirefoxUITestRunner(BaseMarionetteTestRunner):
     def __init__(self, **kwargs):
         BaseMarionetteTestRunner.__init__(self, **kwargs)
-
+        # select the appropriate GeckoInstance
+        self.app = 'fxdesktop'
         if not self.server_root:
             self.server_root = firefox_ui_tests.resources
-
-        # Bug 1146847: This needs a refactoring given that our default
-        # preferences are not coming from the command line
-        self.prefs.update(DEFAULT_PREFS)
-
-        if not kwargs.get('e10s'):
-            self.prefs.update({'browser.tabs.remote.autostart': False})
-
-        self.workspace = kwargs.get('workspace')
 
         self.test_handlers = [FirefoxTestCase]
 
@@ -78,11 +28,3 @@ class FirefoxUITestRunner(BaseMarionetteTestRunner):
             return binary[:end_index]
         else:
             return os.path.dirname(binary)
-
-    def run_tests(self, tests):
-        # Ensure Marionette is always reset before starting tests
-        # This might need support in Marionette base
-        self.marionette = None
-        self.tests = []
-
-        BaseMarionetteTestRunner.run_tests(self, tests)
--- a/firefox_ui_harness/runners/update.py
+++ b/firefox_ui_harness/runners/update.py
@@ -42,7 +42,7 @@ class UpdateTestRunner(FirefoxUITestRunner):
 
     def duplicate_application(self, application_folder):
         """Creates a copy of the specified binary."""
-        target_folder = os.path.join(self.workspace, 'binary.backup')
+        target_folder = os.path.join(self.workspace_path, 'binary.backup')
 
         self.logger.info('Creating a copy of the application at "%s".' % target_folder)
         mozfile.remove(target_folder)
--- a/firefox_ui_harness/runners/update.py
+++ b/firefox_ui_harness/runners/update.py
@@ -38,11 +38,17 @@ class UpdateTestRunner(FirefoxUITestRunner):
         self.run_direct_update = not kwargs.pop('update_fallback_only', False)
         self.run_fallback_update = not kwargs.pop('update_direct_only', False)
 
+        # Used in case no workspace is set
+        self.installer_workspace = kwargs.pop('installer_workspace',
+                                              self.workspace_path)
+
         self.test_handlers = [UpdateTestCase]
 
     def duplicate_application(self, application_folder):
         """Creates a copy of the specified binary."""
-        target_folder = os.path.join(self.workspace_path, 'binary.backup')
+
+        target_folder = os.path.join(self.workspace or self.installer_workspace,
+                                     'binary.backup')
 
         self.logger.info('Creating a copy of the application at "%s".' % target_folder)
         mozfile.remove(target_folder)
--- a/firefox_ui_harness/runners/update.py
+++ b/firefox_ui_harness/runners/update.py
@@ -47,8 +47,7 @@ class UpdateTestRunner(FirefoxUITestRunner):
     def duplicate_application(self, application_folder):
         """Creates a copy of the specified binary."""
 
-        target_folder = os.path.join(self.workspace or self.installer_workspace,
-                                     'binary.backup')
+        target_folder = os.path.join(self.installer_workspace, 'binary.backup')
 
         self.logger.info('Creating a copy of the application at "%s".' % target_folder)
         mozfile.remove(target_folder)
--- a/firefox_ui_harness/runtests.py
+++ b/firefox_ui_harness/runtests.py
@@ -7,71 +7,53 @@ import sys
 import tempfile
 
 import mozinstall
-
-from mozlog import structured
+from marionette.runtests import MarionetteHarness, cli as mn_cli
 
 from firefox_ui_harness.arguments import FirefoxUIArguments
 from firefox_ui_harness.runners import FirefoxUITestRunner
 
 
-def startTestRunner(runner_class, options, tests):
-    install_folder = None
+class FirefoxUIHarness(MarionetteHarness):
+    def __init__(self,
+                 runner_class=FirefoxUITestRunner,
+                 parser_class=FirefoxUIArguments):
+        MarionetteHarness.__init__(self, runner_class, parser_class)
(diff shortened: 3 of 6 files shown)


firefox_ui_harness/runners/base.py
```diff
@@ -11,64 +11,14 @@
...
-    'toolkit.telemetry.enabled': False,
-}
-
-
 class FirefoxUITestRunner(BaseMarionetteTestRunner):
     def __init__(self, **kwargs):
         BaseMarionetteTestRunner.__init__(self, **kwargs)
-
+        # pick up prefs from marionette_driver.geckoinstance.DesktopInstance
+        self.app = 'fxdesktop'
```
whimboo, line 18: What do you mean with `prefs` here? If that should be related to the moved preferences I don't think we have to mention that here. Better we say that we setup the right GeckoInstance by setting the appropriate `app`.

firefox_ui_harness/runners/base.py
```diff
@@ -78,11 +28,3 @@ def get_application_folder(self, binary):
             return binary[:end_index]
         else:
             return os.path.dirname(binary)
-
-    def run_tests(self, tests):
-        # Ensure Marionette is always reset before starting tests
-        # This might need support in Marionette base
-        self.marionette = None
-        self.tests = []
```
whimboo, line 86: Wohoo! :+1:

firefox_ui_harness/runtests.py
```diff
@@ -7,71 +7,57 @@
...
-        else:
-            options.workspace = tempfile.mkdtemp('.{}'.format(os.path.basename(sys.argv[0])))
-
-        options.logger.info('Using workspace for temporary data: "{}"'.format(options.workspace))
+    def process_args(self):
+        # Make the default workspace TMP instead of .
+        # Note gecko.log and Fx profile will also be saved in workspace
+        # Note the workspace is not deleted after the test run
+        if not self.args.workspace:
+            self.args.workspace = tempfile.mkdtemp('.{}'.format(os.path.basename(sys.argv[0])))
```
whimboo: This code is something I do not understand. I thought that Marionette uses TEMP by default if no workspace option has been given.

whimboo:
All in all this looks great @mjzffr! I put a couple of questions in here which we might have to answer first. But a full testrun also with mozharness looks solid!

firefox_ui_harness/runtests.py
mjzffr: The default is to preserve the previous behaviour of Marionette runner, so workspace is cwd. Please see: https://bugzilla.mozilla.org/show_bug.cgi?id=1199588#c14

whimboo: Then we want to keep this behavior of Marionette. There is nothing why our tests should behave different by default.

firefox_ui_harness/runners/base.py
mjzffr, line 18: Will do.

firefox_ui_harness/runtests.py
mjzffr: @whimboo Done. Good thing you mentioned this because it caused me to discover an omission in `UpdateTestRunner`.

mjzffr:
@whimboo All comments addressed.

whimboo:
@mjzffr please try to avoid amending or squashing of commits in the future. This makes it impossible to see the latest changes and I will have to go through the whole PR again :(

firefox_ui_harness/runtests.py
```diff
@@ -7,71 +7,53 @@
...
-    except Exception:
-        logger.error('Failure during execution of the update test.',
-                     exc_info=True)
-        sys.exit(1)
-
+        if self.args.installer:
+            installer = os.path.realpath(self.args.installer)
+
+            dest_folder = os.path.abspath(os.path.expanduser(
+                os.path.join(self.args.workspace or '', 'binary')
```
whimboo: If there is no workspace option given we should also install to the TEMP folder. Kinda similar to the profile creation code. So I don't think we need all this extra abspath and expanduser calls.

firefox_ui_harness/runners/update.py
```diff
@@ -42,7 +42,7 @@ def __init__(self, **kwargs):
 
     def duplicate_application(self, application_folder):
         """Creates a copy of the specified binary."""
-        target_folder = os.path.join(self.workspace, 'binary.backup')
+        target_folder = os.path.join(self.workspace_path, 'binary.backup')
```
whimboo: The same here for the target. We should also use temp in case of no workspace being present. Maybe we can use a virtual workspace for all of the stuff?

mjzffr: My understanding of your comments is that firefox-ui-test-specific installer/binary stuff should go into TMP if no workspace is set.

whimboo: Right but the idea I had here is that we use a single folder under /tmp and have subfolders for the installed binary and the duplicated binary for update tests. So we don't end up with lots of different folders under tmp. But that's just an idea which we don't have to do. Having the correct behavior with --workspace is way more important.

mjzffr:
@whimboo Ok, so now if `--workspace` is not set, default location of gecko.log and Fx profile is the Marionette default (cwd and `$TMP/*.mozrunner` respectively), and default location of installer-related files is `$TMP/*.firefox-ui-update`.

I don't think we should overthink this given that the installer stuff should and will be handled by mozharness soon.

whimboo:
I retriggered all Travis builds and everything is passing now since the new packages have been mirrored to the internal pypi server.

firefox_ui_harness/runtests.py
```diff
@@ -7,71 +7,57 @@
...
-        logger.error('Failure during execution of the update test.',
-                     exc_info=True)
-        sys.exit(1)
-
+        if self.args.installer:
+            installer = os.path.realpath(self.args.installer)
+
+            if not self.args.workspace:
+                self.args.installer_workspace = tempfile.mkdtemp(
+                    '.{}'.format(os.path.basename(sys.argv[0]))
```
whimboo, line 32: Something doesn't play well with profile removal here. When I check the temporary folder once tests have been finished it is still existent. I assume that because we use an internal workspace here, marionette is not going to remove it. So we have two options here.... totally forget about a workspace by default or clean up the created folder under TMP at the end.

whimboo, line 32: Also for update tests we create two different profiles. Those sum up in case you use the workspace option multiple times. So more and more profile subdirs are getting created. I assume that using a specific folder for the profile still doesn't work (--profile workspace/profile) and we always would clone it?

whimboo:
I gave this PR some deep testing and found two issues with the workspace and profile as noted above. Otherwise this PR looks great and works wonderful. Good to see that we can remove so many code which is now part of Marionette proper.

firefox_ui_harness/runtests.py
mjzffr, line 32: A giant summary:
- `--workspace`: nothing in the workspace dir is ever deleted by Marionette runner, except profile clones (see --profile below), because the idea is that you might want to examine the content of the workspace to investigate test failures. If you specify a workspace, temporary profiles will be stored there for later inspection. It's the consumer's responsibility to clear the workspace.  
- `--installer`: in the absence of a workspace, if we create a temp dir to hold the installed binaries, then it's still our responsibility to delete that temp dir. If we "totally forget about a workspace by default" for the binaries, they will just go in cwd. I think it's better to create a temp dir and then clean it up.
- `--profile`: this is used to specify the **source** of a profile to copy, say `./awesome_profile`. Copies of `awesome_profile` are made in TMP or the workspace (if specified), and then they are deleted, since you will still have `./awesome_profile` available if you want to examine it. There's a special case: when `./awesome_profile` is empty to begin with. Then the "clones" are filed with fresh, temporary profile data, but they still get deleted in the end. In other words, if you want to use `--workspace` but you don't want to keep the profiles, you can use `--profile` to specify an empty or non-existant directory.

So, for example, if you run `firefox-ui-tests --installer blah --workspace build --profile build/some_empty_dir` at the end you will have a directory called `build` that contains `binary` (empty), `some_empty_dir` (empty), `gecko.log` and nothing else because the clones of `some_empty_dir` all get deleted from the workspace. 

Your decisions:
- How is the workspace treated in mozmill-ci for firefox-ui-tests? Do you delete it after each job? Do you keep it around for investigations?
- Do you want to keep profiles for investigation or always delete them immediately?

mjzffr:
@whimboo Pushed a clean-up of the installer workspace. Ready for more comments. :)

firefox_ui_harness/runtests.py
whimboo, line 32: Ok, so while talking about firefox-ui-tests we should not mix up any decisions about what gets done in mozmill-ci with workspaces. This is just about the plain `firefox-ui-tests` and `firefox-ui-update` CLI commands. For those I think we should have a behavior which is as close as possible to Marionette:
- **Profiles**: If no `--workspace` option is specified for our tests, no profile should remain at the end. If the option has been given then the profile will exist for examination. The only problem here is for our update tests which internally run Marionette twice and as a result two profiles will remain. Not sure at the moment if we should keep that or if the formerly profile has to be deleted. I assume this is even a question for Marionette and not only us given that this feature of multiple runs is in Marionette proper now.
- **Installer**: If no `--workspace``option is specified the binary gets installed under TMP and is also getting uninstalled. With the option set, the binary gets installed into the workspace AND also gets uninstalled. The latter is important especially for Windows given that the uninstaller cleans-up the registry and such.

If we cannot get those behaviors with a virtual internal workspace setting if it hasn't been specified via the command line, we should completely get rid of it and store everything under individual subfolder of TMP.

mjzffr, line 32: > If no --workspace option is specified for our tests, no profile should remain at the end.

Done

> If the option has been given then the profile will exist for examination. The only problem here is for our update tests which internally run Marionette twice and as a result two profiles will remain. 

In my opinion, this isn't a problem. It's a proper record of the test artifacts: many profiles were used, so many profiles remain for investigation. Again, I think it boils down to the consumer's responsibility to decide how to use or clean up the workspace.

> If no --workspace`option is specified the binary gets installed under TMP and is also getting uninstalled. 

Done + the directory under TMP gets deleted.

> With the option set, the binary gets installed into the workspace AND also gets uninstalled. 

Done

By the way, I'm not quite sure what you mean by "virtual internal workspace", but it seems like all your key requirements are satisfied.

firefox_ui_harness/runtests.py
```diff
@@ -6,72 +6,68 @@
...
+
+
+def cli():
+    mn_cli(runner_class=FirefoxUITestRunner,
+           parser_class=FirefoxUIArguments,
+           harness_class=FirefoxUIHarness)
 
 if __name__ == '__main__':
-    sys.exit(cli())
+    cli()
```
whimboo, line 73: We would still exit with 10 in case of test failures, right?

firefox_ui_harness/runtests.py
whimboo, line 32: > ```
> If the option has been given then the profile will exist for
> examination. The only problem here is for our update tests which
> internally run Marionette twice and as a result two profiles will
> remain.
> ```
> 
> In my opinion, this isn't a problem. It's a proper record of the test
> artifacts: many profiles were used, so many profiles remain for
> investigation. Again, I think it boils down to the consumer's
> responsibility to decide how to use or clean up the workspace.

I may not have phrased it problem. `Question` sounds a bit more what I
was thinking of. :) So yeah, lets see how this works out. And maybe it's
indeed better to have those profile remaining. It's at least something
we have to account for job retention in mozmill-ci and the disk usage
over time. Each profile will consume about 12MB and with about 250
builds per job in history it counts up.

> By the way, I'm not quite sure what you mean by "virtual internal
> workspace", but it seems like all your key requirements are satisfied.

It's about the `installer_workspace` variable which conflicts a bit with
the real workspace. But that's all.

firefox_ui_harness/runtests.py
mjzffr, line 73: Yes: https://dxr.mozilla.org/mozilla-central/source/testing/marionette/client/marionette/runtests.py#92

firefox_ui_harness/runners/update.py
```diff
@@ -38,11 +38,16 @@ def __init__(self, **kwargs):
...
+        # Used in case no workspace is set
+        self.installer_workspace = kwargs.pop('installer_workspace',
+                                              self.workspace_path)
+
         self.test_handlers = [UpdateTestCase]
 
     def duplicate_application(self, application_folder):
         """Creates a copy of the specified binary."""
-        target_folder = os.path.join(self.workspace, 'binary.backup')
+
```
whimboo, line 49: FYI, by default there is no empty line after a docstring (see PEP257).

whimboo:
I tested the PR again and it works as expected now. Thanks @mjzffr for the latest commit. I will go ahead and merge this PR in a moment after squashing and rebasing it.

whimboo:
Merged to mozilla-central as https://github.com/mozilla/firefox-ui-tests/commit/32be49d74e1d10c6bf087235b1d6753c1b840bc4.

PR #289, opened 2015-11-20
Bug 1143928 - Refactor Deck and Panel classes to reduce code duplicat…

utvar:
…ion. r=whimboo


utvar:
Work in Progress.

firefox_puppeteer/ui/deck.py
```diff
@@ -0,0 +1,47 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from marionette_driver import By
+
+from firefox_puppeteer.ui_base_lib import UIBaseLib
+
+
+class BaseDeck(UIBaseLib):
```
whimboo: I would keep `Deck` as the name of the base class and change the name of the specific subclass.

firefox_puppeteer/ui/deck.py
```diff
@@ -0,0 +1,47 @@
...
+                   'checkingForUpdates': CheckingForUpdatesPanel,
+                   'downloadAndInstall': DownloadAndInstallPanel,
+                   'downloadFailed': DownloadFailedPanel,
+                   'downloading': DownloadingPanel,
+                   'feedPanel': FeedPanel,
+                   'generalPanel': GeneralPanel,
+                   'mediaPanel': MediaPanel,
+                   'noUpdatesFound': NoUpdatesFoundPanel,
+                   'permPanel': PermissionsPanel,
+                   'securityPanel': SecurityPanel
```
whimboo: All those panels are specific classes for the update deck in the about window. So they have to stay as part of the subclass in the other file. As of now the class references cannot be found.

firefox_puppeteer/ui/deck.py
```diff
@@ -0,0 +1,47 @@
...
+                   }
+
+        panel = self.element.find_element(By.ID, panel_id)
+        return mapping.get(panel_id, BasePanel)(lambda: self.marionette, self.window, panel)
+
+
+class BasePanel(UIBaseLib):
```
whimboo: Same here. Keep the name as we had before for the base class.

firefox_puppeteer/ui/deck.py
utvar: @whimboo, I have still not found a way to extract the _create_panel_for_id method without also pulling out all the panel classes. It seems the only things remaining in Deck that are common to both about_window and pageinfo, that can be extracted without also pulling out anything specific to one of these two, are three properties: panels, selected_index, and selected_panel. (These three are also in update_wizard, inside the Wizard class.) Is this what you would like refactored out?

whimboo: `_create_panel_for_id()` can be made an abstract method so subclasses have to implement their own mapping logic.

Even better would be to create the mapping object at class level, so the `_create_panel_for_id()` method in the base class could access it via e.g. `self.panel_map`. With that each subclass would only have to create this object and everything else happens magically.

utvar:
@whimboo, not sure why the one Linux build failed -- I looked at the errors and am fairly sure it is not my code. The tests pass locally using firefox-ui-functional. (Using firefox-ui-tests returns ImportError: 'module' object has no attribute 'cli', for the mozilla-central branch as well.)

I'm not sure what will happen to this PR now that the code has been moved, but I at least wanted to return my code to a working state, so Deck is back where it was and only Panel has been refactored out.

whimboo:
This repository stays as the canonical repository for now. At least until we can do per checkin tests in hg.mozilla.org/mozilla-central.

The failing test here is a known intermittent failure I was never able to reproduce myself. So don't worry about it. The rest of this PR would be ready for another review now?

utvar:
@whimboo, Travis results are mixed, but as far as I can tell, it is not related to my code.

whimboo:
This looks pretty fine for the base Panel class. Do you want to fold in the refactoring of the base Deck class in the same pull or do that on another one? I'm open for both.

utvar:
@whimboo, I'd prefer to do a separate PR for the Deck class, because I'm not sure how long it will take me to finish that one.

whimboo:
Alright, I see. I will have a closer look at this PR now.

whimboo:
The code changes look fine and are at the right places. But I noticed that we do not have a rst file for the documentation yet. Please add one for the new ui module under `firefox_puppeteer/docs/ui/`.

utvar:
@whimboo, here is the addition you requested.

whimboo:
I think that looks fine now. I will get it merged. Thanks @utvar!

whimboo:
This PR got merged as https://github.com/mozilla/firefox-ui-tests/commit/023fb06a01b86d56363a0dc46c2317a79aa09570.

PR #291, opened 2015-11-25
Bug 1227570 - Remove support of the --installer command line option

whimboo:
This PR is not yet ready for review. The configuration for Travis needs to be updated to make use of mozharness and it's installer.


whimboo:
The Travis failure here is only for a missing ru build which is a known problem.

mjzffr:
Looks good. :) Just one suggestion.

PR #292, opened 2015-11-26
Bug 1228120 - [WIP] Add SessionStore tests for restarting. r?whimboo

mikeconley:
Hey :whimboo - I'm still really, really new to this, and this test isn't 100% done yet, but I wanted some feedback (and some tips, since this thing doesn't work properly just yet).
1. How come I can't seem to get the locations for each window tab on line 65? When I use pdb, apparently an exception like "**\* NoSuchWindowException: NoSuchWindowException: Unable to locate window: 2147483649" is being fired when I try to access the location, but the testing framework never tells me that itself. What can I do to access each tabs location?
2. Even if I remote the bit where I access the tab locations, this test never exits. It'll just sit there forever. What am I doing wrong that's preventing this test from exiting?

### DIFF
--- a/firefox_ui_tests/functional/sessionstore/test_restore_windows_after_restart.py
+++ b/firefox_ui_tests/functional/sessionstore/test_restore_windows_after_restart.py
@@ -1,0 +1,94 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from marionette_driver import Wait
+
+import time
+
+from firefox_puppeteer.testcases import FirefoxTestCase
+
+
+class TestRestoreWindowsAfterRestart(FirefoxTestCase):
+
+    def setUp(self):
+        FirefoxTestCase.setUp(self)
+
+        # Each list element represents a window of tabs loaded at
+        # some testing URL
+        self.test_windows = [
+          # Window 1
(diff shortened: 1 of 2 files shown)


firefox_ui_tests/functional/sessionstore/test_restore_windows_after_restart.py
```diff
@@ -0,0 +1,94 @@
...
+           self.marionette.absolute_url('layout/mozilla_mission.html')],
+        ]
+
+        # Set browser to restore previous session
+        self.prefs.set_pref('browser.startup.page', 3)
+
+
+    def tearDown(self):
+        try:
+            self.windows.close_all([self.browser.tabbar.tabs[0]])
```
whimboo, line 46: Please note that you have to use `self.browser` here. Otherwise Marionette will hang.  Also close all additional tabs you opened to not get the leak handle exception.

mikeconley, line 46: Hey, sorry whimboo - just getting back to this now. Where should I be using self.browser? Should I be doing self.browser.windows.close_all instead?

whimboo, line 46: No, with `self.browser` I meant the entry in the exclusion list (parameter) of the function call. As of now you are  trying to close all windows, except the first tab of the browser window. By using that Marionette will also close the initial browser window, and will hang until the global timeout.

PR #299, opened 2015-12-15
Bug 1232331 - Factor out test logic from UI classes (about window, software update dialog) to update testcase

whimboo:


whimboo:
Failures as reported by Travis are due to no Russian builds to be downloadable at the moment.

PR #300, opened 2015-12-16
Bug 1232967 - Create Python packages for firefox_ui_harness, firefox_puppeteer, and firefox_ui_tests

whimboo:


firefox_puppeteer/firefox_puppeteer/api/appinfo.py
```diff
@@ -2,7 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
-from firefox_puppeteer.base import BaseLib
+from ..base import BaseLib
```
sydvicious: Be sure to test this in all environments. We had trouble with relative imports in media_tests when we first got off the ground.

whimboo: I digged around a bit for absolute vs. relative imports and found the Python docs: https://docs.python.org/2/whatsnew/2.5.html#pep-328-absolute-and-relative-imports

I will check all packages accordingly.

whimboo: I decided to kill all relative imports and use absolute ones. It definitely makes it safer. Thanks for pointing your issues out.

whimboo:
> It seems weird to me that include the puppeteer tests with firefox_ui_tests, but ok. :)

The problem is that those tests are not unit tests and require Firefox and our testcase classes to run. Because of that those tests are better homed outside of the puppeteer repository. For real unit tests we should make use of mock. That's what I got out of my conversation with @davehunt the other day when we talked about the POM he is working on.

whimboo:
I have fixed all remaining issues and made the proposed changes from the last review. Travis is happy now.

whimboo:
This code here got merged in a new official branch:
https://github.com/mozilla/firefox-ui-tests/tree/python_packages

I'm going to close this PR.

PR #308, opened 2015-12-22
Bug 1232967 - Update requirements.txt and travis settings to work in both packaged and non-packaged mode

whimboo:


whimboo:
Something went wrong here. Given that I cannot move the target pointer I will have to close this PR.

PR #311, opened 2016-01-06
Bug 1228278 - Fix handling of editBookmarkPanel

whimboo:


whimboo:
The failures here seem to be a very strange regression which is also visible for latest Nightly builds on Treeherder without this patch! I'm investigating...

whimboo:
I rebased the branch for latest changes on mozilla-central. This should turn the test results finally green now. Lets see what Travis shows us.

whimboo:
Travis caught a failure! \o/ In case of the `self.bookmark_panel` property being None we currently fail. So I submitted another commit with an additional if.

PR #318, opened 2016-01-18
Bug 1232967 - Create Python packages for firefox_ui_harness, firefox_puppeteer, and firefox-ui-tests

whimboo:


firefox_puppeteer/firefox_puppeteer/api/software_update.py
```diff
@@ -7,9 +7,9 @@
 
 import mozinfo
 
+from firefox_puppeteer.base import BaseLib
 from firefox_puppeteer.api.appinfo import AppInfo
 from firefox_puppeteer.api.prefs import Preferences
-from firefox_puppeteer.base import BaseLib
```
sydvicious, line 13: So, I don't see the rhyme or reason of how these were ordered before or after. Shouldn't this be sorted alphabetically by package and then module?

Note that I won't block the review for this; it's not that big a deal.

sydvicious:
@whimboo r+

firefox_puppeteer/firefox_puppeteer/api/software_update.py
whimboo, line 13: Basically this is true, but given that this is a plain backport, we should get this fixed on the mozilla-central branch first.

PR #328, opened 2016-02-11
Bug 1237396 - Convert testSafeBrowsing_initialDownload Mozmill test to Marionette. r=whimboo, maja_zf

b4handjr:
Hoping everything checks out on this, as it is my first ever contribution on github. Please feel free to tear it up!

### DIFF
--- a/firefox_ui_tests/firefox_ui_tests/functional/security/manifest.ini
+++ b/firefox_ui_tests/firefox_ui_tests/functional/security/manifest.ini
@@ -11,6 +11,7 @@ tags = local
 tags = local
 [test_safe_browsing_notification.py]
 [test_safe_browsing_warning_pages.py]
+[test_safe_browsing_primary_download.py]
 [test_security_notification.py]
 [test_ssl_disabled_error_page.py]
 [test_ssl_status_after_restart.py]
--- a/firefox_ui_tests/firefox_ui_tests/functional/security/test_safe_browsing_primary_download.py
+++ b/firefox_ui_tests/firefox_ui_tests/functional/security/test_safe_browsing_primary_download.py
@@ -24,6 +24,9 @@ class TestSafeBrowsingPrimaryDownload(FirefoxTestCase):
                 "goog-badbinurl-shavar.cache",
                 "goog-badbinurl-shavar.pset",
                 "goog-badbinurl-shavar.sbstore",
+                "goog-unwanted-shavar.cache",
+                "goog-unwanted-shavar.pset",
+                "goog-unwanted-shavar.sbstore",
 
                 # Tracking Protections
                 "mozstd-track-digest256.cache",
--- a/firefox_ui_tests/firefox_ui_tests/functional/security/test_safe_browsing_primary_download.py
+++ b/firefox_ui_tests/firefox_ui_tests/functional/security/test_safe_browsing_primary_download.py
@@ -1,5 +1,6 @@
-import sys
 import os
+import sys
+
 from marionette_driver import By, Wait
 from marionette.marionette_test import skip_if_e10s
 
@@ -7,53 +8,79 @@ from firefox_puppeteer.testcases import FirefoxTestCase
 
 from firefox_puppeteer.ui.browser.window import BrowserWindow
 
-
 class TestSafeBrowsingPrimaryDownload(FirefoxTestCase):
 
-    file_names = [
-        "goog-phish-shavar.cache",
-        "goog-phish-shavar.pset",
-        "goog-phish-shavar.sbstore",
-        "goog-malware-shavar.cache",
-        "goog-malware-shavar.pset",
-        "goog-malware-shavar.sbstore",
-        "goog-badbinurl-shavar.cache",
-        "goog-badbinurl-shavar.pset",
-        "goog-badbinurl-shavar.sbstore"
-    ]
+    data = [
+        {
+            'platform': ['linux', 'win', 'darwin'],
+            'files': [
+                #Phishing
+                "goog-phish-shavar.cache",
+                "goog-phish-shavar.pset",
+                "goog-phish-shavar.sbstore",
+                "goog-malware-shavar.cache",
+                "goog-malware-shavar.pset",
+                "goog-malware-shavar.sbstore",
+                "goog-badbinurl-shavar.cache",
+                "goog-badbinurl-shavar.pset",
+                "goog-badbinurl-shavar.sbstore",
 
-    win_file_names = [
-        "goog-downloadwhite-digest256.cache",
-        "goog-downloadwhite-digest256.pset",
-        "goog-downloadwhite-digest256.sbstore"
-    ]
 
-    tracking_protection = [
-        "mozstd-track-digest256.cache",
-        "mozstd-track-digest256.pset",
-        "mozstd-track-digest256.sbstore",
-        "mozstd-trackwhite-digest256.cache",
-        "mozstd-trackwhite-digest256.pset",
-        "mozstd-trackwhite-digest256.sbstore"
-    ]
+                #Tracking Protections
+                "mozstd-track-digest256.cache",
+                "mozstd-track-digest256.pset",
+                "mozstd-track-digest256.sbstore",
+                "mozstd-trackwhite-digest256.cache",
+                "mozstd-trackwhite-digest256.pset",
+                "mozstd-trackwhite-digest256.sbstore"
+                ]
+            },
+            {
+                'platform': ['win'],
+                'files': [
+                    "goog-downloadwhite-digest256.cache",
+                    "goog-downloadwhite-digest256.pset",
+                    "goog-downloadwhite-digest256.sbstore"
+                ]
+             }
+        ]
 
     def setUp(self):
         FirefoxTestCase.setUp(self)
-        safe_browsing_path = os.path.join(self.marionette.instance.profile.profile, 'safebrowsing')
-        self.files_location = safe_browsing_path
+        self.test_url = 'https://mozqa.com'
+
+        self.prefs.set_pref('browser.safebrowsing.provider.google.lastupdatetime', 1)
+        self.prefs.set_pref('browser.safebrowsing.provider.google.nextupdatetime', 1)
+        self.prefs.set_pref('browser.safebrowsing.enabled', 'true')
+
+        self.sb_files_path = os.path.join(self.marionette.instance.profile.profile, 'safebrowsing')
+        print(self.sb_files_path)
+
+
+        #for item in data:
+        #   if 'linux' in item['platform']:
+        #        print 'platform is linux: %s' % item['platform']
 
     def test_safe_browsing(self):
-        if sys.platform.startswith('win32'):
-            self.verify_win_files_existence()
-        else:
+        with self.marionette.using_context('content'):
+            self.marionette.navigate(self.test_url)
+
+        if sys.platform.startswith('linux'):
             self.verify_files_existence()
+            print("linux")
+        elif sys.platform.startswith('win32'):
+            print('win32')
+            #self.verify_win_files_existence()
 
     def verify_files_existence(self):
-        for file in self.file_names:
-            if self.assertTrue(
-                    Wait(self.marionette, timeout=60).until(
-                        lambda _: os.path.exists(os.path.join(self.files_location, file)))):
-                self.verify_tracking_protection_existence()
+        for data in self.data:
+            if 'linux' in data['platform']:
+                for item in data['files']:
+                    print("true") ## Remove
+                    self.assertTrue(
+                        Wait(self.marionette, timeout=300).until(
+                            lambda _: os.path.exists(os.path.join(self.sb_files_path, item))))
+
 
     def verify_win_files_existence(self):
         for file in self.win_file_names:
@@ -63,7 +90,16 @@ class TestSafeBrowsingPrimaryDownload(FirefoxTestCase):
                 self.verify_tracking_protection_existence()
 
     def verify_tracking_protection_existence(self):
-        for file in self.tracking_protection:
-            self.assertTrue(
-                Wait(self.marionette, timeout=60).until(
-                    lambda _: os.path.exists(os.path.join(self.files_location, file))))
+        for k, v in self.file_names.items():
+            if k == 'tracking':
+                #print('inside linux')
+                #print(v)
+                for file in v:
+                    self.assertTrue(
+                        Wait(self.marionette, timeout=30).until(
+                            lambda _: os.path.exists(os.path.join(self.sb_files_path, file))))
+
+        #for file in self.tracking_protection:
+        #    self.assertTrue(
+        #        Wait(self.marionette, timeout=60).until(
+        #            lambda _: os.path.exists(os.path.join(self.files_location, file))))
--- a/firefox_ui_tests/firefox_ui_tests/functional/security/test_safe_browsing_primary_download.py
+++ b/firefox_ui_tests/firefox_ui_tests/functional/security/test_safe_browsing_primary_download.py
@@ -67,7 +67,7 @@ class TestSafeBrowsingPrimaryDownload(FirefoxTestCase):
         # Set Variable to tmp safebrowsing directory
         self.sb_files_path = os.path.join(self.marionette.instance.profile.profile, 'safebrowsing')
 
-    def test_safe_browsing(self):
+    def test_safe_browsing_initial_download(self):
         for data in self.test_data:
             if sys.platform in data['platforms']:
                 for item in data['files']:
--- a/firefox_ui_tests/firefox_ui_tests/functional/security/test_safe_browsing_primary_download.py
+++ b/firefox_ui_tests/firefox_ui_tests/functional/security/test_safe_browsing_primary_download.py
@@ -9,7 +9,7 @@ class TestSafeBrowsingPrimaryDownload(FirefoxTestCase):
 
     data = [
         {
-            'platform': ['linux', 'win', 'darwin'],
+            'platform': ['linux2', 'win32', 'darwin'],
             'files': [
                 # Phishing
                 "goog-badbinurl-shavar.cache",
@@ -35,7 +35,7 @@ class TestSafeBrowsingPrimaryDownload(FirefoxTestCase):
                 ]
             },
         {
-            'platform': ['win'],
+            'platform': ['win32'],
             'files': [
                 "goog-downloadwhite-digest256.cache",
                 "goog-downloadwhite-digest256.pset",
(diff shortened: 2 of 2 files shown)


whimboo:
So you run into timeout issues, this is most likely because you do not wait long enough or even better let Firefox download the files earlier. For the latter please read through the bug and check comment 5. Setting this preference should help us a lot here.

firefox_ui_tests/firefox_ui_tests/functional/security/test_safe_browsing_primary_download.py
```diff
@@ -0,0 +1,69 @@
+import sys
+import os
+from marionette_driver import By, Wait
+from marionette.marionette_test import skip_if_e10s
+
+from firefox_puppeteer.testcases import FirefoxTestCase
+
+from firefox_puppeteer.ui.browser.window import BrowserWindow
```
whimboo: Please read https://www.python.org/dev/peps/pep-0008/#imports in how the coding style for imports should look like.

firefox_ui_tests/firefox_ui_tests/functional/security/test_safe_browsing_primary_download.py
```diff
@@ -0,0 +1,69 @@
...
+    ]
+
+    tracking_protection = [
+        "mozstd-track-digest256.cache",
+        "mozstd-track-digest256.pset",
+        "mozstd-track-digest256.sbstore",
+        "mozstd-trackwhite-digest256.cache",
+        "mozstd-trackwhite-digest256.pset",
+        "mozstd-trackwhite-digest256.sbstore"
+    ]
```
whimboo: Personally I would make one dictionary with all those file names. It can be separated by type of protection and limitation for platforms.

firefox_ui_tests/firefox_ui_tests/functional/security/test_safe_browsing_primary_download.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def setUp(self):
+        FirefoxTestCase.setUp(self)
+        safe_browsing_path = os.path.join(self.marionette.instance.profile.profile, 'safebrowsing')
+        self.files_location = safe_browsing_path
```
whimboo: You don't really need the `safe_browsing_path`. Just assign it directly to e.g. `self.sb_files_path`.

firefox_ui_tests/firefox_ui_tests/functional/security/test_safe_browsing_primary_download.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def test_safe_browsing(self):
+        if sys.platform.startswith('win32'):
+            self.verify_win_files_existence()
+        else:
+            self.verify_files_existence()
```
whimboo: With a single dictionary as mentioned above you can have a for loop iterating over each item, and do the check if the platform matches. So the logic will become way easier.

b4handjr:
@whimboo Would you suggest both a 5 min timout and the nextupdatetime variable set?

whimboo:
For setting the preference you can use the following method: http://firefox-puppeteer.readthedocs.org/en/latest/api/prefs.html#firefox_puppeteer.api.prefs.Preferences.set_pref

whimboo:
5 minutes is definitely too long. With the preferences set the files should appear within seconds. Maybe you record some timings to see how fast they download?

firefox_ui_tests/firefox_ui_tests/functional/security/test_safe_browsing_primary_download.py
```diff
@@ -0,0 +1,88 @@
...
+                # Phishing
+                "goog-phish-shavar.cache",
+                "goog-phish-shavar.pset",
+                "goog-phish-shavar.sbstore",
+                "goog-malware-shavar.cache",
+                "goog-malware-shavar.pset",
+                "goog-malware-shavar.sbstore",
+                "goog-badbinurl-shavar.cache",
+                "goog-badbinurl-shavar.pset",
+                "goog-badbinurl-shavar.sbstore",
```
fmarier: You can also add `goog-unwanted-shavar`.

firefox_ui_tests/firefox_ui_tests/functional/security/test_safe_browsing_primary_download.py
```diff
@@ -0,0 +1,88 @@
...
+    ]
+
+    def setUp(self):
+        FirefoxTestCase.setUp(self)
+
+        # Set Browser URL
+        self.test_url = 'https://mozqa.com'
+
+        # Set Browser Preferences
+        self.prefs.set_pref('browser.safebrowsing.provider.google.lastupdatetime', 1)
```
fmarier: I think that `browser.safebrowsing.provider.google.lastupdatetime` is not needed (sorry about that, it was my fault for mentioning it in the first place).

firefox_ui_tests/firefox_ui_tests/functional/security/test_safe_browsing_primary_download.py
```diff
@@ -0,0 +1,88 @@
...
+        self.prefs.set_pref('browser.safebrowsing.provider.google.nextupdatetime', 1)
+        self.prefs.set_pref('browser.safebrowsing.enabled', 'true')
+        self.prefs.set_pref('browser.safebrowsing.malware.enabled', 'true')
```
fmarier: These two prefs default to `true`, but if you want to be explicit, I would also do:

```
browser.safebrowsing.downloads.enabled = true
browser.safebrowsing.downloads.remote.enabled = true
privacy.trackingprotection.pbmode.enabled = true
```

firefox_ui_tests/firefox_ui_tests/functional/security/test_safe_browsing_primary_download.py
```diff
@@ -0,0 +1,88 @@
+import os
+import sys
+
+from marionette_driver import By, Wait
+from marionette.marionette_test import skip_if_e10s
```
fmarier: If marionette works on e10s, I would recommend not skipping it. Safe Browsing and Tracking Protection both work fine with e10s.

firefox_ui_tests/firefox_ui_tests/functional/security/test_safe_browsing_primary_download.py
b4handjr: Is there a better way than to set each preference individually?

firefox_ui_tests/firefox_ui_tests/functional/security/test_safe_browsing_primary_download.py
```diff
@@ -0,0 +1,97 @@
...
+        # Restart Browser
+        self.restart()
+
+        # Set Browser URL
+        self.test_url = 'https://mozqa.com'
+
+        # Set Browser Preferences
+        self.prefs.set_pref('browser.safebrowsing.provider.google.nextupdatetime', 1)
+        self.prefs.set_pref('browser.safebrowsing.provider.mozilla.nextupdatetime', 1)
+        self.prefs.set_pref('browser.safebrowsing.provider.mozilla.lastupdatetime', 1)
```
fmarier: `lastupdatetime` shouldn't be necessary.

firefox_ui_tests/firefox_ui_tests/functional/security/test_safe_browsing_primary_download.py
whimboo: I would prefer to have an array similar to files with the name and value. So we can easily loop through all preferences here and have all test data definition at the same place.

firefox_ui_tests/firefox_ui_tests/functional/security/test_safe_browsing_primary_download.py
whimboo: This skip is indeed not necessary if the test works in e10s mode. If Travis passes without it, we can be sure about it.

firefox_ui_tests/firefox_ui_tests/functional/security/manifest.ini
```diff
@@ -11,6 +11,7 @@ tags = local
 tags = local
 [test_safe_browsing_notification.py]
 [test_safe_browsing_warning_pages.py]
+[test_safe_browsing_primary_download.py]
```
whimboo, line 14: nit: please sort the files alphabetically.

firefox_ui_tests/firefox_ui_tests/functional/security/test_safe_browsing_primary_download.py
```diff
@@ -0,0 +1,96 @@
+import os
+import sys
+
+from marionette_driver import By, Wait
+
+from firefox_puppeteer.testcases import FirefoxTestCase
+
+from firefox_puppeteer.ui.browser.window import BrowserWindow
```
whimboo: nit: no need for the additional empty line between the puppeteer imports.

whimboo: Whereby I don't see that BrowserWindow is used at all. So you can remove line 8.

firefox_ui_tests/firefox_ui_tests/functional/security/test_safe_browsing_primary_download.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+class TestSafeBrowsingPrimaryDownload(FirefoxTestCase):
+
+    data = [
+        {
+            'platform': ['linux', 'win', 'darwin'],
```
whimboo: nit: platforms given that you have a list here.

firefox_ui_tests/firefox_ui_tests/functional/security/test_safe_browsing_primary_download.py
```diff
@@ -0,0 +1,96 @@
...
+                "goog-downloadwhite-digest256.sbstore"
+            ]
+        }
+    ]
+
+    def setUp(self):
+        FirefoxTestCase.setUp(self)
+
+        # Restart Browser
+        self.restart()
```
whimboo, line 64: You have to set the preferences before you restart the browser.

firefox_ui_tests/firefox_ui_tests/functional/security/test_safe_browsing_primary_download.py
```diff
@@ -0,0 +1,96 @@
...
+        # Set Browser Preferences
+        self.prefs.set_pref('browser.safebrowsing.provider.google.nextupdatetime', 1)
+        self.prefs.set_pref('browser.safebrowsing.provider.mozilla.nextupdatetime', 1)
+        self.prefs.set_pref('browser.safebrowsing.enabled', 'true')
+        self.prefs.set_pref('browser.safebrowsing.malware.enabled', 'true')
+        self.prefs.set_pref('browser.safebrowsing.downloads.enabled', 'true')
+        self.prefs.set_pref('browser.safebrowsing.downloads.remote.enabled', 'true')
+        self.prefs.set_pref('privacy.trackingprotection.pbmode.enabled', 'true')
+
+        # Set variable to join path
```
whimboo: The comment should reflect what the variable contains, and not pseudo-code of the next line.

firefox_ui_tests/firefox_ui_tests/functional/security/test_safe_browsing_primary_download.py
```diff
@@ -0,0 +1,96 @@
...
+        self.sb_files_path = os.path.join(self.marionette.instance.profile.profile, 'safebrowsing')
+
+    def test_safe_browsing(self):
+        with self.marionette.using_context('content'):
+            self.marionette.navigate(self.test_url)
```
whimboo: Is there a reason why we have to navigate to this page? I don't see that this would trigger the download of the wanted files.

firefox_ui_tests/firefox_ui_tests/functional/security/test_safe_browsing_primary_download.py
```diff
@@ -0,0 +1,96 @@
...
+    def verify_files_existence(self):
+        for data in self.data:
+            if ('linux' or 'darwin') in data['platform']:
+                for item in data['files']:
+                    self.assertTrue(
+                        Wait(self.marionette, timeout=300).until(
+                            lambda _: os.path.exists(os.path.join(self.sb_files_path, item))))
+
+    # Test windows specific files
+    def verify_win_files_existence(self):
```
whimboo: This method is code-wise completely identical to the one above. you won't need any of those. Just put all the code in the test itself.

firefox_ui_tests/firefox_ui_tests/functional/security/test_safe_browsing_primary_download.py
```diff
@@ -0,0 +1,96 @@
...
+        self.sb_files_path = os.path.join(self.marionette.instance.profile.profile, 'safebrowsing')
+
+    def test_safe_browsing(self):
+        with self.marionette.using_context('content'):
+            self.marionette.navigate(self.test_url)
+
+        if sys.platform.startswith('linux') or sys.platform.startswith('darwin'):
+            self.verify_files_existence()
+        elif sys.platform.startswith('win32'):
+            self.verify_win_files_existence()
```
whimboo: No need for this if/else construct. When you iterate through the `data` list with `for` simply check if the current platform is contained in the `platforms` list. If yes check the files, otherwise `continue`.

firefox_ui_tests/firefox_ui_tests/functional/security/test_safe_browsing_primary_download.py
```diff
@@ -0,0 +1,96 @@
...
+
+    # Test all besides windows specific
+    def verify_files_existence(self):
+        for data in self.data:
+            if ('linux' or 'darwin') in data['platform']:
+                for item in data['files']:
+                    self.assertTrue(
+                        Wait(self.marionette, timeout=300).until(
+                            lambda _: os.path.exists(os.path.join(self.sb_files_path, item))))
```
whimboo: When you use `Wait().until()` you do not have to use `self.assert*()`. The method `until()` will raise a TimeoutError. Also please reduce the timeout to 30s which is the default timeout for accessing objects which are served remotely. Maybe we can hijack the `self.browser.timeout_page_load` value here.

b4handjr:
I hope this is everything!

b4handjr:
@whimboo Please let me know if the platform list variable name is okay

firefox_ui_tests/firefox_ui_tests/functional/security/manifest.ini
whimboo, line 14: Benny, this still needs a correction.

firefox_ui_tests/firefox_ui_tests/functional/security/test_safe_browsing_primary_download.py
```diff
@@ -0,0 +1,75 @@
+import os
+import sys
+
+from marionette_driver import By, Wait
```
whimboo: nit: `By` is also not used in this test, so no need to import it.

firefox_ui_tests/firefox_ui_tests/functional/security/test_safe_browsing_primary_download.py
```diff
@@ -0,0 +1,75 @@
...
+from firefox_puppeteer.testcases import FirefoxTestCase
+
+
+class TestSafeBrowsingPrimaryDownload(FirefoxTestCase):
+
+    data = [
+        {
+            'platform_list': ['linux2', 'win32', 'darwin'],
```
whimboo: As suggested please use `platforms` here so it is equal to `files`. Usually you should not use the data type in the variable name given that it will cause problems when the type changes.

firefox_ui_tests/firefox_ui_tests/functional/security/test_safe_browsing_primary_download.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def setUp(self):
+        FirefoxTestCase.setUp(self)
+
+        # Restart Browser
+        self.restart()
+
+        # Set Browser Preferences
+        for item, value in self.browser_prefs.items():
+            self.prefs.set_pref(item, value)
```
whimboo, line 61: We still want this before the restart otherwise it could cause race conditions. Also I haven't seen any comment from you regarding my question on IRC earlier this week, if the preferences as set here are getting reset with the restart. Please check that.

whimboo, line 61: Maybe it's a problem that we do the restart in setup(). Try to do it in the test itself and check if that makes a difference. IMO this is even the better place to do it.

firefox_ui_tests/firefox_ui_tests/functional/security/test_safe_browsing_primary_download.py
```diff
@@ -0,0 +1,75 @@
...
+
+        # Set Variable to tmp safebrowsing directory
+        self.sb_files_path = os.path.join(self.marionette.instance.profile.profile, 'safebrowsing')
+
+    def test_safe_browsing(self):
```
whimboo: For a better name please include `initial_download`. In case of test failures this will be better distinguishable on Treeherder.

firefox_ui_tests/firefox_ui_tests/functional/security/test_safe_browsing_primary_download.py
```diff
@@ -0,0 +1,75 @@
...
+        for data in self.data:
+            if sys.platform in data['platform_list']:
```
whimboo: Lets stay with the marionette capabilities here, it's a nice abstraction and ensures we have the right values. Especially on Windows with mingw involved.

firefox_ui_tests/firefox_ui_tests/functional/security/test_safe_browsing_primary_download.py
```diff
@@ -0,0 +1,75 @@
...
+                for item in data['files']:
+                    Wait(self.marionette, timeout=self.browser.timeout_page_load).until(
+                        lambda _: os.path.exists(os.path.join(self.sb_files_path, item)))
```
whimboo: In case of a timeout what is the output here? If it's only `False != True` we should add a comment as extra parameter for the call to `until()`.

firefox_ui_tests/firefox_ui_tests/functional/security/test_safe_browsing_primary_download.py
whimboo: Puppeteer also has a property for that: https://github.com/mozilla/firefox-ui-tests/blob/mozilla-central/firefox_puppeteer/firefox_puppeteer/__init__.py#L66. So you can easily use `self.platform` here.

